Issue #66, opened 2021-05-25
Document Calico Addon


kcoleman731:
Overview of Calico
Link to further docs
Default network policies we apply
https://github.com/aws-quickstart/quickstart-ssp-amazon-eks/projects/1

Issue #31, opened 2021-05-25
[PATTERN] - Add Kubernetes network policies to Sample Team


kcoleman731:
To demonstrate how a customer can leverage the K8s network policies, we will add network policies to one of our teams which restricts traffic from outside of the namespace and only allows egress to certain resources (TBD). 

We will also need to document these policies.

PR #87, opened 2021-06-15
Feature/calico

youngjeong46:
Issue #66 - Documenting Calico Addon

Added Calico documentations under addon


docs/addons/calico-allow-burnham-riker.yaml
```diff
@@ -0,0 +1,21 @@
+# This GlobalNetworkPolicy uses Calico's CRD
```
kcoleman731, line 1: do we use these files anywhere in the project? Are they just samples? If so, are they redundant since you have them in docs?

youngjeong46: These are just samples, since I copied them straight into the doc, I can remove those two yaml files

youngjeong46: files are removed.

docs/addons/calico.md
```diff
@@ -0,0 +1,190 @@
+# Calico Add-on
+
+The `Calico` addon adds support for [Calico CNI](https://docs.projectcalico.org/about/about-calico) to an EKS cluster.
```
shapirov103, line 3: We should state that it adds support for Kubernetes Network Policies as opposed to CNI. Calico is never installed as full CNI (vpc-cni is used) and it is only leveraged as a network policy engine.

docs/addons/calico.md
```diff
@@ -0,0 +1,190 @@
...
+      account: <AWS_ACCOUNT_ID>,
+      region: <AWS_REGION>,
+  },
+});
+```
+## Securing your environment with Calico network policies
+
+### Installing calicoctl
+
+In order to use Calico Customer Resource Definitions (CRD), you need to install `calicoctl`.
```
shapirov103, line 27: It is Custom (not customer)

shapirov103: Let's add a sentence on why customer would want to have custom resource definitions. We should state that by default just installing Calico for network policy support will enable customers to define standard k8s policies. However, Calico also allows custom extensions (examples). For that, customer may consider installing the CLI.

docs/addons/calico.md
```diff
@@ -0,0 +1,190 @@
...
+64 bytes from 10.0.184.139: icmp_seq=3 ttl=254 time=0.073 ms
+```
+### Applying Policies using Calico to block traffic
+
+One of the powerful things about Calico is the ability to apply a Global Network Policy - policy across all namespaces.
+
+Let's apply the following Global Network Policy:
+
+```yaml
+apiVersion: projectcalico.org/v3
```
shapirov103, line 99: This requires CRD, correct? We may need an example with vanilla network policy.

shapirov103 (commented):
Looks good overall, it makes sense to decouple sections on standard network policies (could be applied globally or per namespace) and CRD driven network policies.

docs/addons/calico.md
youngjeong46: blurb added.

docs/addons/calico.md
youngjeong46: Revised.

docs/addons/calico.md
youngjeong46: Are we going to show both examples then? Currently there are two CRDs shown in the example:
- One that blocks all the resource communication other than kube-system ns
- One that will allow ping connection from team-riker to team-burnham

I can probably write one of them to use a vanilla network policy. I'm just a bit confused as the idea of vanilla network policy was going to be addressed in the other Issue (#31 )

shapirov103: The reason why we use Calico is not because we want to promote their specific features, but just for regular network policy support which is not available with vpc-cni. Atm, if customer does not use Calico addon, there will be no network policy support at any level. 

#31 is addressing example  in code for our SSP patterns repository which will also have documentation in the SSP patterns describing it.

For this documentation task it makes sense to call out that generic network policy can also be enabled with any other alternate CNI providers (which we have none atm) while global network policy binds customers to Calico.

docs/addons/calico.md
```diff
@@ -0,0 +1,197 @@
...
+        namespaceSelector: name == 'team-riker'
+      icmp:
+        type: 8
+
+```
+
+Apply the new GlobalNetworkPolicy:
+
+```bash
+calicoctl apply -f calico-allow-burnham-riker.yaml
```
shapirov103, line 194: Since this resource is using CRD, do we really need calicoctl to apply this to the cluster? What happens if we just use `kubectl apply -f` in this case?

shapirov103 (commented):
a few small comments

docs/addons/calico.md
youngjeong46: Understood. I've included doc changes to call out the distinction between k8s network policy & CRDs and replaced CRDs with k8s in the example. Also included the reason we use Calico

docs/addons/calico.md
youngjeong46: replaced this with k8s network policy. It uses kubectl now.

docs/addons/calico.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Pod to Pod communications with no policies
+
+If you deploy [App of Apps installed using ArgoCD](https://github.com/aws-quickstart/quickstart-ssp-amazon-eks/blob/feature/calico/docs/getting-started.md#deploy-workloads-with-argocd), you can verify that there are no network policies in place:
+
+```bash
+kubectl get networkpolicy -A
+```
+
+This means that any resources within the cluster should be able to make ingress and egress connections with other resources within and outside the cluster. You can verify, for example, that you are able to ping from `team-riker` pod to `team-burnham` pod.
```
shapirov103, line 44: Based on the security groups defined: 1/ node level 2/ explicit security groups that customers can associate with the workloads.

shapirov103 (approved):

docs/addons/calico.md
shapirov103: Overall, looks very good, approving.

PR #90, opened 2021-06-15
Including non-ts files in ts output

kcoleman731:
*Description of changes:*

Ensures that non-ts files are included in ts output.


package.json.orig
```diff
@@ -0,0 +1,57 @@
+{
+    "name": "@shapirov/cdk-eks-blueprint",
+<<<<<<< HEAD
+    "version": "0.1.8",
+    "main": "dist/lib/index.js",
+    "types": "dist/lib/index.d.ts",
+    "bin": {
+        "main": "dist/bin/main.js"
+    },
+=======
```
shapirov103, line 10: Let's drop this file, it appears to be redundant.

shapirov103 (commented):
Let's drop the .orig file from the code base.

shapirov103 (approved):

PR #94, opened 2021-06-17
Removing sample dirs and updating bin/main

kcoleman731:
*Description of changes:*

Removes all example dirs except one. Updates `bin/main` to call the remaining sample. Small readme updates as well. 


shapirov103 (approved):
lgtm

Issue #93, opened 2021-06-16
Add SSM Agent Addon 


kcoleman731:
Add support for SSM agent via an Addon 

https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/install-ssm-agent-on-amazon-eks-worker-nodes-by-using-kubernetes-daemonset.html

PR #99, opened 2021-06-18
Feature/ssm agent add on

askulkarni2:
*Issue #, if available:* #93

*Description of changes:*

Installs SSM Agent on worker nodes using DaemonSet.

Before deployment:

```sh
aws ssm start-session --target i-06bbfe4fdeb6839d3

An error occurred (TargetNotConnected) when calling the StartSession operation: <instance-id> is not connected.
```

After deployment:
```sh
aws ssm start-session --target i-06bbfe4fdeb6839d3                             

Starting session with SessionId: kuapoorv-Isengard-0622c845c3ffc12e1
sh-4.2$ 
```


docs/addons/ssm-agent.md
```diff
@@ -0,0 +1,96 @@
...
+
+This add-on uses the Kubernetes DaemonSet resource type to install AWS Systems Manager Agent (SSM Agent) on all worker nodes, instead of installing it manually or replacing the Amazon Machine Image (AMI) for the nodes. DaemonSet uses a CronJob on the worker node to schedule the installation of SSM Agent.
+
+A common use-case for installing SSM Agent on the worker nodes is to be able open a terminal session on an instance without the need to create a bastion instance and without having to install SSH keys on the worker nodes.
+
+The AWS Identity and Access Management (IAM) managed role **AmazonSSMManagedInstanceCore** provides the required permissions for SSM Agent to run on EC2 instances. This role is automatically attached to the instances when this add-on is enabled.
+
+**Limitations**
+
+* This add-on isn't applicable to AWS Fargate, because DaemonSets aren't supported on the Fargate platform.
```
shapirov103, line 11: Just a reminder for myself: we need a way to mark add-ons as incompatible with cluster providers, e.g. this one should be incompatible with Fargate / Bottlerocket.

docs/addons/ssm-agent.md
```diff
@@ -0,0 +1,96 @@
...
+aws ssm start-session --target $instance_id
+```
+
+### Use Case: Private Clusters
+
+If you are disabling public access for your EKS cluster endpoint such that the cluster endpoint is provisioned as private only i.e `endpointPublicAccess=false` and `endpointPrivateAccess=true`, then you can use one of the worker nodes as a TCP jump box to your EKS cluster api.
+
+To set up a TCP tunnel with your worker node as a jump box:
+
+1. Use SSM `send-command` api to create a TCP tunnel to Cluster API using `socat`:
```
shapirov103, line 57: assuming standard EKS AMI. socat may not be available for custom AMIs, we should call it out.

lib/addons/ssm-agent/index.ts
```diff
@@ -0,0 +1,87 @@
...
+                    spec: {
+                        containers: [
+                            {
+                                image: "amazonlinux",
+                                imagePullPolicy: "Always",
+                                name: "ssm",
+                                command: ["/bin/bash"],
+                                args: [
+                                    "-c",
+                                    "echo '* * * * * root yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm & rm -rf /etc/cron.d/ssmstart' > /etc/cron.d/ssmstart"
```
shapirov103, line 39: OMG, installs on every run? Isn't there an image with SSM pre-installed? This mutation won't work for any compliance programs, i am afraid. I am curious when eksctl is used with ssm = true, are they doing the same? (I realize that this is per documentation).

shapirov103 (commented):
Overall, looks good and thank you for this. There are a couple of comments, one of them mainly research if there is an immutable image for ssm.

docs/addons/ssm-agent.md
askulkarni2: Added callout in Limitations

lib/addons/ssm-agent/index.ts
askulkarni2: As per https://github.com/awslabs/amazon-eks-ami/issues/404 and https://github.com/awslabs/amazon-eks-ami/issues/127, SSM agent is not considered required for EKS and hence not added in the AMI.

shapirov103: I was thinking more along the lines of an AWS authored container image with SSM preinstalled. It would be nice to have it in AMI, but we also custom AMIs which won't have it.

shapirov103: A few things before we close it out:

1. Let's use https://github.com/jicowan/ssm-agent to build the image and upload it to a public ECR.
2. We need public ECR repo for a number of things including bootstrapping so let's create a task for that (@kcoleman731 wdyt?)
3. Let's upvote the following on GitHub: https://github.com/aws/amazon-ssm-agent/issues/135 (@parkand1 @youngjeong46 please join in).

askulkarni2: @shapirov103 added $1 from above comment. Will need $2 (cc: @kcoleman731)

shapirov103: Just making sure, as it stands it will work by pulling public image from your repository, correct? I am going to merge it right away to test if that is the case.

askulkarni2: That's correct. Any one can docker pull that image.

lib/addons/ssm-agent/ssm-install.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+echo '* * * * * root yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm && rm -rf /etc/cron.d/ssmstart' > /etc/cron.d/ssmstart
```
shapirov103, line 3: Can we point to an explicit version of the agent and tag the image with that version, so that we could have deterministic upgrades? The use case that I have in mind: we will upgrade the agent and publish it implicitly as latest. Customers will pick it up immediately so that in the existing cluster the SSP will be running a different version of the agent with the same SSP version. It is good when the agent works flawlessly, but it is great if one of the versions fails for any reason and we can't reproduce the state.

shapirov103 (commented):
A couple of easily addressable comments. The version/tag one is important IMO. I will merge right after, I realized that I need a private cluster to really test it out.

docs/addons/ssm-agent.md
```diff
@@ -0,0 +1,97 @@
...
+CLUSTER_API=$(aws eks describe-cluster --name $CLUSTER_NAME | jq -r '.cluster.endpoint' | awk -F/ '{print $3}')
+
+aws ssm send-command \
+  --instance-ids $instance_id \
+  --document-name "AWS-RunShellScript" \
+  --comment "tcp tunnel to cluster api" \
+  --parameters commands="nohup sudo socat TCP-LISTEN:443\,fork TCP:$CLUSTER_API:443 &"
+```
+
+2. Update `~/.kube/config` to use port 8443 instead of 443 as your local host may not allow you to bind port 443 (my machine does not allow me to bind to 443)
```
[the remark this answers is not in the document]
askulkarni2: done.

askulkarni2:
Found a `CrashLoopBackOff` issue. I am preparing a fix for it.

docs/addons/ssm-agent.md
kcoleman731: Ticket here. https://github.com/aws-quickstart/quickstart-ssp-amazon-eks/issues/134

docs/addons/ssm-agent.md
```diff
@@ -0,0 +1,97 @@
...
+1. Get the EC2 Instance Id of a worker node
+
+```bash
+instance_id=$(kubectl get nodes -o custom-columns=NAME:.metadata.name,INSTANCEID:.spec.providerID | awk -F/ 'FNR == 2 {print $5}')
+```
+
+2. Use the `start-session` api to see if you can open a terminal into the instance
+
+```bash
+aws ssm start-session --target $instance_id
```
kcoleman731, line 48: @askulkarni2 during validation, I ran into an issue where I didn't have the `Session Manager plugin for the AWS CLI installed`. We should add that as a step here. 

https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html

kcoleman731 (approved):
@shapirov103 validated this works like a charm. Nice work @askulkarni2 

I left one comment re the Session Manager plugin needing to be installed. Please add that and we will be good to merge.

docs/addons/ssm-agent.md
askulkarni2: addressed

lib/addons/ssm-agent/index.ts
```diff
@@ -0,0 +1,99 @@
...
+            metadata: {
+                name: "ssm-installer",
+            },
+            spec: {
+                selector: { matchLabels: appLabel },
+                updateStrategy: { type: "RollingUpdate" },
+                template: {
+                    metadata: { labels: appLabel },
+                    spec: {
+                        containers: [
```
askulkarni2, line 32: Added a pause container to avoid `CrashLoopBackoff` errors.

lib/addons/ssm-agent/ssm-install.sh
askulkarni2: addressed

kcoleman731 (approved):

PR #118, opened 2021-06-29
Feature/xray

shapirov103:
*Issue #, if available:*

*Description of changes:*
App Mesh fixed (did not work originally for practical cases)
App Mesh integration with X-Ray
XRay Addon
Documentation changes


docs/addons/app-mesh.md
```diff
@@ -21,9 +21,67 @@ new EksBlueprint(app, 'my-stack-name', addOns, [], {
...
+```
+## Tracing Integration
+
+App Mesh integrates with a number of tracing providers for distributed tracing support. At the moment it supports AWS X-Ray, Jaeger and Datadog providers. 
+X-Ray integration at present requires either a managed node group or a self-managed auto-scaling group backed by EC2. Fargate is not supported. 
+
+Enabling integration:
+
+```typescript
+const appMeshAddOn = new ssp.AppMeshAddOn({enableTracing: true, tracingProvider: "x-ray"}),
```
kcoleman731, line 51: should make X-Ray an enum. We can add support for Jaeger and Datadog over time.

lib/addons/appmesh/index.ts
```diff
@@ -13,13 +50,19 @@ export class AppMeshAddOn implements ClusterAddOn {
...
-        const cloudMapPolicy = ManagedPolicy.fromAwsManagedPolicyName("AWSCloudMapFullAccess")
+        const cloudMapPolicy = ManagedPolicy.fromAwsManagedPolicyName("AWSCloudMapFullAccess");
         sa.role.addManagedPolicy(cloudMapPolicy);
 
         // App Mesh Full Access policy.
-        const appMeshPolicy = ManagedPolicy.fromAwsManagedPolicyName("AWSAppMeshFullAccess")
+        const appMeshPolicy = ManagedPolicy.fromAwsManagedPolicyName("AWSAppMeshFullAccess");
         sa.role.addManagedPolicy(appMeshPolicy);
 
+        if(this.appMeshOptions.enableTracing && this.appMeshOptions.tracingProvider === "x-ray") {
```
kcoleman731, line 60: x-ray an enum...then you can switch on these values.

lib/addons/xray/index.ts
```diff
@@ -0,0 +1,40 @@
+import { KubernetesManifest } from "@aws-cdk/aws-eks";
+import { ManagedPolicy } from "@aws-cdk/aws-iam";
+import { assertEC2NodeGroup } from "../../cluster-providers";
+import { ClusterAddOn, ClusterInfo } from "../../stacks/cluster-types";
+import { loadYaml, readYamlDocument } from "../../utils/yaml-utils";
+
+export class XrayAddOn implements ClusterAddOn {
```
kcoleman731, line 7: Doc bloc

lib/addons/xray/xray-ds.yaml
```diff
@@ -0,0 +1,121 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
```
kcoleman731, line 2: where did this come from? Is there a Helm chart we can use?

shapirov103: I used the one from the EKS Workshop: https://www.eksworkshop.com/intermediate/245_x-ray/x-ray-daemon/ 
Unfortunately, there is no helm chart. The approach in the official documents is even more horrible than the one used in the workshop since it is not using IRSA.

docs/addons/app-mesh.md
shapirov103: yep, makes sense. will address.

lib/addons/xray/index.ts
shapirov103: Will address

docs/addons/app-mesh.md
shapirov103: since it is a string enum and I need string values, I am going to just restrict the type. It works for switch as well and less verbose.

lib/addons/xray/index.ts
shapirov103: Well, it appears not a single addon has it atm.  Why are you singling me out :))?

shapirov103: done

lib/addons/appmesh/index.ts
shapirov103: I added type restrictions which is less verbose

kcoleman731 (approved):

Issue #32, opened 2021-05-25
ArgoCD Bootstrapping must be automatic


kcoleman731:
At present each provisioned cluster requires ArgoCD bootstrapped which requires the following:

1. Install ArgoCD CLI matching the version on the server
2. Obtain initial credentials to Argo (secret is created in the target namespace)
3. Using the credentials, log in to Argo server and change the initial password.
4. Bootstrap the app of apps with specific parameters (e.g. we could pass param for kustomize). 

Suggested approaches:

1. Kubernetes Job that is deployed automatically with each cluster. Configured through the blueprint to have the uber app git URL and credentials (potentially through Sercret Manager). Runs once, warning on fail and ability to restart the job if needed. Job image will have kubectl and argocd CLI preinstalled.
2. CodeBuild project that performs the above, similar functional requirements. 
3. Lambda function (as customer CFN resource) that performs the above. 

Drawbacks for #2 and #3: requires download of argoCLI at runtime as well as access to K8S APi server for initial secret lookup.

kcoleman731:
Merged in #148

PR #148, opened 2021-08-02
Feature/argocd bootstrap

shapirov103:
*Issue #, if available:* #32 

*Description of changes:*  
Support for automatic workload bootstrapping with ArgoCD
Support for private repos with basic, SSH and token authentication.
Support for initial admin password with AWS Secrets Manager integration.
General refactoring to allow async provisioning add-ons that return a promise. 
Documentation update. 


docs/addons/argo-cd.md
```diff
@@ -23,8 +25,92 @@ new EksBlueprint(app, 'my-stack-name', addOns, [], {
...
+
+Inside ArgoCD, the admin password is stored as a `bcrypt` hash. This step will be performed by the framework and stored in the ArgoCD admin `secret`. 
+
+You can change the admin password through the Secrets Manager, but it will require rerunning the provisioning pipeline. Automatic sync with the Secrets Manager will be added in the future. 
+
+## Bootstrapping 
+
+The SSP framework provides an approach to bootstrap workloads and/or additional add-ons from a customer GitOps repository. In a general case, the bootstrap GitOps repository may contains an [App of Apps](https://argoproj.github.io/argo-cd/operator-manual/cluster-bootstrapping/#app-of-apps-pattern) that points to all workloads and add-ons.  
+
+In order to enable bootstrapping, the add-on allows passing an `ArgoApplicationRepository` at construction time. The following repository types are supported at present:
```
kcoleman731, line 49: do we want to make this more generic? Will Flux need repo support eventually?

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
 import { ClusterAddOn, ClusterInfo, ClusterPostDeploy } from "../../stacks/cluster-types";
 import { Team } from "../../teams";
+import * as yaml from 'yaml';
+import { btoa } from '../../utils/string-utils';
+import { Construct } from "@aws-cdk/core";
+
+import bcrypt = require('bcrypt');
+
+
+export interface ArgoApplicationRepository {
```
kcoleman731, line 14: might want to move this to its own file to decouple `repo` concept from `Argo`

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+     * Namespace where add-on will be deployed. 
+     */
+    namespace?: string,
+
+    /**
+     * If provided, the addon will bootstrap the app or apps in the provided repository.
+     * In general, the repo is expected to have the app of apps, which can enable to bootstrap all workloads,
+     * after the infrastructure and team provisioning is complete. 
+     */
+    bootstrapRepo?: ArgoApplicationRepository,
```
kcoleman731, line 61: should we make this an array? What if we want to have one repo for configuration and another for applications?

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+    namespace: "argocd"
+}
+
+/**
+ * Local function to create a secret reference for SSH key.
+ * @param url 
+ * @param secretName 
+ * @returns 
+ */
+const sshRepoRef = (url: string, secretName: string): string => yaml.stringify(
```
kcoleman731, line 90: move these to utils file in the same dir?

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+            throw error;
+        }
+    }
+
+    /**
+     * Creates a service account that can access secrets
+     * @param clusterInfo 
+     * @returns 
+     */
+    protected createServiceAccount(clusterInfo: ClusterInfo): ServiceAccount {
```
kcoleman731, line 349: seems like this is a utils function other addons would likely be able to use?

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+        return result;
+    }
+
+    /**
+     * Gets secret value from AWS Secret Manager. Requires access rights to the secret, specified by the secretName parameter.
+     * @param secretName name of the secret to retrieve
+     * @param region 
+     * @returns 
+     */
+    protected async getSecretValue(secretName: string, region: string): Promise<string> {
```
kcoleman731, line 324: Same comment as below. Like a function other addons could leverage. Not specific to argo.

kcoleman731 (commented):
Looking awesome.

lib/addons/argocd/index.ts
shapirov103: I think it is a good idea, although who knows how they could diverge, e.g. options for Argo vs Flux. At present wego defined their own structure for it, so a bit late for convergence. However, agree that it is a generic structure. Will address.

lib/addons/argocd/index.ts
shapirov103: I think it is a good idea, although who knows how they could diverge, e.g. options for Argo vs Flux. At present wego defined their own structure for it, so a bit late for convergence. However, agree that it is a generic structure. Will address.

lib/addons/argocd/index.ts
shapirov103: It is an app of apps, so it enables all sorts of combinations. They can have a separate repo for applications, separate repo (or multiple) for plugins and just reference them in the app of apps. We could make it an array but it sort opens up non-GitOps onboarding, e.g. by modifying the code and adding another repo. I think we should force them to use app of apps properly.

lib/addons/argocd/index.ts
shapirov103: This SA is specific to secret manager and it has a pretty broad permission set. Other service accounts that we generate (e.g. for Apoorva's CSI / secrets addon) are specific to a resource. But if we move out the policy out of this method, there is not much left compared to `cluster.addServiceAccount`.

lib/addons/argocd/index.ts
shapirov103: sure

docs/addons/argo-cd.md
shapirov103: We could do that, but auth methods are limited to those supported by Argo atm.

lib/addons/argocd/index.ts
shapirov103: I will move it out to something like secret-utils under utils.

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+            manifest: [{
+                apiVersion: "argoproj.io/v1alpha1",
+                kind: "Application",
+                metadata: {
+                    name: appRepo.name ?? "bootstrap-apps",
+                    namespace: this.options.namespace
+                },
+                spec: {
+                    destination: {
+                        namespace: "default",
```
askulkarni2, line 204: Should this be `this.options.namespace`?

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -60,12 +60,15 @@ export class EksBlueprint extends cdk.Stack {
...
         const postDeploymentSteps = Array<ClusterPostDeploy>();
+        const promises = Array<Promise<any>>();
 
         for (let addOn of (blueprintProps.addOns ?? [])) { // must iterate in the strict order
-            addOn.deploy(clusterInfo);
-            const postDeploy : any  = addOn;
+            const result : any = addOn.deploy(clusterInfo);
+            if(result) {
+                promises.push(<Promise<any>>result);
+            }
```
askulkarni2, line 70: 👍🏽

askulkarni2 (commented):
Minor comments, but otherwise lgtm!

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
+                secret: {
+                    argocdServerAdminPassword: adminSecret
+                }
+            }
+        }
+
+        this.chartNode = clusterInfo.cluster.addHelmChart("argocd-addon", {
             chart: "argo-cd",
-            
+            release: "ssp-addon",
```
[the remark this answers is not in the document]
shapirov103: Yes, I proposed we do. Many helm charts will add the release in the pod/svc/deployment names to allow for more than one installation. However, it also results in non-deterministic names and that makes any post installation instructions moot, e.g. find deployment that "looks like <something>-argocd-server" and such.
I have created a constant for that, updated code with it.

lib/addons/argocd/index.ts
```diff
@@ -1,26 +1,359 @@
...
             repository: "https://argoproj.github.io/argo-helm",
-            version: '3.2.3',
-            namespace: this.namespace,
+            version: '3.10.0',
```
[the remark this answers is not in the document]
shapirov103: This applies across the board. If we don't test with another version, chances are, our logic won't work because of argo changes. It also makes it less supportable. Hence, fixed version of the chart is mandatory across the board, if we shoot for a managed solution. Let's discuss if needed.

lib/addons/argocd/index.ts
shapirov103: Good catch. Apparently it does not affect anything for the app of apps, since there is no real deployable. Each app in the GitOps repo defines its own namespace, so everything works. Will fix.

shapirov103 (commented):
I have addressed all comments, please take a look @kcoleman731 @askulkarni2 
All good feedback.

lib/addons/argocd/index.ts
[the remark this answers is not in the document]
askulkarni2: That makes sense.

askulkarni2 (approved):
lgtm!

docs/addons/argo-cd.md
kcoleman731: gotcha

lib/addons/argocd/index.ts
kcoleman731: sort of...I thinking of an example where you have sep repos for applications and configuration.

kcoleman731 (approved):

PR #155, opened 2021-08-06
pr bot

jaymccon:
pr bot automation has been enabled. This pr makes small changes to facilitate that:

* buildspec updated to build source based off pr if it was invoked by the pr bot
* buildspec to bootstrap cdk if needed
* add CODEOWNERS so that the bot user is requested for review

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @aws-quickstart/ssp-eks-internal @aws-ia-ci
--- a/ci/buildspec.yml
+++ b/ci/buildspec.yml
@@ -7,14 +7,20 @@ phases:
     runtime-versions:
       nodejs: 14
     commands:
+      - if [ ! -z "${COMMIT_ID}" ]; then git checkout main; git merge --no-edit ${COMMIT_ID}; fi
+      - | 
+        npm install aws-cdk@1.113.0 -g
+        npm install
+      - >
+        # if there's no CDKToolkit stack then bootstrap cdk
+        aws cloudformation describe-stacks --stack-name CDKToolkit --region ${AWS_REGION} 
+        || CDK_NEW_BOOTSTRAP=1 cdk bootstrap aws://$(aws sts get-caller-identity --output text --query Account --region ${AWS_REGION})/${AWS_REGION}
   pre_build:
     commands:
       - >
         [ -z "$CONTEXT_LOCATION" ] || aws s3 cp $CONTEXT_LOCATION .
   build:
     commands:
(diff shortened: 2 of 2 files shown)


shapirov103 (approved):
Looks good and thank you for your contribution.

A small comment (note for discussion):
At present, we reflect the CDK version right in the buildspec.yml. However, it is something that changes every week. Ideally, the version of CDK should be part of the repo. Otherwise, the way it stands now, we will need a process to apply changes to this file and deploy it on the CI account.

jaymccon:
Ideally whatever we need to do on the ci account should be encapsulated in the buildspec. I'd propose one of 2 approaches:

* run bootstrap on every ci run, which should update to whatever version of cdk is declared in the buildspec.
* completely remove cdk bootstrapped things from the account at the end of each ci run using the `finally` block

For both options we should fetch the version number from package.json instead of hardcoding it in the buildspec

shapirov103:
@askulkarni2 please see ^^. The spec should be self-sufficient. I assume bootstrap can run on demand, however CDK version should be installed based on the version file in the source, otherwise it will stop working the minute we upgrade.

askulkarni2:
In that case, should `aws-cdk` itself be in package.json? `npm install` would be enough in that case.

shapirov103:
It is enough to install but the install is performed not in global scope. Will `cdk` command work after a local install? If not, what would be the alternative of running it if it is installed locally? @askulkarni2

askulkarni2:
@shapirov103 from https://docs.aws.amazon.com/cdk/latest/guide/cli.html

>If you regularly work with multiple versions of the AWS CDK, you may want to install a matching version of the AWS CDK Toolkit in individual CDK projects. To do this, omit -g from the npm install command. Then use npx cdk to invoke it; this will run the local version if one exists, falling back to a global version if not.

shapirov103:
That makes sense, let's review the user experience in this case. 

I propose we do the following across the board:

1. Add CDK version in package.json
2. Modify the existing make command to include a `cdk` target as `npx cdk` and use it in all CI scripts and for testing.
3. Optionally supply an .sh file in the ssp-patterns repo, such as `ssp.sh` that will invoke `make cdk`. You could then run `ssp list` and such. Users could copy to ~/bin as well.
4. Rely in npm i for all installations and modify the `make` to  install automatically if fresh repo. In this case running `make cdk list` or `make cdk deploy` will take care of that step not only for CI but also for the customers, reming 95% of issues I observed so far. 

@askulkarni2 @kcoleman731

Issue #65, opened 2021-05-25
Elastic Search Addon


kcoleman731:
Add support for Managed Elasticsearch

Issue #60, opened 2021-05-25
[ADDON] - Fluent Bit 


kcoleman731:

elamaran11:
We support Fluent Bit Addon today. Closing this issue.

PR #157, opened 2021-08-11
Feature/65 managed elasticsearch support

kcoleman731:
*Issue #, if available:*
#65 #168 #60 

*Description of changes:*

This pull requests add support for a `FluentBit` add-on and an `Elasticsearch` stack. These two components work in concert to enable Elasticsearch and Kibana support for capturing and viewing logs within the context of an SSP deployment. 

For the time being, the `Elasticsearch` stack only supports [`Public access`](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/fgac.html#fgac-access-policies). 


docs/addons/managed-elasticsearch.md
```diff
@@ -0,0 +1,20 @@
+# AWS Managed Elasticsearch
```
askulkarni2, line 1: Probably a short description of what's being done here.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        // Configure helm chart and deploy.
+        if (this.props?.elasticsearch) {
+            this.applyForElasticsearch(cluster, namespace, sa)
+        } else {
+            this.applyHelmChart(cluster, namespace)
+        }
+    }
+
+    postDeploy(clusterInfo: ClusterInfo, teams: Team[]): void {
```
askulkarni2, line 78: Throw "Not Implemented" error or remove it.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,133 @@
...
+         * Allows the SA to make requests to the Elasticsearch domain. 
+         */
+        const domainProps = this.props?.elasticsearch
+        const domainArn = domainProps?.domainArn!
+        const policy = new PolicyStatement({
+            effect: Effect.ALLOW,
+            actions: ['es:ESHttp*'],
+            resources: [domainArn]
+        })
+        serviceAccount.addToPrincipalPolicy(policy);
```
askulkarni2, line 100: Maybe do lookup using the `elasticsearch.Domain.fromDomainEndpoint()` method. Then you can do something like `domain.grantWrite(serviceAccount)` instead of hand-crafting policy statements.

askulkarni2 (commented):
First pass, added some minor comments. Will review again for future iterations.

lib/addons/argocd/index.ts
```diff
@@ -1,14 +1,13 @@
...
 import { ManagedPolicy } from "@aws-cdk/aws-iam";
-import { ApplicationRepository, ClusterAddOn, ClusterPostDeploy, ClusterInfo, Team } from "../../spi";
 
 import { getSecretValue } from '../../utils/secrets-manager-utils';
 import { sshRepoRef, userNameRepoRef } from './manifest-utils';
 import { btoa } from '../../utils/string-utils';
-import bcrypt = require('bcrypt');
 import { Constants } from "..";
-
+import * as spi from "../../spi";
```
shapirov103, line 10: Minor: What is the incentive to use wildcard import with prefix vs clear imports? Is it a number of objects in import? Namespacing?
I was going with 5 imports before using wildcards.

Let's agree and make a lint rule for this.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * The namespace for the FluentBit add-on.
+     * @default `logging`
+     */
+    readonly namespace: string;
+
+    /**
+     * Props for Elasticsearch support.
+     */
+    readonly elasticsearch?: es.DomainAttributes
```
shapirov103, line 24: What other destinations can we support (if any)? If unknown, may be allowing a free form configuration to be passed?

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export class FluentBitAddOn implements ClusterAddOn {
+
+    readonly props: FluentBitAddOnProps
+
+    constructor(props?: FluentBitAddOnProps) {
+        this.props = { ...defaultProps, ...props }
+    }
+
+    deploy(clusterInfo: ClusterInfo): void {
```
shapirov103, line 46: May be return a promise here, if you think any post deploy steps will depend on it.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,117 @@
...
+                name: serviceAccount.serviceAccountName,
+                create: false
+            },
+            elasticsearch: {
+                enabled: true,
+                host: domainEndpoint,
+                awsRegion: cluster.stack.region
+            }
+        }
+        this.applyHelmChart(cluster, namespace, values)
```
shapirov103, line 101: I am pretty sure that it will be a race condition between SA and the helm chart. You need to add `helmChart.addDependency(sa)`.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    deploy(clusterInfo: ClusterInfo): void {
+        // Create the FluentBit namespace.
+        const cluster = clusterInfo.cluster;
+        const namespace = this.props.namespace
+        createNamespace(cluster, namespace)
+
+        // Create the FluentBut service account.
+        const serviceAccountName = 'fluent-bit-service-account'
+        const sa = cluster.addServiceAccount(serviceAccountName, {
```
shapirov103, line 54: Need a dependency between sa and namespace, otherwise intermittent failures on race conditions.

lib/addons/fluent-bit/index.ts
```diff
@@ -0,0 +1,117 @@
...
+import { Constants } from "..";
+import { ClusterInfo } from "../../spi/types";
+import { Team } from "../../spi/team-contracts";
+import { ClusterAddOn } from "../../spi/addon-contracts"
+import { createNamespace } from "../../utils/namespace";
+
+/**
+ * Configuration options for the FluentBit add-on.
+ */
+export interface FluentBitAddOnProps {
```
shapirov103, line 14: Thinking more about the design, it sounds like the plugin could support not just other destinations, but also multiple destinations at the same time. So purely for design perspective we could create an interface of destination providers and allow customers to pass those in.

It also sounds like the service account should be created by us upfront and each destination may add specific policies to it. 

this will make it more complex, while usage convenience will remain roughly the same. Could be deferred for pragmatic reasons, but sounds like a one way door once it is out there.

lib/stacks/elasticsearch-stack.ts
```diff
@@ -0,0 +1,130 @@
+import * as cdk from '@aws-cdk/core';
```
shapirov103, line 1: Do you propose we add such stacks to quick-start as opposed to patterns? 

It is surely a nice and handy stack. 

It is a more conceptual question: if customers are expected to use it as an example, e.g. clone and customize, then it should be in patterns. 

I am afraid it may be a too much for our team to incorporate quick starts for other managed services in here, we barely have resources for cluster level as it stands. Going down this path, we will have to provide stacks for AMP/AMG as well as RDS, Dynamo, etc. 

I see the value for the platform though, let's discuss.

shapirov103 (commented):
Looks nice!

lib/addons/argocd/index.ts
kcoleman731: lol I had the same comment on your pr today let's agree tomorrow.

lib/stacks/elasticsearch-stack.ts
kcoleman731: Sounds great. Let's discuss.

lib/stacks/elasticsearch-stack.ts
```diff
@@ -0,0 +1,130 @@
...
+
+        /**
+         * Deploy the Elasicsearch domain.
+         */
+        const version = this.props.version!
+        const masterUserName = this.props.adminUsername
+        const adminPasswordSecretKey = this.props.adminPasswordSecretKey!
+        const masterUserPassword = cdk.SecretValue.secretsManager(adminPasswordSecretKey)
+
+        const defaultDomainProps = {
```
shapirov103, line 97: It would be nice to move these out of the stack construct and make it discoverable through the npm package (e.g. export this const) so that customers are aware of the defaults.

kcoleman731:
Closing for the time being.

Issue #152, opened 2021-08-06
[BUG] - Pipeline stack Error


allamand:
Trying to follow the pipeline stack from default but I had several errors : 

1. npm ci error
  ```
  npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.
  ```
  
  I fixed this by commiting the `package-lock.json` file

2. error on cdk synth

  ```
  [Container] 2021/08/06 14:05:29 Running command npx cdk synth
  current credentials could not be used to assume 'arn:aws:iam::xxxxxx:role/cdk-hnb659fds-lookup-role-xxxxxx-eu-west-1', but are for the right account. Proceeding anyway.
  [Error at /blueprint-construct-dev] You are not authorized to perform this operation.
  Found errors
  ```
  
  I fied this by commiting my cdk.context.json file

3. the Assets jobs failed

  with 1. and 2. fix, the build was ok, but it fails now at the Assets stage:
  
  ![image](https://user-images.githubusercontent.com/8956643/128528201-dd9a6051-c423-4dbe-80bb-b64e29ea06c6.png)

  the error message is
  ```
  Error: Cannot read asset manifest at 'assembly-pipeline-blueprint-dev/pipelineblueprintdeveks1AFA1B3B.assets.json': ENOENT: no such file or directory, stat 'assembly-pipeline-blueprint-dev/pipelineblueprintdeveks1AFA1B3B.assets.json'
  ```

I'm not sur how to recover from the last one, i was testing in eu-west-1

kcoleman731:
Thanks for reporting this @allamand. Let us look into this and get back to you.

shapirov103:
This is fixed by https://github.com/aws-quickstart/ssp-amazon-eks/pull/184 and https://github.com/aws-samples/ssp-eks-patterns/pull/2. 

You will need to pull the latest from https://github.com/aws-samples/ssp-eks-patterns and follow the example in the `lib/pipeline-stack/index.ts`.

At present, it requires having AWS credentials configured on AWS CLI (e.g. `aws configure`). 
<img width="1249" alt="Screen Shot 2021-08-24 at 3 56 38 PM" src="https://user-images.githubusercontent.com/60622975/130682185-565d6398-dd18-45a8-a452-68b9a85a8bee.png">

shapirov103:
I will keep it open for a bit in case you have any feedback.

Issue #149, opened 2021-08-02
[TASK] - Add QuickStart Identifiers


shapirov103:
As discussed here are the details for gethering launch metrics:

Each template that you want to track must have a qs-id at the end of the Description field, enclosed in (). Example description: https://github.com/aws-quickstart/quickstart-amazon-eks/blob/main/templates/amazon-eks-metrics-server.template.yaml#L5

for each template you add an id to, you will need to provide me with:
* id
* name (friendly name for quicksight reporting , eg.: flux-addon, nodegroup-stack, etc)
* count (True/False) should we treat launches of this template as a customer launch. Typically only one id has htis set to true, like a templates that is always deployed regardless of how the customer configured options before synth/create.

PR #184, opened 2021-08-17
Feature/usage tracking, blueprint and pipeline builders, tests 

shapirov103:
*Issue #, if available:* #149
Fix for #152 (validated with version 0.11..4-test of the CDK blueprint)

*Description of changes:*
Feature/usage tracking, blueprint and pipeline builders, tests for usage and builders, CDK moved to 1.119 

The PR for patterns is coming (pipelines and builders). 

GitHub actions to add tests will also be added. 


Makefile
```diff
@@ -23,6 +23,9 @@ list:
 	$(DEPS)
 	$(CDK) list
 
+run-test:
```
shapirov103, line 26: Might be easier to follow if you start with test/stacks.test.ts: https://github.com/aws-quickstart/ssp-amazon-eks/pull/184/files#diff-59f0283006bf485e281b97535ff10c05afe6253b30c4ab3b6b298851ffd3579b

kcoleman731: tests in this project are a wonderful thing. Thank you.

lib/pipelines/code-pipeline.ts
```diff
@@ -2,57 +2,166 @@ import * as cdk from '@aws-cdk/core';
 import * as pipelines from '@aws-cdk/pipelines';
 import * as codepipeline from '@aws-cdk/aws-codepipeline';
 import * as actions from '@aws-cdk/aws-codepipeline-actions';
+import { Construct, StackProps } from '@aws-cdk/core';
```
kcoleman731, line 5: we should get aligned on when to spread imports and when to use `import * as`. We are inconsistent in the project currently.

lib/pipelines/code-pipeline.ts
```diff
@@ -2,57 +2,166 @@ import * as cdk from '@aws-cdk/core';
...
+
+    private props: Partial<PipelineProps>;
 
-    public static build = (props: PipelineProps) => {
+
+    constructor() {
+        this.props = { stages: []};
+    }
+
+    public name(name: string): CodePipelineBuilder {
```
kcoleman731, line 74: are these meant to be setters? If so, consider renaming to `setName`. Also what does it return this?

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -38,23 +39,106 @@ export class EksBlueprintProps {
...
+        account?: string,
+        region?: string
+    }
+
+    constructor() {
+        this.props = { addOns: new Array<ClusterAddOn>(), teams: new Array<Team>()};
+        this.env = {};
+    }
+
+    public name(name: string) : this {
```
kcoleman731, line 69: same thing here...this pattern seems odd to me. Also returning this? Is that valid typescript?

kcoleman731 (commented):
@shapirov103 lets talk about where the builder pattern comes from? Is this a CDK pattern you have seen?

lib/pipelines/code-pipeline.ts
shapirov103: I mentioned elsewhere, I was using common conventions from my prev experience: wildcard after 5 imports.

lib/pipelines/code-pipeline.ts
shapirov103: They are not traditional setters. That the convention for the builder pattern. The difference is fluid API as each method returns `this`.  Alternative is adding `with` to each property but that seems overly verbose for little gain.

lib/stacks/eks-blueprint-stack.ts
shapirov103: I was using this for years with different programming languages. You can try `make build` `make run-test` `make lint` on the branch.

askulkarni2 (approved):
A couple of minor points but otherwise looks slick. Please kick off the e2e tests well when you are ready.

ci/buildspec.yml
```diff
@@ -18,7 +18,6 @@ phases:
         make deps
       - >
         make bootstrap-cdk
-
   pre_build:
     commands:
       - >
```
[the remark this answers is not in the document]
shapirov103: 100%

lib/pipelines/code-pipeline.ts
```diff
@@ -65,13 +174,13 @@ export class CodePipeline {
                 // Install deps
                 'npm install',
                 // Install global CDK.
-                'npm install -g aws-cdk@1.113.0'
+                'npm install -g aws-cdk@1.119.0'
             ],
             buildCommands: ['npm run build'],
             synthCommand: 'cdk synth'
```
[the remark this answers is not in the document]
shapirov103: Definitely. Will address.

lib/pipelines/code-pipeline.ts
kcoleman731: gotcha. My vote would be to use setter function names. Makes it explicit that you are "setting" something as opposed to reading. Up to you.

lib/pipelines/code-pipeline.ts
```diff
@@ -2,57 +2,166 @@ import * as cdk from '@aws-cdk/core';
...
+        return this;
+    }
+
+    public stage(stackStage: StackStage) : CodePipelineBuilder {
+        this.props.stages?.push(stackStage);
+        return this;
+    }
+    
+    build(scope: cdk.Construct, id: string, stackProps?: cdk.StackProps): cdk.Stack {
+        console.assert(this.props.name && this.props.owner && this.props.repository!.credentialsSecretName && this.props.stages,
```
kcoleman731, line 95: consider asserting on each individually. Helps be specific about which property is missing (and avoids a super long line of code!)

kcoleman731 (approved):

kcoleman731:
@shapirov103 meta comment here...we should think about how to organize stacks and files in each stack. For instance, do we want to move the pipeline dir into stacks and then have `pipeline` directory and a `cluster` directory. Within the pipeline directory perhaps we have a stack.ts, a builder.ts, and a stage.ts. Same thing with the cluster directory...perhaps we have a stacks.ts and a builder.ts?

Not married to an approach but something to consider.

lib/pipelines/code-pipeline.ts
shapirov103: Appears to be resolved, as the number of imports before wildcards is 3. We need to look into making it a lint rule. So far I only observe a way to code that rule, so deferring for now.

lib/pipelines/code-pipeline.ts
shapirov103: Builders are used to reduce verbosity as one of the objectives and extra prefixes are not quite helping. I am leaning towards keeping it as is, even though i understand your perspective.

shapirov103:
> @shapirov103 meta comment here...we should think about how to organize stacks and files in each stack. For instance, do we want to move the pipeline dir into stacks and then have `pipeline` directory and a `cluster` directory. Within the pipeline directory perhaps we have a stack.ts, a builder.ts, and a stage.ts. Same thing with the cluster directory...perhaps we have a stacks.ts and a builder.ts?
> 
> Not married to an approach but something to consider.

We can apply this, it is a bit more deterministic and consistent to what we have now. I will defer this one, probably a separate task or could be handled within your ES PR.

ci/buildspec.yml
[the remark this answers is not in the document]
shapirov103: Addressed

lib/pipelines/code-pipeline.ts
shapirov103: Addressed

shapirov103:
@kcoleman731 @askulkarni2 please take another look. 
Also the the PR to the patterns that was validated to work in my account (although against a branch). 
https://github.com/aws-samples/ssp-eks-patterns/pull/2

kcoleman731 (approved):

Issue #63, opened 2021-05-25
[ADDON] - Open Policy Agent


kcoleman731:
Add OPA addon with sensible defaults. 

https://github.com/jasonumiker/eks-quickstart/tree/main/gatekeeper-policies

parkand1:
OPA Gatekeeper add-on merged - closing

PR #188, opened 2021-08-30
Cluster Provider Cleanup and Docs

kcoleman731:
*Issue #, if available:*
#63 

*Description of changes:*
This PR updates the cluster provider classes and adds documentation for each. The classes have been renamed to reflect their functionality (i.e. EC2ClusterProvider -> MngClusterProvider). Props have been extended and documented for the classes as well. 

Note - after doing this cleanup work, there is a fair amount of overlap and duplicate code in the provider implementations. We should evaluate if it would be worth consolidating these classes into a single cluster class. Perhaps moving the implementation into the blueprint class itself.


shapirov103 (commented):
Looks nice! 
Was it tested/validated in any way?  We need a process here, otherwise we will be in a reactive mode. 

Why not include an instance of each provider in the e2e test?

shapirov103:
> > Looks nice!
> > Was it tested/validated in any way? We need a process here, otherwise we will be in a reactive mode.
> > Why not include an instance of each provider in the e2e test?
> 
> Manual validation. We need to talk about tests here. Our example lib is going to grow again...which makes our example lib look like the patterns lib.

I am less concerned with similarities between patterns and this repo, more with quality assurance. For quality, we have to do what is needed, and to be honest, it would be best to have a pipeline on patterns which is exactly how customers use it. As we discussed, let's create a ticket for this work and see if anyone has bandwidth to address.

shapirov103 (approved):
lgtm

Issue #183, opened 2021-08-17
[BUG] - ExternalDNS + Nginx deployment order


kcoleman731:
If Nginx is deployed after ExternalDNS, the NLB does not get deployed. 

If ArgoCD is deployed before either, the ingress do not create the relevant records in Route53.

shapirov103:
1. 'Nginx is deployed after ExternalDNS` - this has no impact on NLB. I assume you meant Nginx deployed before AWS Load Balancer controller. 
2. ArgoCD app creation is a post deployment hook and workloads are pulled in asynchronously through GitOps. I assume because NLB was not provisioned in the first place, nothing will work.

PR #195, opened 2021-09-08
Fix/183 external dns nginx deployment order

askulkarni2:
*Issue #, if available:* #183

*Description of changes:*
Adds a mechanism for handling inter addon dependencies such as between nginx and external-load-balancer add on. Now addons can depend on the promise of construct from the addon that its dependent on. The dependent addon can then wait for promise to be fulfilled and create its dependency as shown in the nginx example.

### DIFF
--- a/lib/addons/nginx/index.ts
+++ b/lib/addons/nginx/index.ts
@@ -81,12 +81,12 @@ export class NginxAddOn implements ClusterAddOn {
         const props = this.options;
 
         const dependencies = Array<Promise<Construct>>();
-        const awsLoadBalancerControllerAddOnPromise = clusterInfo.getpreProvisionedAddOn('AwsLoadBalancerControllerAddOn');
+        const awsLoadBalancerControllerAddOnPromise = clusterInfo.getScheduledAddOn('AwsLoadBalancerControllerAddOn');
         console.assert(awsLoadBalancerControllerAddOnPromise, 'NginxAddOn has a dependency on AwsLoadBalancerControllerAddOn');
         dependencies.push(awsLoadBalancerControllerAddOnPromise!);
 
         if (props.externalDnsHostname) {
-            const externalDnsAddOnPromise = clusterInfo.getpreProvisionedAddOn('ExternalDnsAddon');
+            const externalDnsAddOnPromise = clusterInfo.getScheduledAddOn('ExternalDnsAddon');
             console.assert(externalDnsAddOnPromise, 'NginxAddOn has a dependency on ExternalDnsAddOn');
             dependencies.push(externalDnsAddOnPromise!);
         }
--- a/lib/addons/nginx/index.ts
+++ b/lib/addons/nginx/index.ts
@@ -118,11 +118,10 @@ export class NginxAddOn implements ClusterAddOn {
             values
         });
 
-        const asyncTasks = Promise.all(dependencies.values()).then((constructs) => {
+        Promise.all(dependencies.values()).then((constructs) => {
             constructs.forEach((construct) => {
                 nginxHelmChart.node.addDependency(construct);
             });
-        });
-        asyncTasks.catch(err => { throw new Error(err) });
+        }).catch(err => { throw new Error(err) });
     }
 }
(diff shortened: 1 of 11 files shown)


lib/addons/nginx/index.ts
```diff
@@ -97,13 +109,20 @@ export class NginxAddOn implements ClusterAddOn {
...
+        const nginxHelmChart = clusterInfo.cluster.addHelmChart("nginx-addon", {
             chart: "nginx-ingress",
             repository: "https://helm.nginx.com/stable",
             release: Constants.SSP_ADDON,
             namespace: props.namespace,
             version: props.version,
             values
         });
+
+        const asyncTasks = Promise.all(dependencies.values()).then((constructs) => {
```
shapirov103, line 121: You don't need the const here, just `Promise.all(dependencies.values()).then(...).catch(err)`.

lib/addons/nginx/index.ts
```diff
@@ -97,13 +109,20 @@ export class NginxAddOn implements ClusterAddOn {
...
+            constructs.forEach((construct) => {
+                nginxHelmChart.node.addDependency(construct);
+            });
+        });
+        asyncTasks.catch(err => { throw new Error(err) });
```
shapirov103, line 126: Ideally, I would like to refactor to some common addon implementation where you can specify in constructor the ids of dependencies and it will manage it automatically, as opposed to spreading similar code across multiple add-ons. E.g base class `BasicClusterAddOn` has an array of dependency names and . I am sure this will be used not just by us. 

A better approach to explore is to use a decorator pattern like https://www.typescriptlang.org/docs/handbook/decorators.html. I have not used it myself with typescript yet (used a lot with java) but it can look something like:

```
 class MyAddOn implements ClusterAddOn {
     
     @dependable('AwsLoadBalancerController')
     deploy(clusterInfo:  ClusterInfo): Promise<> {
      .....
     }
     
 }
```

We can defer this as well as functionally it serves the puprose (e.g. create a task to address later on).

shapirov103 (commented):
Looks great and as test show, works as well 👍 
Some comments for you, mostly minor/deferrable.

lib/spi/types.ts
```diff
@@ -94,4 +96,23 @@ export class ClusterInfo {
...
         }
     }
+
+    /**
+     * Set the preProvisionedAddOn map with the promise for the construct
+     * of the addon being provisioned
+     * @param addOn
+     * @param promise
+     */
+     public addpreProvisionedAddOn(addOn: string, promise: Promise<cdk.Construct>) {
```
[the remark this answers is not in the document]
askulkarni2: `addScheduledAddOn` sounds great! Added.

lib/spi/types.ts
```diff
@@ -94,4 +96,23 @@ export class ClusterInfo {
...
+        this.preProvisionedAddOns.set(addOn, promise);
+    }
+
+    /**
+     * Returns the promise for the Addon construct
+     * @param addOn
+     * @returns Promise<cdk.Construct>
+     */
+    public getpreProvisionedAddOn(addOn: string): Promise<cdk.Construct> | undefined {
```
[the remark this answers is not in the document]
askulkarni2: Fixed.

lib/addons/nginx/index.ts
askulkarni2: Fixed.

lib/addons/nginx/index.ts
askulkarni2: I've thought about the decorator pattern here as well (used it extensively in Python) but I too am new to it within the context of typescript. I will create a refactor task for it so we can unblock other PRs for now.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -249,4 +252,13 @@ export class EksBlueprint extends cdk.Stack {
 
         return vpc;
     }
+
+    /**
+     * Returns AddOn Id if defined else returns the class name
+     * @param addOn 
+     * @returns 
+     */
+    private getAddOnNameorId(addOn: spi.ClusterAddOn): string {
```
shapirov103, line 261: Very minor:

1. `getAddOnNameOrId` - Or (capital)
2. It has not dependency on this, so does not have to be a member function. We can make it a regular function outside of Blueprint class and it will reduce verbosity.

askulkarni2: addressed, please review

shapirov103 (approved):
lgtm

Issue #143, opened 2021-07-24
[ADDON] - Velero 


starchx:
Often required by enterprise customer as per their compliance, we should support this backup addon:
https://github.com/vmware-tanzu/velero

kcoleman731:
Dupe. https://github.com/aws-quickstart/quickstart-ssp-amazon-eks/issues/125. 
But agree!

PR #205, opened 2021-09-19
[Feature/Velero] - Create the Velero AddsOn

haofeif:
*Description of changes:*
* This is to close the Issue #143 
* Create the Velero AddsOn
* By default create a private S3 bucket (blocking all public access) with SSE-KMS encryption with AWS Managed Key from KMS(Encryption At Rest) for the Velero Backup destination
* Create an S3 VPC endpoint to direct traffic to S3 within VPC
* Configure S3 Bucket policy to enable encryption in transit
* Create the IAM Role for Service Account for Velero pod to make API calls to AWS S3 and EC2 to backup and restore.
*  Preset [Velero Helm Chart Values](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml)
* Allow users to pass in with [Velero Helm Chart Values](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml) for customisation purposes.
* Complete the documentation and example with screenshots


lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+                    actions: ['s3:*'],
+                    resources: [bucket.arnForObjects('*')],
+                    principals: [new iam.AnyPrincipal()],
+                    conditions:{
+                        Bool:{
+                            'aws:SecureTransport': 'false',
+                        }
+                    }
+                })
+            );
```
askulkarni2, line 120: I think this can be replaced with `enforceSSL: true` prop in the Bucket() creation.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+                    principals: [new iam.AnyPrincipal()],
+                    conditions:{
+                        Bool:{
+                            'aws:SecureTransport': 'false',
+                        }
+                    }
+                })
+            );            
+            // Create S3 VPC Endpoint for the Velero pod to access S3 via VPC Endpoint instead of going to internet
+            cluster.vpc.addGatewayEndpoint('velero-backup-bucket-vpcEndPoint', {
```
askulkarni2, line 122: Lets think through if endpoint policy is a good idea or not. I would imagine enterprise customers would require the policies to be created. Thoughts?

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+                    kind: 'Namespace',
+                    metadata: { 
+                        name: props.namespace.name
+                    }
+                }
+                )
+                veleroNamespace = props.namespace.name;
+            }
+            else{
+                console.log ("namespace:" + props.namespace.name + " exists, not creating new namespace")
```
askulkarni2, line 148: We haven't been adding such console messages so this would stand out. Something for us to think about if we want to keep this and make it consistent across the board or remove it.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+            ]
+        };
+
+        const veleroCustomPolicyDocument = iam.PolicyDocument.fromJson(veleroPolicyDocument);
+        const veleroPolicy = new iam.ManagedPolicy(cluster, "velero-managed-policy", {
+            document: veleroCustomPolicyDocument
+        });
+        veleroServiceAccount.role.addManagedPolicy(veleroPolicy);
+
+        const valueVariable:VeleroAddOnProps = {
```
askulkarni2, line 202: Nitpick: missing space, `valueVariable:VeleroAddOnProps` -> `valueVariable: VeleroAddOnProps`

package.json
```diff
@@ -17,6 +17,7 @@
         "@types/node": "^10.17.27",
         "@typescript-eslint/eslint-plugin": "^4.23.0",
         "@typescript-eslint/parser": "^4.23.0",
+        "@shapirov/cdk-eks-blueprint": "0.12.0",
```
askulkarni2, line 20: This should be removed from this `package.json`

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+                "Effect": "Allow",
+                "Action": [
+                    "s3:GetObject",
+                    "s3:DeleteObject",
+                    "s3:PutObject",
+                    "s3:AbortMultipartUpload",
+                    "s3:ListMultipartUploadParts",
+                    "s3:ListBucket"
+                ],
+                "Resource": "*"
```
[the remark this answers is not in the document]
haofeif: @askulkarni2  it did not work once locked down to the s3:arn/*, i am not sure why.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
+import { ClusterAddOn, ClusterInfo } from "../../spi";
+import { Constants } from "..";
+import * as s3 from "@aws-cdk/aws-s3";
+import * as iam from "@aws-cdk/aws-iam";
+import { GatewayVpcEndpointAwsService } from "@aws-cdk/aws-ec2";
+import deepmerge = require('deepmerge');
```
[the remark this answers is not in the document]
haofeif: @askulkarni2  the spread operator wont work as it will only work as the 1st layer of json documents. 
We have to use deepmerge, otherwise it wont be able to deeply merge.

lib/addons/velero/index.ts
haofeif: the idea here is to get the traffic not going through the public internet to access S3.

lib/addons/velero/index.ts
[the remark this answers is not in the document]
haofeif: resolved.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+
+}
+
+/**
+ * Defaults options for the add-on
+ */
+const defaultProps: VeleroAddOnProps = {
+    version: "2.23.6",
+    namespace: {
+        name: 'velero',
```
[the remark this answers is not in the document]
haofeif: resolved

lib/addons/velero/index.ts
haofeif: thanks. it works and resolved.

lib/addons/velero/index.ts
haofeif: if it is decided not to keep it more than happy to delete them.

package.json
haofeif: resolved.

lib/addons/velero/index.ts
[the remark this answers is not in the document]
haofeif: example:
```json
    values:{
        initContainers:[
            {
                name: "velero-plugin-for-aws",
                image: "velero/velero-plugin-for-aws:v1.2.0",
                imagePullPolicy: "IfNotPresent",
                volumeMounts:[
                    {
                        mountPath: "/target",
                        name: "plugins"
                    }
                ]
            }
        ],
        configuration: {
            provider: "aws",
            backupStorageLocation:{
                name: "default",
                config:{}
            },
            volumeSnapshotLocation:{
                name: "default",
                config:{}
            },
        },
        serviceAccount: {
            server:{}
        }
```

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,227 @@
...
+            })
+            bucketName = bucket.bucketName;
+        }
+        else {
+            bucketName = props.values.configuration.backupStorageLocation.bucket
+        }
+
+        // Create Namespace if not specified
+        if (props.namespace){
+            // Create Namespace if the "create" option is false
```
shapirov103, line 119: minor: wrong comment. I assume create namespace if create option is true.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,227 @@
...
+            cluster.vpc.addGatewayEndpoint("velero-backup-bucket-vpcEndPoint", {
+                service: GatewayVpcEndpointAwsService.S3
+            })
+            bucketName = bucket.bucketName;
+        }
+        else {
+            bucketName = props.values.configuration.backupStorageLocation.bucket
+        }
+
+        // Create Namespace if not specified
```
shapirov103, line 117: Did you mean "Create namespace if specified"?

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,227 @@
...
+        if (props.namespace){
+            // Create Namespace if the "create" option is false
+            if (props.namespace.create) {
+                console.log ("namespace:" + props.namespace.name + " does not existed, creating")
```
shapirov103, line 121: Let's drop the console.log. If there is any output that you think customer should be aware of, please use `CfnOutput`

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,227 @@
...
+
+        // Setup IAM Role for Service Accounts (IRSA) for the Velero Service Account
+        const veleroServiceAccount = cluster.addServiceAccount (
+            "velero-account",
+            {
+                name: "velero-account",
+                namespace: veleroNamespace
+            }
+        );
+        const s3bucket = s3.Bucket.fromBucketName(cluster, "S3Bucket", bucketName);
```
shapirov103, line 150: We aspire to adhere to "Clean Code" conventions. My recommendation is to approach this "script style" method structurally and break into more manageable methods, for example:
```
this.createNamespaceIfNeeded( ...)
this.createServiceAccount(...)
this.createS3Bucket(...)
... run helmChart ...
```

This will at the very least make it easier to read and maintain.

package.json
```diff
@@ -37,16 +37,23 @@
...
         "bcrypt": "^5.0.1",
+        "deepmerge": "^4.2.2",
         "js-yaml": "3.13.1",
+        "lint": "^0.7.0",
         "sync-request": "6.1.0",
         "user": "^0.0.0",
         "yaml": "^1.10.2"
-    }
+    },
+    "bundledDependencies": [
```
shapirov103, line 56: Do we need this?  E.g. do you not expect this to come from npm? (maybe a moot point if we decide to move forward with ts-deepmerge).

shapirov103 (commented):
Very nice PR, a lot of work, thank you! 
Some comments, mostly minor/style related.

lib/addons/velero/index.ts
askulkarni2: Yep, I understand. I am not opposed to using endpoints. I specifically mean if an [endpoint resource policy](https://docs.aws.amazon.com/vpc/latest/privatelink/vpc-endpoints-access.html) should be used here or not. As a security best practice it should be and know for a fact that several FIS customers actually mandate it.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,239 @@
...
+        else {
+            bucketName = props.values.configuration.backupStorageLocation.bucket
+        }
+
+        // Create Namespace if not specified
+        if (props.namespace){
+            // Create Namespace if the 'create' option is false
+            if (props.namespace.create) {
+                console.log ("namespace:" + props.namespace.name + " does not existed, creating")
+                cluster.addManifest('velero-namespace',
```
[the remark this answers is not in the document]
haofeif: resolved.

lib/addons/velero/index.ts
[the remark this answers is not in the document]
haofeif: Thanks @shapirov103  did not know the `ts-deepmerge` before. Thanks for letting me know and update the package accordingly.

lib/addons/velero/index.ts
haofeif: Cross them out as suggested.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,227 @@
...
+     * Velero for the Velero Helm Chart.
+     * @default 2.23.6
+     */
+    version?: string;
+   
+    /**
+     * Namespace for the add-on. If the namespace does not exist, specify create to false
+     * @default {namespace: "velero", create: true}
+     */
+    namespace?:  {
```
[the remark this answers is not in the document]
haofeif: sorry i think it is a bit confusing.  Changed to below format:
```json
    namespace?:  {
        name: string,   // default value is "velero", if it is specified then it will be the namespace of where velero is deployed to.
        create: boolean // default is true, if it is false, no new namespace will be created
    };
```

lib/addons/velero/index.ts
haofeif: thanks resolved accordingly.

lib/addons/velero/index.ts
haofeif: Totally agreed with you on the clean code conventions. But given that this line of code 
```typescript
const s3bucket = s3.Bucket.fromBucketName(cluster, "S3Bucket", bucketName);
```
is only to be passed to the iam resource name, and with the lines of code to create the s3 bucket, i am thinking to keep the code as it is instead of wrapping them into a function.

package.json
haofeif: yes resolved with the suggestion.

lib/addons/velero/index.ts
shapirov103: I guess I was not clear in my comment. I was not referring to the specific line of code, more referring to general structure of the deploy method. Indeed, extracting just s3 bucket does not make sense on its own. Here is what I propose:

```
    this.createNamespaceIfNeeded( ...)
    this.createServiceAccount(...)
    this.createS3Bucket(...)
... run helmChart ...
    return Promise.resolve(helmChartConstruct); 
}
```

Also please note the return value as it can be leveraged for dependency if any other add-on needs to depend on this one. Also if returned it becomes accessible in team constructs and post deployment steps (e.g. apply some golden state from S3 bucket on cluster post install).

shapirov103 (commented):
Thank you for promptly addressing all the feedback. We are mostly done, I just left one more comment, please let me know if it is clear. Feel free to ping me on slack to discuss if needed.

lib/addons/velero/index.ts
haofeif: I removed the gateway endpoint creation assuming that in the enterprise there might have an existing endpoint they refer to. It does not have to rely on the endpoint created here.

lib/addons/velero/index.ts
haofeif: @shapirov103  i got what you meant. I will refer to how argocd addon was written and wrap that to a similar format. will get back to you shortly.

haofeif: push the latest change please help to review to see whether it works as you expected. @shapirov103

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Return S3 Bucket
+     * @param clusterInfo 
+     * @param id S3-Bucket-Postfix 
+     * @param existingBucketName exiting provided S3 BucketName if it exists 
+     * @returns the existing provided S3 bucket name or the newly created S3 bucket name
+     */
+    protected returnS3BucketName(clusterInfo: ClusterInfo, id: string, existingBucketName: null|string ): string {
```
shapirov103, line 161: Please rename to `createS3BucketIfNeeded` or `getOrCreateS3Bucket`. We don't use "return" in method names, since most of them return something.

shapirov103: Please change the return type from `string` to `s3.Bucket`. You will need it in the `deploy` method to create proper CFN dependencies.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Return Velero Namespace where Velero will be installed onto
+     * @param clusterInfo
+     * @param defaultName the Default Namespace for Velero if nothing specified 
+     * @param namespace
+     * @returns the namespace created or existed.
+     */
+    protected createNamespaceIfNeeded(clusterInfo: ClusterInfo, defaultName: string, namespace?: {name: string, create: boolean}): string {
```
shapirov103, line 183: If namespace is created, you will need the actual namespace node, so that you can create proper dependencies. For example, you cannot create service account until namespace is created.  See `sa.node.addDependency(namespace);` in ArgoCD add-on.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Return Velero Namespace where Velero will be installed onto
+     * @param clusterInfo
+     * @param id
+     * @param namespace Velero namespace name
+     * @param s3BucketName the S3 BucketName where Velero will stores the backup onto
+     * @returns the service Account Name
+     */
+    protected createIRSA(clusterInfo: ClusterInfo, id: string, namespace: string, s3BucketName: string): string {
```
shapirov103, line 206: You will need to return the actual ServiceAccount as opposed to string to setup dependencies.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
...
+ 
+        this.chartNode = cluster.addHelmChart("velero-addon", {
+            chart: "velero",
+            repository: "https://vmware-tanzu.github.io/helm-charts/",
+            release: Constants.SSP_ADDON,
+            namespace: veleroNamespace,
+            version: props.version,
+            values: values
+        });
+        return this.chartNode;
```
shapirov103, line 151: You need to add dependency on the service account before returning here.

lib/addons/velero/index.ts
shapirov103: IRSA stands for IAM Roles for Service Accounts. So the method name is `createIAMRolesForServiceAccounts` which does not sound right, although I understand what you mean. Consider renaming to `createServiceAccountWithIamRoles` or just `createServiceAccount`.

shapirov103 (commented):
Definitely better, getting there. 

We have to leverage the CDN dependency mechanism to prevent sporadic failures. CFN fires all constructs in parallel, unless there is a clear dependency order specified. 

That means that your namespace, serviceaccount, helmchart will be executed in parallel and will start occasionally failing when for example service account is created before the namespace. 

So the way your add-ons stands now, it will result in race conditions and will fail  intermittently. I added comments about dependencies in the code, please take a look.

lib/addons/velero/index.ts
haofeif: I wrote it to return s3.iBucket taking into account the scenario of users providing an existing bucketName.

lib/addons/velero/index.ts
haofeif: I referred to ArgoCD add-on, but in ArgoCD a namespace has to be created. But in Velero's case, it supports an existing namespace. I have commented in the actual code and please have a look.

lib/addons/velero/index.ts
haofeif: resolved.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
...
+               
+        // Create S3 bucket if no existing bucket, create s3 bucket and corresponding KMS key
+        const s3Bucket = this.getOrCreateS3Bucket(clusterInfo, "backup-bucket", props.values.configuration.backupStorageLocation.bucket)
+
+        // Create Namespace if namespace is not explicied defined.
+        const veleroNamespace = this.createNamespaceIfNeeded(clusterInfo, "velero", props.namespace);
+
+        // Setup IAM Role for Service Accounts (IRSA) for the Velero Service Account
+        const veleroServiceAccount = this.createServiceAccountWithIamRoles(clusterInfo, "velero-account", veleroNamespace, s3Bucket);
+        //veleroServiceAccount.node.addDependency(veleroNamespace);
```
haofeif, line 113: @shapirov103 Line 113 doesnt work, as in line 109 `veleroNamespace` is a string, as in this scenario it takes into account an existing namespace, in that case no namespace is required to be created. That is why i commented it out.

haofeif: However, line 112 has already had the inner dependency, as the parameter `veleroNamespace` is referred to the line 109, thus, it has to rely on namespace being created first.

```typescript
const veleroServiceAccount = this.createServiceAccountWithIamRoles(clusterInfo, "velero-account", veleroNamespace, s3Bucket);
```

lib/addons/velero/index.ts
shapirov103: I understand your comment. However, in case namespace is created lack of dependency between service account and namespace WILL cause sporadic failures. 
It means that the code must create a CFN dependency between namespace and service account *if namespace is created by us*.

You can make this function do this:

```typescript

protected createNamespaceIfNeeded( ... ) : { name: string, manifest?: KubnernetesManifest { 
...
    return {name: somename, manifest: (either manifest or undefined)
}
```

Then in deploy (pseudocode):

```typescript

const namespace = createNamespaceIfNeeded(...);
sa = createServiceAccount(...);

if(namespace.manifest) {
   sa.node.addDependency(namespace.manifest);
}

```

lib/addons/velero/index.ts
shapirov103: The thing about CDK is that it will produce a CloudFormation Template, and for the template the order that you defined in your code does not matter. You need to use CFN dependency mechanism to handle this.

shapirov103 (commented):
Looking good. Please see a couple of comments that I left on the remaining questions. Once addressed, I will kick off the build and merge.

lib/addons/velero/index.ts
```diff
@@ -0,0 +1,259 @@
+import { ClusterAddOn, ClusterInfo } from "../../spi";
+import { Constants } from "..";
+import { Construct } from "@aws-cdk/core";
+import * as s3 from "@aws-cdk/aws-s3";
+import * as iam from "@aws-cdk/aws-iam";
+import { HelmChart, KubernetesManifest, ServiceAccount } from "@aws-cdk/aws-eks";
```
askulkarni2, line 6: `KubernetesManifest` can be removed from import.

askulkarni2 (commented):
Pending @shapirov103's comments, lgtm!

lib/addons/velero/index.ts
haofeif: resolved.

lib/addons/velero/index.ts
haofeif: Thanks for the detailed code recommendations, it worked perfectly. Much appreciated for your patience, I learnt a lot from this process.

lib/addons/velero/index.ts
haofeif: after addressing Mikhail's comments, it because useful again :)

shapirov103 (approved):
LGTM

Issue #197, opened 2021-09-10
[TASK] - Refactor addon code for decorator pattern to handle dependencies


askulkarni2:
Ideally, I would like to refactor to some common addon implementation where you can specify in constructor the ids of dependencies and it will manage it automatically, as opposed to spreading similar code across multiple add-ons. E.g base class `BasicClusterAddOn` has an array of dependency names and . I am sure this will be used not just by us. 

A better approach to explore is to use a decorator pattern like https://www.typescriptlang.org/docs/handbook/decorators.html. I have not used it myself with typescript yet (used a lot with java) but it can look something like:

```
 class MyAddOn implements ClusterAddOn {
     
     @dependable('AwsLoadBalancerController')
     deploy(clusterInfo:  ClusterInfo): Promise<> {
      .....
     }
     
 }
```

We can defer this as well as functionally it serves the puprose (e.g. create a task to address later on).

_Originally posted by @shapirov103 in https://github.com/aws-quickstart/ssp-amazon-eks/pull/195#discussion_r705628590_

PR #207, opened 2021-09-22
refactor for decorator pattern for handling inter-addon dependencies

askulkarni2:
*Issue #, if available:* #197

*Description of changes:*
Adding a new `@dependable([list of AddOns])` decorator to handle addon dependencies to remove the repeatable code for asserting an dependent addon is scheduled as well as create resource dependencies in the cloudformation resources.

### DIFF
--- a/lib/addons/nginx/index.ts
+++ b/lib/addons/nginx/index.ts
@@ -1,6 +1,7 @@
 import { Construct } from "@aws-cdk/core";
 import { Constants } from "..";
 import { ClusterAddOn, ClusterInfo } from "../../spi";
+import { dependable } from "../../utils";
 
 /**
  * Properties available to configure the nginx ingress controller.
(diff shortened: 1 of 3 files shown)


lib/utils/addon-utils.ts
```diff
@@ -8,4 +9,44 @@ import { ClusterAddOn } from '../spi';
...
+}
+
+/**
+ * Decorator function that accepts a list of AddOns and
+ * ensures addons are scheduled to be added as well as
+ * add them as dependencies
+ * @param addOns 
+ * @returns 
+ */
+export function dependable(addOns: Array<string>) {
```
shapirov103, line 21: I suggest we use this form of declaration:

`dependable(...addOns: string )`

This will allow to declare dependency as  `@dependendable('AwsLoadbalancerController', 'SomeOtherAddon')`

shapirov103 (commented):
Awesome stuff, thank you! 

Small comment on usage. Do you think we could add a test for this, e.g. new Blueprint with NginxAddOn (no AWS Load balancer), expect Fail, something like that?

lib/addons/nginx/index.ts
```diff
@@ -76,21 +77,11 @@ export class NginxAddOn implements ClusterAddOn {
         this.options = { ...defaultProps, ...props };
     }
 
-    deploy(clusterInfo: ClusterInfo): void {
+    @dependable(['AwsLoadBalancerControllerAddOn'])
```
kcoleman731, line 80: we should probably create an addon enum so we don't have to rely on string here.

lib/utils/addon-utils.ts
kcoleman731: @askulkarni2 would love to see some tests for this.

kcoleman731 (approved):

lib/utils/addon-utils.ts
askulkarni2: Yep, adding the tests.

lib/addons/nginx/index.ts
askulkarni2: Did you mean just locally or a global `enum` for all addons?

shapirov103: We cannot envision all the add-ons that will be there, including those contributed by the partners or custom add-ons that customers will implement. We could provide a convenience enum (similar to GlobalResources) that makes it easy for the add-ons that we know about. In addition, we now have add-ons that can be applied more than once (such as nested stack) and dependency between those will have to be based on ids provided at design time.

lib/utils/addon-utils.ts
askulkarni2: addressed

shapirov103 (approved):
lgtm

PR #230, opened 2021-11-03
Feature/argocd config

shapirov103:
*Issue #, if available:*

*Description of changes:* Work in progress but works end to end with app mesh add on. 


shapirov103:
GitOps addon management is not a requirement for the CDK version as communicated by Kevin.

PR #235, opened 2021-11-10
bug fix : property 'addons' does not exist when running the quickstart

pmadc:
*Issue #, if available:*

*Description of changes:*
Removing the addons property as it does not exist on the ssp object when running the quickstart example.


pmadc:
@shapirov103 thanks for your comment. I got confused by my editor showing error "property 'addons' does not exist" but indeed when you run cdk deploy there is no compile error. I tried with another editor and there is no error message. I am closing this pull-request.

PR #282, opened 2022-02-02
Updating opa gatekeeper documentation

parkand1:
Refresh of OPA Gatekeeper documentation:

1. Added usage section for adding gatekeeper to your blueprint 
2. Added section for key terminology
3. Used simpler example (all namespaces must be created with labels)
4. Added additional links for exploration


askulkarni2 (dismissed):

docs/addons/opa-gatekeeper.md
```diff
@@ -15,181 +15,87 @@ In the context of a Shared Services Platform running on Amazon EKS, platform tea
...
+```typescript
+import * as ssp from '@aws-quickstart/ssp-amazon-eks';
 
-input_containers[container] {
-  container := input.request.object.spec.containers[_]
-}
+const app = new cdk.App();
+const account = <AWS_ACCOUNT_ID>;
+const region = <AWS_REGION>;
+const env: { account, region },
```
shapirov103, line 36: you don't need env if it is only for account/region. You set those on lines 39 and 40.

shapirov103 (commented):
Looks good, one tiny comment.

docs/addons/opa-gatekeeper.md
parkand1: addressed

shapirov103 (approved):
lgtm. let's plan on a pattern that applies those policies with OPA (either through pipeline or through gitops).

PR #289, opened 2022-02-08
Feature/ebsCsiAddon

### LINKED ISSUE #232: [ADDON] - EBS CSI Driver
Add support for installing the EBS CSI Driver in a cluster via Helm.

https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html

Example managed add-on implementation https://github.com/aws-quickstart/ssp-amazon-eks/blob/main/lib/addons/coredns/index.ts

camba1:
New EBS CSI Driver add-on

Fixes #232 

### DIFF
--- a/docs/addons/index.md
+++ b/docs/addons/index.md
@@ -17,7 +17,7 @@ The framework currently supports the following add-ons.
 | [`ClusterAutoscalerAddOn`](./cluster-autoscaler)                       | Adds the standard cluster autoscaler                                                                                                  |
 | [`ContainerInsightsAddOn`](./container-insights)                       | Adds Container Insights support integrating monitoring with CloudWatch                                                                |
 | [`CoreDnsAddOn`](./coredns.md)                                         | Adds CoreDNS Amazon EKS add-on. CoreDNS is a flexible, extensible DNS server that can serve as the Kubernetes cluster DNS             |
-| [`EbsCSIDriverAddOn`](./ebs-csi-driver.md)                             | Adds EBS CSI Driver Amazon EKS add-on. This driver manages the lifecycle of Amazon EBS volumes for persistent storage                 |
+| [`EbsCsiDriverAddOn`](./ebs-csi-driver.md)                             | Adds EBS CSI Driver Amazon EKS add-on. This driver manages the lifecycle of Amazon EBS volumes for persistent storage                 |
 | [`KubecostAddOn`](./kubecost.md)                                       | Adds [Kubecost](https://kubecost.com) cost analyzer to the EKS cluster                                                                |
 | [`KarpenterAddOn`](./karpenter.md)                                     | Adds [Karpenter](https://github.com/awslabs/karpenter) support for Amazon EKS.                                                        | 
 | [`KubeProxyAddOn`](./kube-proxy.md)                                    | Adds kube-proxy Amazon EKS add-on. Kube-proxy maintains network rules on each Amazon EC2 node                                         |
--- a/docs/addons/index.md
+++ b/docs/addons/index.md
@@ -8,28 +8,29 @@ Within the context of the `ssp-amazon-eks` framework, an add-on is abstracted as
 
 The framework currently supports the following add-ons.
 
-| Add-on             | Description                                                                       |
-|-------------------|-----------------------------------------------------------------------------------|
-| [`AppMeshAddOn`](./app-mesh) | Adds an AppMesh controller and CRDs (pending validation on the latest version of CDK) |
-| [`ArgoCDAddOn`](./argo-cd) | Provisions Argo CD into your cluster. |
-| [`AWS Load Balancer Controller`](./aws-load-balancer-controller) | Provisions the AWS Load Balancer Controller into your cluster. |
-| [`CalicoAddOn`](./calico) | Adds the Calico 1.7.1 CNI/Network policy engine |
-| [`ClusterAutoscalerAddOn`](./cluster-autoscaler) | Adds the standard cluster autoscaler |
-| [`ContainerInsightsAddOn`](./container-insights) | Adds Container Insights support integrating monitoring with CloudWatch |
-| [`CoreDnsAddOn`](./coredns.md) | Adds CoreDNS Amazon EKS add-on. CoreDNS is a flexible, extensible DNS server that can serve as the Kubernetes cluster DNS |
-| [`KubecostAddOn`](./kubecost.md) | Adds [Kubecost](https://kubecost.com) cost analyzer to the EKS cluster |
-| [`KarpenterAddOn`](./karpenter.md) | Adds [Karpenter](https://github.com/awslabs/karpenter) support for Amazon EKS. | 
-| [`KubeProxyAddOn`](./kube-proxy.md) | Adds kube-proxy Amazon EKS add-on. Kube-proxy maintains network rules on each Amazon EC2 node |
-| [`MetricsServerAddOn`](./metrics-server.md) | Adds metrics server (pre-req for HPA and other monitoring tools)|
-| [`ExternalDnsAddOn`](./external-dns) | Adds [External DNS](https://github.com/kubernetes-sigs/external-dns) support for AWS to the cluster, integrating with Amazon Route 53
-| [`MetricsServerAddOn`](./metrics-server) | Adds metrics server (pre-req for HPA and other monitoring tools)|
-| [`NginxAddOn`](./nginx.md) | Adds NGINX ingress controller |
-| [`SecretsStoreAddOn`](./secrets-store.md) |  Adds AWS Secrets Manager and Config Provider for Secret Store CSI Driver to the EKS Cluster |
-| [`SSMAgentAddOn`](./ssm-agent.md) | Adds [Amazon SSM Agent](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html) to worker nodes |
-| [`VpcCniAddOn`](./vpc-cni.md) | Adds the Amazon VPC CNI Amazon EKS addon to support native VPC networking for Amazon EKS |
-| [`Weave GitOps`](https://github.com/weaveworks/weave-gitops-ssp-addon) | [Weave GitOps Core](https://www.weave.works/product/gitops-core/) AddOn |
-| [`XrayAddOn`](./xray) | Adds XRay Daemon to the EKS Cluster |
-| [`VeleroAddOn`](./velero.md) | Adds [Velero](https://velero.io/) to the EKS Cluster |
+| Add-on                                                                 | Description                                                                                                                           |
+|------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|
+| [`AppMeshAddOn`](./app-mesh)                                           | Adds an AppMesh controller and CRDs (pending validation on the latest version of CDK)                                                 |
+| [`ArgoCDAddOn`](./argo-cd)                                             | Provisions Argo CD into your cluster.                                                                                                 |
+| [`AWS Load Balancer Controller`](./aws-load-balancer-controller)       | Provisions the AWS Load Balancer Controller into your cluster.                                                                        |
+| [`CalicoAddOn`](./calico)                                              | Adds the Calico 1.7.1 CNI/Network policy engine                                                                                       |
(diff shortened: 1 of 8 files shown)


lib/addons/ebs-csi-driver/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        // Permissions
+
+        const serviceAccount = cluster.addServiceAccount(defaultProps.addOnName, {
+            name: defaultProps.addOnName,
+            namespace: defaultProps.namespace,
+        });
+
+        AmazonEksEbsCsiDriverPolicy.Statement.forEach((statement) => {
+            serviceAccount.addToPrincipalPolicy(iam.PolicyStatement.fromJson(statement));
+        });
```
askulkarni2, line 49: Lets move the IRSA creation to `core-addon.ts` and add another property to the interface to accept `PolicyStatement[]`. This way its available for the next EKS managed addon that needs an IRSA role and also lets us use `extends` instead of implementing the `ClusterAddOn` interface here.

```ts
constructor(version?: string) {
    super({
        addOnName: "aws-ebs-csi-driver",
        version: version ?? "v1.4.0-eksbuild.preview"
     });
}
```
@shapirov103 thoughts?

Also should we take this opportunity to provide for a `ConfigMap` property on the interface that lets users alter the fields in the add-on configuration?

askulkarni2:
Add example to `examples/blueprint-construct/index.ts` so it gets tested with our e2e tests.

lib/addons/ebs-csi-driver/index.ts
shapirov103: It makes sense to centralize IRSA for core addons, although the approach for vpc-cni was a bit different and we did not manage to use SA when add-on was added after the cluster has been already provisioned. I assume EBS add-on may not fall into the same bucket. 
If anything, you can leverage our `sa-utils.ts`, we have `createServiceAccount` there as well.

docs/addons/index.md
```diff
@@ -8,28 +8,29 @@ Within the context of the `ssp-amazon-eks` framework, an add-on is abstracted as
...
-| [`MetricsServerAddOn`](./metrics-server) | Adds metrics server (pre-req for HPA and other monitoring tools)|
-| [`NginxAddOn`](./nginx.md) | Adds NGINX ingress controller |
-| [`SecretsStoreAddOn`](./secrets-store.md) |  Adds AWS Secrets Manager and Config Provider for Secret Store CSI Driver to the EKS Cluster |
-| [`SSMAgentAddOn`](./ssm-agent.md) | Adds [Amazon SSM Agent](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html) to worker nodes |
-| [`VpcCniAddOn`](./vpc-cni.md) | Adds the Amazon VPC CNI Amazon EKS addon to support native VPC networking for Amazon EKS |
-| [`Weave GitOps`](https://github.com/weaveworks/weave-gitops-ssp-addon) | [Weave GitOps Core](https://www.weave.works/product/gitops-core/) AddOn |
-| [`XrayAddOn`](./xray) | Adds XRay Daemon to the EKS Cluster |
-| [`VeleroAddOn`](./velero.md) | Adds [Velero](https://velero.io/) to the EKS Cluster |
+| Add-on                                                                 | Description                                                                                                                           |
```
shapirov103, line 11: Is there a formatting issue here? Why all of the entries highlighted as change? It makes it harder to figure out what changed specifically.

shapirov103 (commented):
Looks great, let's leverage the utils for SA creation. We can refactor of there is any other add-on relying on IRSA.

docs/addons/index.md
camba1: It is a formatting in the original document. According to the IDE, there are potential rendering issues with markdown tables when the columns do not align properly. As such the table was reformatted. The only new entry for this add-on is :

| [`EbsCsiDriverAddOn`](./ebs-csi-driver.md)                             | Adds EBS CSI Driver Amazon EKS add-on. This driver manages the lifecycle of Amazon EBS volumes for persistent storage                 |

shapirov103: It was clear to me that IDE reformatted, but we did not have rendering issues. And, yes, I surely saw the add-on. I think we should turn off automatic reformat, and only apply it when the intention is to reformat. 
it is minor.

camba1:
@askulkarni2 and @shapirov103: As discussed,  I refactored the coreAddOn to take an optional policy document and namespace. Those are then used to create the service account using the sa-utils createServiceAccount function. 
I also refactored the ebs csi driver to extend the new version of the coreAddOn. 
Let me know if there is anything else we can improve upon and want to do at this time.
Thanks!

lib/addons/ebs-csi-driver/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor(version?: string) {
+        super({
+            addOnName: defaultProps.addOnName,
+            version: version ?? defaultProps.version,
+            policyDocument: PolicyDocument.fromJson(AmazonEksEbsCsiDriverPolicy),
+            namespace: defaultProps.namespace
+        });
+    }
+}
```
askulkarni2, line 27: Love it!

lib/addons/core-addon/index.ts
```diff
@@ -20,11 +40,25 @@ export class CoreAddOn implements ClusterAddOn {
...
     deploy(clusterInfo: ClusterInfo): void {
+
+        // Create a service account if user provides namespace and service account
+        let serviceAccountRoleArn: string | undefined = undefined;
+        let namespace: string = this.coreAddOnProps?.namespace ? this.coreAddOnProps.namespace : DEFAULT_NAMESPACE;
+
+        if (this.coreAddOnProps?.policyDocument) {
+            const serviceAccount = createServiceAccount(clusterInfo.cluster, this.coreAddOnProps.addOnName,
+                namespace, this.coreAddOnProps.policyDocument);
+            serviceAccountRoleArn = serviceAccount?.role?.roleArn;
```
askulkarni2, line 51: I think this should be `serviceAccountRoleArn = serviceAccount.role.roleArn;`  as `createServiceAccount()` is guaranteed to return an SA.

camba1: Updated. Thanks!

lib/addons/core-addon/index.ts
```diff
@@ -20,11 +40,25 @@ export class CoreAddOn implements ClusterAddOn {
     }
 
     deploy(clusterInfo: ClusterInfo): void {
+
+        // Create a service account if user provides namespace and service account
+        let serviceAccountRoleArn: string | undefined = undefined;
+        let namespace: string = this.coreAddOnProps?.namespace ? this.coreAddOnProps.namespace : DEFAULT_NAMESPACE;
```
shapirov103, line 46: Please correct me if I am wrong, but if customers supply a custom namespace name (e.g. ebs-addon) then this code will fail since namespace won't exist. 
If my line of reasoning is correct, then let's use namespace-utils.createNamespace. This one won't fail if it already exists.

lib/addons/ebs-csi-driver/index.ts
shapirov103: +1

shapirov103 (commented):
Looks great! I left one comment, please take a look. 
Once addressed and e2e passes, will merge.

lib/addons/core-addon/index.ts
camba1: @shapirov103 : My thinking was that while currently all add-ons go into Kube-system, allowing them to pass it in would future proof the core-addon since it is a building block. If that is not necessary, let me know and I can change it.

shapirov103: I think it is a good option to add. Some customers have weird organization of namespaces. Let's just call createNamespace (e.g. kubectl apply through our utils) and we should be good.

shapirov103:
Hmm, the test failed due to karpenter failure. @camba1 did you run e2e with Karpenter in your account?

shapirov103:
@camba1 e2e tests are failed twice with the same error, so appears to be a consistent failure on Karpenter. Have you observed the same in your account?

shapirov103:
> @shapirov103: I ran cdk list, npm test and they ran fine. However when I try deploying Karpenter and EBS to EKS in my account, I just got a `time out waiting for the condition.` error. Is that the same error are you seeing? Where can I see it? How do I enable e2e tests 

@camba1 you can see what the CI test is doing in `ci/buildspec.yml`. This will deploy all clusters. For you purposes you can just run `cdk deploy eks-blueprint-vpc` and `cdk deploy blueprint-construct-dev`.

Yes, we get the same error in the CI account. Only a couple of contributors have access to the account atm, but normally if it runs in your account it should work in the CI account as well.

For the namespace create, please try `createNamespace(name, cluster, true)`.

camba1:
@shapirov103: Thanks for the info about ci/buildspec.yml. I will start running that on a regular basis.
In terms of the Karpenter error, I played with it on a test cluster. I noticed the error is thrown if Karpenter is added to a cluster containing ebs driver. But, if the ebs driver is added to a cluster that already has Karpenter running, then some of the ebs drivers pods are not scheduled as the scheduler complains that there is not enough cpu. So, I added a second node to the cluster using:

```
const clusterProviderProps: MngClusterProviderProps = {
    minSize: 2,
    maxSize: 2,
    desiredSize: 2,
    version: KubernetesVersion.V1_20
}
```
And at that point I was able to deploy everything properly without any other changes. Should we update the size of the cluster we are using for testing?

On the ebs driver throwing an error when explicitly creating the namespace. I still get the same error even with `createNamespace(name, cluster, true)`

shapirov103:
@camba1 you can add the desired node size as part of the PR to the blueprint-construct-dev cluster. i assume the cluster CA should scale it up but it fails on a timeout. 

With respect to the namespace: it is seems odd, since the overwrite flag means it will be applied with `kubectl apply` as per [this](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-eks.KubernetesManifestProps.html#overwrite).
Can you check if that change made into the synthesized template just to make sure? If it fails consistently let's discuss on the plan.

camba1:
I have upgraded the test cluster. It now will instantiate 2 m5.xlarge which successfully deploys everything. I tried with 2 m5;large, but that still saw the same error. In the end, there was just not enough cpu in the m5 for all the add-ons we are deploying for testing

For the namespace issue, the error is throw during cdk list/synth, so it is not even making it to Kubernetes.  If I remove the line and do cdk synth, there is one reference in the stack to that object:
`Metadata: aws:cdk:path: blueprint-construct-dev/kube-system-namespace-struct/Resource/Default.`  within the `kubesystemnamespacestructA6B3FF7C` resource

lib/addons/core-addon/index.ts
askulkarni2: @shapirov103 looks like `CfnAddon()` accepts `serviceAccountRoleArn` but no `namespace` as parameter. Will  EKS addon honor the `namespace` passed via a `serviceAccount`? Or are we just future proofing here?

askulkarni2 (approved):

lib/addons/core-addon/index.ts
shapirov103: We don't need to future-proof it in my opinion. It may confuse customers. Plus it is taking serviceAccountRoleArn rather than sa itself so I doubt it will honor the sa namespace (@camba1 have you tested with it?).

shapirov103:
e2e failing with a timeout from i observe. No errors. I will give it another try. I think more than 2 hours for an e2e test is a bit pushing it.

shapirov103:
@camba1  the e2e test failure is due to a timeout. Let's do the following:

1. Comment out the custom AMI blueprint creation for now. I will address it with my PR for multi-node group support. It is under `bin/main.ts` lines 37 and all the way down.
2. For the blueprint construct let's keep the node group to desired size of 1 and use a bigger box, e.g. m5.2xlarge.

camba1:
@shapirov103  I made the changes we discussed: 
- Removed the namespace option from core-addon and ebs csi add-on
- Commented out second for custom AMI creation blueprint
- Updated blueprint construct to use one node (default) and use a m5.2xLarge box
Blueprint cluster creation was successful and took 23 minutes in my environment 

Thanks!

shapirov103 (approved):
lgtm!

PR #290, opened 2022-02-08
Feature/KubeviousAddon

### LINKED ISSUE #292: [ADDON] - Kubevious

camba1:
New [Kubevious](https://github.com/kubevious/kubevious) add-on (ported from the workshop)

Fixes #292 


lib/addons/kubevious/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    namespace: "kubevious",
+    chart: "kubevious",
+    version: "0.8.15",
+    release: "kubevious",
+    repository:  "https://helm.kubevious.io",
+    values: {},
+
+    ingressEnabled: false,
+    kubeviousServiceType: "ClusterIP",
+    mysqlRootPassword: "kubevious"
```
askulkarni2, line 31: Seems like these are defaults. Let customers pass `values` if they wish to override.

lib/addons/kubevious/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+function populateValues(helmOptions: KubeviousAddOnProps): Values {
+    const values = helmOptions.values ?? {};
+
+    setPath(values, "ingress.enabled",  helmOptions.ingressEnabled);
+    setPath(values, "kubevious.service.type",  helmOptions.kubeviousServiceType);
+    setPath(values, "mysql.root.password",  helmOptions.mysqlRootPassword);
+
+    return values;
+}
```
askulkarni2, line 66: Add example to `examples/blueprint-construct/index.ts` so it gets tested with our e2e tests.

askulkarni2 (commented):
Add exampl

lib/addons/kubevious/index.ts
camba1: I added the line below to examples/blueprint-construct/index.ts:
 new ssp.addons.KubeviousAddOn(),

My understanding was that adding that would be enough to trigger the add-on to be tested. If my understanding is not correct, please let me know.

lib/addons/kubevious/index.ts
camba1: Hi  @askulkarni2 , the defaultProps contains the defaults while the user can pass override value props on the class constructor. The default values and the user provided values are loaded into this.options. The class uses this.options to instantiate the add-on . As such the user can override the defaults already. Please let me know if I missunderstood the request

docs/addons/kubevious.md
```diff
@@ -0,0 +1,55 @@
...
+    .addOns(new ssp.KubeviousAddOn() )
+    .build(app, 'my-cluster');
+```
+
+## Configuration Options
+
+- `version`: Version fo the Helm Chart to be used to install Kubevious
+- `ingressEnabled`: Indicates whether to expose Kubevious using an ingress gateway. Set to false by default
+- `kubeviousServiceType`: Type of service used to expose Kubevious backend. Set to 'ClusterIP' by default
+- `mysqlRootPassword`: Default password for the root account in the MySql database. Set to 'kubevious' by default
```
shapirov103, line 27: this seems to be a security issue. Is there a better way to pass this password? Can we just generate this password and store in a secret instead? Otherwise it will show up as clear text in the CFN stack.

lib/addons/kubevious/index.ts
shapirov103: Yes, that is the intention. Default values mimic the values in the external props allowing customers to overwrite. I think we are good.

lib/addons/kubevious/index.ts
shapirov103: I see it, looks good.

shapirov103 (commented):
Looks great! One small issue on sensitive data passed as plain text.

docs/addons/kubevious.md
shapirov103: let's replace this with `mySqlRootPasswordSecret` and look it up using `secret-manager-utils.getSecretValue`. Ideally we could generate the secret but there is an inherent problem with CDK Kubectl provider not being able to use secrets (there is an issue on github but they won't fix). 
I would not mind it slide, but after that every partner will point to this. 
An alternative is to remove this field from the properties and let it default unless they pass it in values. (security by obscurity, not ideal, but is not a glaring hole).

askulkarni2 (approved):
LGTM!

shapirov103:
@camba1 I assume you tested the secret provider functionality? 
I will launch e2e on it but I see that csi secret store is not going to be leveraged.

camba1:
@shapirov103: Completed the changes we discussed this morning and the add-on should be ready to go

docs/addons/kubevious.md
```diff
@@ -0,0 +1,65 @@
...
```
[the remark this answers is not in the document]
camba1: Thanks for catching the @shapirov103 . The doc has been updated.

shapirov103 (approved):
LGTM

PR #291, opened 2022-02-08
Feature/Allow creation of IngressClass for ALB

camba1:
Modification of the existing AWS ALB add-on:
- Upgrade ALB version from 1.2.3 to 1.3.3 as the old version does not allow creating an IngressClass using the the Helm values. (See [issue 612](https://github.com/aws/eks-charts/pull/612) in the ALB repo)
- Added props to explicitly indicate to the user the creation of the ingressClass

### DIFF
--- a/lib/addons/aws-loadbalancer-controller/index.ts
+++ b/lib/addons/aws-loadbalancer-controller/index.ts
@@ -22,7 +22,18 @@ export interface AwsLoadBalancerControllerProps extends HelmAddOnUserProps {
     /**
      * Enable WAFV2 (must be false for CN partition)
      */
-    enableWafv2?: boolean
+    enableWafv2?: boolean,
+
+    /**
+     * Create the ingressClass to be used by the ALB controller
+     */
+    createIngressClassResource?: boolean
+
+    /**
+     * Name of ingressClass to the ALB controller will satisfy. If not provided
+     * the value will be defaulted to "alb"
+     */
+    ingressClass?: string
 }
 
 
@@ -37,10 +48,12 @@ const defaultProps: AwsLoadBalancerControllerProps = {
     chart: AWS_LOAD_BALANCER_CONTROLLER,
     repository: 'https://aws.github.io/eks-charts',
     release: AWS_LOAD_BALANCER_CONTROLLER,
-    version: '1.2.3',
+    version: '1.3.3',
     enableShield: false,
     enableWaf: false,
-    enableWafv2: false
+    enableWafv2: false,
+    createIngressClassResource: true,
+    ingressClass: "alb"
 };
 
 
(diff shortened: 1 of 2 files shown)


askulkarni2 (approved):
LGTM!

lib/addons/aws-loadbalancer-controller/index.ts
```diff
@@ -22,7 +22,18 @@ export interface AwsLoadBalancerControllerProps extends HelmAddOnUserProps {
...
+    /**
+     * Create the ingressClass to be used by the ALB controller
+     */
+    createIngressClassResource?: boolean
+
+    /**
+     * Name of ingressClass to the ALB controller will satisfy. If not provided
+     * the value will be defaulted to "alb"
+     */
+    ingressClass?: string
```
shapirov103, line 36: I suppose it is ignored if `createIngressClassResource` is false? We can call it out.

lib/addons/aws-loadbalancer-controller/index.ts
```diff
@@ -37,10 +48,12 @@ const defaultProps: AwsLoadBalancerControllerProps = {
     chart: AWS_LOAD_BALANCER_CONTROLLER,
     repository: 'https://aws.github.io/eks-charts',
     release: AWS_LOAD_BALANCER_CONTROLLER,
-    version: '1.2.3',
+    version: '1.3.3',
```
shapirov103, line 51: :+1:

lib/addons/aws-loadbalancer-controller/index.ts
```diff
@@ -74,6 +87,8 @@ export class AwsLoadBalancerControllerAddOn extends HelmAddOn {
             enableShield: this.options.enableShield,
             enableWaf: this.options.enableWaf,
             enableWafv2: this.options.enableWafv2,
+            createIngressClassResource: this.options.createIngressClassResource,
+            ingressClass: this.options.ingressClass
```
shapirov103, line 91: Just curious what is the outcome if ingress class is specified but createIngressClassResource is false?

shapirov103 (commented):
Looks great, minor comments. I will look into the e2e test failure, once they pass will merge.

lib/addons/aws-loadbalancer-controller/index.ts
camba1: If the createIngressClassResource is false the ingressClass does not get created and it is assumed that the user will manually create the ingressClass with the same name as the one defined in this.options.ingressClass (default = alb). This is the same behavior as the underlying Helm chart.

lib/addons/aws-loadbalancer-controller/index.ts
camba1: This value is not ignored even if the createIngressClassResource is set to false. In that case, the helm chart will create the ALB and will expect the user to create an IngressClass with the correct name matching the ingressClass name above (default = alb). This is the same behavior as the underlying Helm chart.

shapirov103:
> /do-e2e-tests

AWS::S3::Bucket                       | blueprint-construct-dev/velero-${id} (blueprintconstructdevveleroid98A45405) You have attempted to create more buckets than allowed (Service: Amazon S3; Status Code: 400; Error Code: TooManyBuckets; Request ID: B5AMBJRJC5JK0ZE3; S3 Extended Request ID: dKW+Tj8R1tr0xDNNXMUaO5NwD70DSLRbXF4TmUeFw43QEynjdvxop70iV+gWMot6IWVGLnd75sM=; Proxy: null)

We will need to clean up that account. I am in an offsite, but I will get to it at some point.

PR #297, opened 2022-02-17
Add namespace labels option to assist with the AWS App Mesh requirement

thegippygeek:
*Issue #, if available:* 293

*Description of changes:* Added namespace labels to the teams namespace  

### DIFF
--- a/lib/teams/index.ts
+++ b/lib/teams/index.ts
@@ -29,6 +29,11 @@ export class TeamProps {
     readonly namespaceAnnotations? : { [key: string]: any; } = { "argocd.argoproj.io/sync-wave": "-1" };
 
     /**
+     * Labels such as necessary for AWS AppMesh 
+     */
+    readonly namespaceLabels? : { [key: string]: any; } = {"appmesh.k8s.aws/sidecarInjectorWebhook": "enabled"}
+
+    /**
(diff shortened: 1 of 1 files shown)


lib/teams/index.ts
```diff
@@ -28,6 +28,11 @@ export class TeamProps {
      */
     readonly namespaceAnnotations? : { [key: string]: any; } = { "argocd.argoproj.io/sync-wave": "-1" };
 
+    /**
+     * Labels such as necessary for AWS AppMesh 
+     */
+    readonly namespaceLabels? : { [key: string]: any; } = {"appmesh.k8s.aws/sidecarInjectorWebhook": "enabled"};
```
shapirov103, line 34: Let's not assign this default to the namespace. We should have labels, but the content should be provided by the customers when they define their teams.

shapirov103 (commented):
Thank you @thegippygeek for your contribution. It is definitely a needed feature. I left one comment, we can merge after.

lib/teams/index.ts
thegippygeek: Yes 100% agree, removed as suggested.

shapirov103 (approved):
LGTM

PR #301, opened 2022-02-28
Use correct helm repo for secret-store-csi-driver

aceat64:
*Issue #, if available:*

*Description of changes:*

The repo URL used is not a valid chart repository, causing the following error:

> Error: looks like "https://raw.githubusercontent.com/kubernetes-sigs/secrets-store-csi-driver/main/charts" is not a valid chart repository or cannot be reached: failed to fetch https://raw.githubusercontent.com/kubernetes-sigs/secrets-store-csi-driver/main/charts/index.yaml : 404 Not Found

I've updated the repo URL to the one given by the upstream projects documentation on installing via Helm:

https://secrets-store-csi-driver.sigs.k8s.io/getting-started/installation.html#deployment-using-helm

### DIFF
--- a/lib/addons/secrets-store/index.ts
+++ b/lib/addons/secrets-store/index.ts
@@ -47,7 +47,7 @@ const defaultProps: SecretsStoreAddOnProps = {
     namespace: 'kube-system',
     version: 'v0.0.23',
     release: 'ssp-addon-secret-store-csi-driver',
-    repository: 'https://raw.githubusercontent.com/kubernetes-sigs/secrets-store-csi-driver/master/charts',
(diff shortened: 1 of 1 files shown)


shapirov103 (commented):
@aceat64 Thank you for your contribution, great catch. 

I was scratching my head trying to comprehend how this slipped through given that we run e2e tests with this code base on each PR. 

The root cause is renaming of master to main branch for the sigs, which happened a while ago, however the raw.githubusercontent apparently lingered for a while. 

I am going to run the e2e pipeline to validate the change, and merge right after.

shapirov103 (approved):
LGTM

PR #305, opened 2022-03-07
fix: Deploying to GovCloud throws malformed document error 

### LINKED ISSUE #304: Deploying to GovCloud throws malformed document error
When deploying the SSP on GovCloud, the load balancer add-on throws the following error:

`Partition "aws" is not valid for resource "arn:aws:ec2:*:*:security-group/*". (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: b8f7457b-00fe-470b-8ef7-077901db9b06; Proxy: null)`

This is because arn prefixes in commercial are different that GovCloud (`arn:aws` vs  `arn:aws-us-gov`). In particular, the Load balancer has a policy document where it is referring to resources like arn:aws:ec2 which will not work in GovCloud (`arn:aws-us-gov:ec2` is the correct version). 

We should update the code to check where the blueprint will be deployed and adjust the arns accordingly

camba1:
fixes #304

Some add-on policies had the partition section set to `aws`. However, the partition differs depending on which environment the cluster is deployed. For example, for GovCloud, the partition should be `aws-us-gov` . 
The change allows the partition to be dynamically populated based on the cluster.stack.partition for the following components:

- load balancer controller
- ebs csi add on
- core add on
- vpc utils

Note that for the load balancer, a repository mapping was created to ensure that the image is pulled from the appropriate repository as per [the load balancer documentation](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) . Trying to pull from the wrong repository causes the load balancer pods not to come online due to imagePullBackOff error.

The test blueprint with all the add-ons was deployed successfully in both commercial and gov cloud test environments. 


lib/addons/ebs-csi-driver/iam-policy.ts
```diff
@@ -1,145 +1,147 @@
...
-        "Resource": [
-            "arn:aws:ec2:*:*:volume/*",
-            "arn:aws:ec2:*:*:snapshot/*"
-        ],
-        "Condition": {
-            "StringEquals": {
-                "ec2:CreateAction": [
-                    "CreateVolume",
-                    "CreateSnapshot"
+export const AmazonEksEbsCsiDriverPolicy = (partition: string): object => {
```
shapirov103, line 1: Since it is a function now let's name it accordingly:

```typescript
export function getEbsDriverPolicyDocument(partition: string) : PolicyDocument 
```

Otherwise it is confusing to see some that is named like a constant used with arguments.

lib/addons/core-addon/index.ts
```diff
@@ -17,7 +17,7 @@ export class CoreAddOnProps {
     /**
      * Policy document required by the add-on to allow it to interact with AWS resources
      */
-    readonly policyDocument?: PolicyDocument;
+    readonly policyDocument?: {(partition: string): object};
```
shapirov103, line 20: Since it is a function let's call it policyDocumentProvider. Also why do we need {} around it? We can drop.

The last point: we can return the PolicyDocument from this function since it is seems the only thing we can do with it.

```
readonly policyDocumentProvider?: (partition: string) => PolicyDocument;
```

lib/addons/core-addon/index.ts
```diff
@@ -39,8 +39,9 @@ export class CoreAddOn implements ClusterAddOn {
         let serviceAccountRoleArn: string | undefined = undefined;
 
         if (this.coreAddOnProps?.policyDocument) {
+            const policyDoc = PolicyDocument.fromJson(this.coreAddOnProps.policyDocument(clusterInfo.cluster.stack.partition));
```
shapirov103, line 42: If this is the only way to use the policy document, let's make the policyProvider function return PolicyDocument as opposed to some json.

shapirov103 (commented):
Awesome PR, thank you for this!

A few comments, mostly coding style related.

camba1:
Thanks for the comments @shapirov103 ! All changes have been made and committed

shapirov103 (approved):
LGTM

Issue #233, opened 2021-11-09
EFS CSI Driver


kcoleman731:
Add support for installing the EFS CSI Driver via Helm

https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

youngjeong46:
Closing the issue as EFS CSI Driver is implemented as an addon.

Issue #21, opened 2021-05-25
[ADDON] - Amazon EFS CSI Driver


kcoleman731:
Add support for EFS CSI via an add-on

https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html

camba1:
duplicate of #233

youngjeong46:
Closing as finished.

PR #310, opened 2022-03-15
Feature: EFS CSI Driver add-on

camba1:
*Issue #233 and #21 (Dups)

*Description of changes:*

- Added a new EFS CSI Driver add-on
- Moved the container registry list from the aws LoadBalancers add-on (aws-loadbalancer-controller/registryMap.ts) to  utils/registry-utils so that it may be shared by multiple add-ons (including the load balancer and the efs csi driver)
- Added the new EFS CSI add-on to examples.ts along with references to the appropriate govCloud AMIs so that it can be deployed to govCloud
- 


docs/addons/efs-csi-driver.md
```diff
@@ -0,0 +1,55 @@
...
+A couple of things to note:
+
+- Driver is not compatible with Windows-based container images
+- The number of replicas to be deployed must be less or equal to the number of nodes in the cluster
+
+For more information on the driver, please review the [user guide](https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html).
+
+## Prerequisites
+
+- The EFS file system itself must be created in AWS separately as the driver uses the EFS for storage, but it does not create it.
```
shapirov103, line 15: For a typical use case, do you think it makes sense to create the actual FS as well? Would it be a valuable feature? 
If yes, we can add it as a follow-up feature. this looks good.

docs/addons/efs-csi-driver.md
```diff
@@ -0,0 +1,55 @@
...
+
+ssp.EksBlueprint.builder()
+    .addOns(new ssp.EfsCsiDriverAddOn() )
+    .build(app, 'my-cluster');
+```
+
+## Configuration Options
+
+- `version`: Version of the EFS CSI Driver add-on to be installed. Version 2.2.3 will be installed by default if a value is not provided
+- `replicaCount`: Number of replicas to be deployed. If not provided, two replicas will be deployed. Note that the number of replicas
```
shapirov103, line 33: can we assess minSize on the node group to determine the low boundary for the replicas? What if somebody rolls out a 1 node cluster as an experiment?

lib/addons/efs-csi-driver/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        EfsDriverPolicyDocument().Statement.forEach((statement) => {
+            serviceAccount.addToPrincipalPolicy(iam.PolicyStatement.fromJson(statement));
+        });
+
+        // Lookup appropriate image repo
+        const repo = registries.get(clusterInfo.cluster.stack.region) + EFS_REGISTRY_SUFFIX;
+
+        // Define chart
+        const efsCsiDriverChart = this.addHelmChart(clusterInfo, {
+            clusterName: cluster.clusterName,
```
shapirov103, line 66: Since the generic helm props allow passing values from the user, we generally merge whatever user passed with the values that we generated, with user passed value taking precedence. 

I observe quite a bit of options here: https://github.com/kubernetes-sigs/aws-efs-csi-driver

Example of value merging is for example here: https://github.com/aws-quickstart/ssp-amazon-eks/blob/main/lib/addons/appmesh/index.ts#L100

shapirov103 (commented):
Looks very good! A couple of comments, should be close to merge shortly.

docs/addons/efs-csi-driver.md
camba1: I think this would be useful, but it should be optional in case they have already created the EFS

lib/addons/efs-csi-driver/index.ts
camba1: Oh....yeah.....my bad, I' fix it

docs/addons/efs-csi-driver.md
camba1: Fair point. but should it check min size or desired size?

camba1: @shapirov103 How can I get the minsize (or desired size) from within the add-on? Also, what if they are using Karpenter ? As far as I know Karpenter does not use node groups and just limits on [CPUs or Memory](https://karpenter.sh/v0.7.2/tasks/set-resource-limits/)

docs/addons/efs-csi-driver.md
shapirov103: For sure. I am just thinking about an enterprise use case. E.g. they want to deploy an app that is using a file system. This will roll out to dev/test/prod. If we don't supply this feature, will it mean that someone will have to roll it out by accessing those account directly? I am thinking of a workaround that would leverage a custom `spi.ResourceProvider` in the platform to do this with https://docs.aws.amazon.com/cdk/api/v1/docs/aws-efs-readme.html.

Let's create an issue, I am going to get interns and will assign it to them.

docs/addons/efs-csi-driver.md
shapirov103: @camba1 fair point :).  I see that once nodegroup is created you cannot really get the minSize parameter. We can get all the config options from the blueprint/add-on configuration but it is not going to be worth the effort. The current approach is simple and straightforward.

shapirov103 (approved):
LGTM

PR #360, opened 2022-04-15
initial push for falco add-on

parkand1:
Initial push for Falco add-on


lib/addons/falco/index.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+ const defaultProps: HelmAddOnProps & FalcoAddOnProps = {
+    name: "blueprints-falco-addon",
+    namespace: "falco",
+    chart: "falco",
+    version: "1.17.4",
+    release: "falco",
+    repository:  "https://github.com/falcosecurity/charts",
+    createNamespace: true,
+    values: {
```
shapirov103, line 45: We generally don't set defaults for the values. You have properties like kubernetesSupportEnabled, etc. at the root level. Defaults should follow the same logic.

lib/addons/falco/index.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * populateValues populates the appropriate values used to customize the Helm chart
+ * @param helmOptions User provided values to customize the chart
+ */
+function populateValues(helmOptions: FalcoAddOnProps): Values {
+    const values = helmOptions.values ?? {};
+
+    setPath(values, "kubernetes.support.enabled",  helmOptions.kubernetesSupportEnabled);
```
shapirov103, line 76: The value may not be provided by the customer as `kubernetesSupportEnabled` is optional. You will overwrite your own default in the values. A better approach is to check for null and applying a default if not defined:

```
setPath(values, "kubernetes.support.enabled", helmOptions.kubernetesSupportEnabled ?? true); // or whatever the default is.
```

package.json
```diff
@@ -27,6 +27,7 @@
         "typescript": "~4.5.4"
     },
     "dependencies": {
+        "@aws-quickstart/eks-blueprints": "^0.1.4",
```
shapirov103, line 30: this is already in eks-blueprints, why do we need this dependency?

shapirov103 (commented):
Looks great, a few comments.

lib/addons/falco/index.ts
parkand1: addressed

lib/addons/falco/index.ts
parkand1: addressed

package.json
parkand1: removed - I ran into some weird dependency issues when i was testing but it ended up just being an issue with the version of CDK I was using

PR #361, opened 2022-04-17
Karpenter Bug Fix

youngjeong46:
*Issue #, if available:*

*Description of changes:* 
- Karpenter bug fix. 
- Upgrades to latest (v8.2), including subnet/security group tagging
- Provisioner Specs are Optional

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -1,8 +1,9 @@
 import { Construct } from "constructs";
-import { Role, ManagedPolicy, ServicePrincipal, CfnInstanceProfile, PolicyDocument } from 'aws-cdk-lib/aws-iam';
+import { Tags, CfnJson } from "aws-cdk-lib";
+import * as iam from 'aws-cdk-lib/aws-iam';
 import { ClusterInfo } from '../../spi';
 import { HelmAddOn, HelmAddOnProps, HelmAddOnUserProps } from '../helm-addon';
-import { createNamespace, setPath, createServiceAccount, conflictsWith, tagSubnets } from '../../utils';
+import { createNamespace, setPath, conflictsWith, tagSubnets } from '../../utils';
 import { KarpenterControllerPolicy } from './iam';
 
 /**
@@ -11,13 +12,27 @@ import { KarpenterControllerPolicy } from './iam';
 interface KarpenterAddOnProps extends HelmAddOnUserProps {
     /**
      * Specs for a Provisioner (Optional) - If not provided, the add-on will
-     * deploy without a Provisioner.
+     * deploy a Provisioner with default values.
      */
     ProvisionerSpecs?: { 
-        'node.kubernetes.io/instance-type': string[],
-        'topology.kubernetes.io/zone': string[],
-        'kubernetes.io/arch': string[],
-        'karpenter.sh/capacity-type': string[],
+        'node.kubernetes.io/instance-type'?: string[],
+        'topology.kubernetes.io/zone'?: string[],
+        'kubernetes.io/arch'?: string[],
+        'karpenter.sh/capacity-type'?: string[],
+    }
+
+    /**
+     * Tags needed for subnets - If not provided, users will need to manually provide them.
+     */
+     subnetTags?: { 
+        [key: string]: string
+    }
+
+    /**
+     * Tags needed for security groups - If not provided, users will need to manually provide them.
+     */
+     securityGroupTags?: { 
+        [key: string]: string
     }
 }
 
@@ -29,7 +44,7 @@ const KARPENTER = 'karpenter';
 const defaultProps: HelmAddOnProps = {
     name: KARPENTER,
     namespace: KARPENTER,
-    version: '0.5.3',
+    version: '0.8.2',
     chart: KARPENTER,
     release: "blueprints-addon-karpenter",
     repository: 'https://charts.karpenter.sh',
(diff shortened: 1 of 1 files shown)


lib/addons/karpenter/index.ts
```diff
@@ -54,23 +69,66 @@ export class KarpenterAddOn extends HelmAddOn {
...
+        
+        // Tag VPC Security Group
+        if (sgTags){
+            Object.entries(sgTags).forEach(
+                ([key,value]) => {
+                    tagSecurityGroup(cluster.stack, cluster.clusterSecurityGroupId, key,value);
+            });
+        }
+        
+        const conditions = new CfnJson(cluster, "ConditionPlainJson", {
```
shapirov103, line 92: Statements from line 92 to line 116 is just a registration of an IRSA account the hard way. 

Why not just use standard IRSA service account support in CDK, i.e. 

```
const sa = cluster.addServiceAccount(roleName, opts);
sa.role.addManagedPolicy()...
```

This automatically creates OIDC provider and trust relationship.

lib/addons/karpenter/index.ts
```diff
@@ -82,33 +140,36 @@ export class KarpenterAddOn extends HelmAddOn {
...
-
-        karpenterChart.node.addDependency(sa);
-
-        // (Optional) default provisioner - defaults to 30 seconds for scale down for
-        // low utilization
-        if (this.options.ProvisionerSpecs){
+        setPath(values, "clusterEndpoint", endpoint);
+        setPath(values, "clusterName", name);
+        values['serviceAccount'] = {
+            create: true,
```
shapirov103, line 150: If we use standard IRSA account we should be able to state false here and just pass the sa name. We can still use this construct if for whatever reason we want karpenter to create its own service account, but worth a try.  I see nothing special in their helm chart that would require a different approach, but the test will tell: https://github.com/aws/karpenter/blob/main/charts/karpenter/templates/serviceaccount.yaml

lib/utils/vpc-utils.ts
```diff
@@ -2,14 +2,37 @@ import * as ec2 from 'aws-cdk-lib/aws-ec2';
...
- * @param subnetArns 
- * @param tag 
- * @param value 
+ * @param stack - CDK Stack
+ * @param subnets - a list of subnets
+ * @param key - Tag Key
+ * @param value - Tag Value
  */
-export function tagSubnets(stack: Stack, subnets: ec2.ISubnet[], tag: string, value: string): void {
+export function tagSubnets(stack: Stack, subnets: ec2.ISubnet[], key: string, value: string): void {
```
shapirov103, line 35: Is there an issue against CDK to support this tagging without custom resources? I recall you mentioned something related to it before. 
If there is no issue:
1. let's create an issue against CDK (customers are very reluctant to see lambda functions in their accounts)
2. let's create an issue against this repo to update this logic once CDK adds native support.

If there an issue already, then only 2/ applies.

shapirov103 (changes requested):
I see a lot of work that you put into it, it is awesome and great contribution! I believe I missed the initial review, hence my comments. I believe we can simplify this add-on a bit. Looks good overall!

lib/addons/karpenter/index.ts
youngjeong46: The problem is the SA role needs to follow a naming convention in the form of `${clusterName}-karpenter`. addServiceAccount autogenerates the role name. This is something I've considered opening an issue for against CDK.

lib/addons/karpenter/index.ts
```diff
@@ -11,13 +12,27 @@ import { KarpenterControllerPolicy } from './iam';
 interface KarpenterAddOnProps extends HelmAddOnUserProps {
     /**
      * Specs for a Provisioner (Optional) - If not provided, the add-on will
-     * deploy without a Provisioner.
+     * deploy a Provisioner with default values.
      */
     ProvisionerSpecs?: {
```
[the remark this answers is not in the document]
youngjeong46: addressed

lib/addons/karpenter/index.ts
youngjeong46: Addressed - used our own `createServiceAccount`

lib/addons/karpenter/index.ts
youngjeong46: Addressed, somewhat related to lines 92-116

lib/utils/vpc-utils.ts
youngjeong46: I think I raised an issue, let me confirm. It had to do with tokens not resolving for us to use CDK native Tags. I will create an issue against this repo as well. Consider it addressed.

docs/addons/karpenter.md
```diff
@@ -23,7 +23,23 @@ import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
 
-const addOn = new blueprints.addons.KarpenterAddOn();
+const karpenterAddonProps = {
+  ProvisionerSpecs: {
```
shapirov103, line 27: This needs an update as well to match the code.

lib/addons/karpenter/index.ts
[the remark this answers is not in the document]
shapirov103: Don't forget the docs update because of this API change.

lib/addons/karpenter/index.ts
```diff
@@ -82,33 +116,39 @@ export class KarpenterAddOn extends HelmAddOn {
...
-
-        // (Optional) default provisioner - defaults to 30 seconds for scale down for
-        // low utilization
-        if (this.options.ProvisionerSpecs){
+        setPath(values, "clusterEndpoint", endpoint);
+        setPath(values, "clusterName", name);
+        values['serviceAccount'] = {
+            create: false,
+            name: RELEASE,
+            annotations: {
```
shapirov103, line 131: most likely the annotations are ignored if create is false.

docs/addons/karpenter.md
youngjeong46: addressed

lib/addons/karpenter/index.ts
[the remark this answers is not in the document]
youngjeong46: done

lib/addons/karpenter/index.ts
```diff
@@ -54,23 +69,66 @@ export class KarpenterAddOn extends HelmAddOn {
         const cluster = clusterInfo.cluster;
         const values = { ...this.props.values ?? {} };
```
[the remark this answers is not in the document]
youngjeong46: deep merged before the helm chart creation.

shapirov103 (approved):
LGTM! Awesome stuff!

Issue #367, opened 2022-04-20
[FEATURE][ADDONS][Cluster-Autoscaler] - Cluster Autoscaler should migrate to use IRSA and not instance profile


tsahiduek:
Currently cluster-autoscaler rely on instance-profile to interact with AWS resources (autoscaling & ec2). 
Per Amazon EKS [docs](https://docs.aws.amazon.com/eks/latest/userguide/autoscaling.html), the preferred method is to use IRSA (IAM Roles for ServiceAccount)

youngjeong46:
PR is merged. Closing the issue.

PR #368, opened 2022-04-22
Implement Cluster Autoscaler Add-on with IRSA

youngjeong46:
*Issue #, if available:* #367

*Description of changes:* Leverage IRSA to implement Cluster Autoscaler.


lib/addons/cluster-autoscaler/index.ts
```diff
@@ -72,30 +76,39 @@ export class ClusterAutoScalerAddOn extends HelmAddOn {
...
-            cloudProvider: 'aws',
-            autoDiscovery: {
-                clusterName: cluster.clusterName
-            },
-            awsRegion: clusterInfo.cluster.stack.region
-        });
+        // Create IRSA
+        const sa = createServiceAccount(cluster,
+            RELEASE, 
+            "kube-system",
```
shapirov103, line 98: This should use the namespace provided by the customer.

lib/addons/cluster-autoscaler/index.ts
```diff
@@ -72,30 +76,39 @@ export class ClusterAutoScalerAddOn extends HelmAddOn {
...
+            autoscalerPolicyDocument
+        );
+        
+        // Create Helm Chart
+        setPath(values, "cloudProvider", "aws");
+        setPath(values, "autoDiscovery.clusterName", cluster.clusterName);
+        setPath(values, "awsRegion", cluster.stack.region);
+        setPath(values, "rbac.serviceAccount.create", false);
+        setPath(values, "rbac.serviceAccount.name", RELEASE);
+        setPath(values, "rbac.serviceAccount.annotations", {"eks.amazonaws.com/role-arn": sa.role.roleArn});
```
shapirov103, line 108: Is not this annotation automatically created for the service account?

shapirov103 (commented):
Looks great a couple of small comments, one is important. I will wait for the change before kicking off the e2e test.

lib/addons/cluster-autoscaler/index.ts
youngjeong46: addressed

lib/addons/cluster-autoscaler/index.ts
youngjeong46: removed.

lib/addons/cluster-autoscaler/index.ts
```diff
@@ -62,8 +64,12 @@ export class ClusterAutoScalerAddOn extends HelmAddOn {
 
         const cluster = clusterInfo.cluster;
         const nodeGroups = assertEC2NodeGroup(clusterInfo, "Cluster Autoscaler");
+        const values = this.options.values || {};
+        const namespace = this.options.namespace!;
+        const createANamespace = (this.options.namespace == "kube-system") ? false : true;
```
shapirov103, line 69: A typical approach from the customers is to create a namespace like `addons` and put all those installed by the platform in there. This will not work in such cases, since the namespace will be reused across multiple add-ons. This is a bigger problem though, not only applicable in this case. 

The correct approach is:
1. add field `createNamespace` to the properties.
2. If set to true - create, otherwise assume it is already there.

minor: createANamespace => createNamespace

shapirov103 (commented):
Almost there, one comment

shapirov103 (approved):
LGTM
Will include in the next minor release

PR #374, opened 2022-05-05
Initial PR for Istio add-on

edgarsilva948:
This PR implements two new add-ons, one for Istio Base (which installs istio's CRDs) and the other for Istio Control Plane.


lib/addons/istio-control-plane/index.ts
```diff
@@ -0,0 +1,42 @@
+import { Construct } from "constructs";
+import merge from "ts-deepmerge";
+import { ClusterInfo } from "../../spi";
+import { HelmAddOn, HelmAddOnUserProps } from "../helm-addon";
+import { dependable } from '../../utils';
+import { ValuesSchema } from "./values";
+
+export interface IstioControlPlaneAddOnProps extends Omit<HelmAddOnUserProps, "namespace"> {
```
shapirov103, line 8: Why omit namespace?

docs/addons/istio-base.md
```diff
@@ -0,0 +1,64 @@
...
+serviceentries.networking.istio.io           2022-05-05T02:16:23Z
+sidecars.networking.istio.io                 2022-05-05T02:16:23Z
+telemetries.telemetry.istio.io               2022-05-05T02:16:23Z
+virtualservices.networking.istio.io          2022-05-05T02:16:23Z
+wasmplugins.extensions.istio.io              2022-05-05T02:16:23Z
+workloadentries.networking.istio.io          2022-05-05T02:16:23Z
+workloadgroups.networking.istio.io           2022-05-05T02:16:23Z
+```
+
+Note that the all the CRDs are created in the `istio-system` namespace.
```
shapirov103, line 47: this is correct with respect to the default add-on configuration. However, unlikes the istio-control-plane add-on, namespace for this add-on can also be passed in by the user. If that is allowed, then this statement should be changed to the "provided" namespace (istio-system by default). 
If it is not allowed, let's discuss.

docs/addons/istio-control-plane.md
```diff
@@ -0,0 +1,82 @@
...
+ - `values`: Arbitrary values to pass to the chart as per https://istio.io/v1.4/docs/reference/config/installation-options/
+
+```typescript
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import * as blueprints from '@edgarsilva948/eks-blueprints';
+
+const app = new cdk.App();
+
+const IstioControlPlaneAddOnProps = {
```
shapirov103, line 51: minor: const / variable name should be lower-camel-case based on the [convention](https://google.github.io/styleguide/tsguide.html#identifiers) that we use.

docs/addons/istio-control-plane.md
```diff
@@ -0,0 +1,82 @@
...
+  values: {
```
shapirov103, line 52: This is great!

shapirov103 (commented):
Looks great overall! A few minor comments. 
We need to test it before we can merge. Do you see any issue if it is added to the test blueprint [here]9https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts) so that I can kick off end2end tests?

shapirov103:
@edgarsilva948 following up on my previous comment, please take a look at the few comments when you get a chance,

lib/addons/istio-control-plane/index.ts
edgarsilva948: this addon is dependent on istio-base, so I wanted to make sure that different namespaces would not be informed in each of them. But, I'm looking again, if the istio-base namespace is changed, it will fail just the same :)
I will remove the omit.

edgarsilva948:
> Looks great overall! A few minor comments. We need to test it before we can merge. Do you see any issue if it is added to the test blueprint [here]9https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts) so that I can kick off end2end tests?

I don't see any issue.

docs/addons/istio-base.md
edgarsilva948: Fixed

docs/addons/istio-control-plane.md
edgarsilva948: Fixed

docs/addons/istio-control-plane.md
edgarsilva948: :D

shapirov103 (commented):
@edgarsilva948 looks great but we need to test it. The way it works for the end-to-end test is by adding the add-on to the test blueprint [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts).  I assume both add-ons should be added.

If you add a commit with that change, I will kick off the e2e test and merge if it passes. Some more info on internal process is [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/internal/readme-internal.md#triggering-e2e-testing) (it can surely be improved).

shapirov103 (commented):
@edgarsilva948 were the issue template files removed on purpose?

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -1,100 +0,0 @@
----
-name: "🐛 Bug Report"
```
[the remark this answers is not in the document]
edgarsilva948: It wasn't intentional. I'm trying to understand what happened.

shapirov103 (dismissed):
LGTM

shapirov103 (approved):
LGTM

Issue #384, opened 2022-05-15
Pipelines: cross-account deployment needs shared KMS


youngjeong46:
### Describe the bug

I'm leveraging cross-account deployment using pipelines as follows, and have bootstrapped with the correct permissions:

```
blueprints.CodePipelineStack.builder()
      .name("eks-blueprints-pipeline-test")
      .owner("youngjeong46")
      .repository({
          repoUrl: 'eks-blueprints-pipeline-test',
          credentialsSecretName: 'github-token',
          targetRevision: 'main'
      })
      .wave({
        id: 'dev',
        stages: [
          { id: "dev-1", stackBuilder: blueprint.clone('us-east-1')},
          { id: "dev-2", stackBuilder: blueprint.clone('us-west-2')},
        ]
      })
      .wave({
        id: "prod",
        stages: [
          { id: "west-1", stackBuilder: blueprint.clone('us-east-1', prodAccount)},
        ]
      })
      .build(scope, id+'-stack', props);
  }
```

When I try deploying, the cdk produces the error needing either a bucket with a key or a shared KMS cross-account key (needed when creating the cross-account Artifact bucket).

### Expected Behavior

Deploy as intended (dev into the current account, and prod to a designated prod account)

### Current Behavior

The following error occurs:

```
/Users/yojeo/src/kubernetes/eks-blueprints-pipeline-test/node_modules/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts:726
        throw new Error(
              ^
Error: Artifact Bucket must have a KMS Key to add cross-account action 'Prepare' (pipeline account: '661554271967', action account: '181460066119'). Create Pipeline with 'crossAccountKeys: true' (or pass an existing Bucket with a key)
    at Pipeline.getRoleFromActionPropsOrGenerateIfCrossAccount (/Users/yojeo/src/kubernetes/eks-blueprints-pipeline-test/node_modules/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts:726:15)
    at Pipeline.getRoleForAction (/Users/yojeo/src/kubernetes/eks-blueprints-pipeline-test/node_modules/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts:698:27)
    at Pipeline._attachActionToPipeline (/Users/yojeo/src/kubernetes/eks-blueprints-pipeline-test/node_modules/aws-cdk-lib/aws-codepipeline/lib/pipeline.ts:529:29)
[...]

PR #388, opened 2022-05-18
bugfix: add crossAccountKeys boolean flag for Pipelines

jyidiego:
…MS key for artifact bucket

*Issue #384

*Description of changes:
Added method to set crossAccountKeys to true. This method sets [aws-cdk-lib » pipelines » CodePipeline](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.pipelines.CodePipeline.html#crossaccountkeys) constructor property crossAccountKeys to true. otherwise it is false if not called.

### DIFF
--- a/lib/pipelines/code-pipeline.ts
+++ b/lib/pipelines/code-pipeline.ts
@@ -48,6 +48,11 @@ export type PipelineProps = {
     owner: string;
 
     /**
+     * Enable/Disable allowing cross-account deployments.
+     */
+    crossAccountKeys: boolean;
+
+    /**
      * Repository for the pipeline
      */
     repository: GitHubSourceRepository;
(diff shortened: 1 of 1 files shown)


youngjeong46 (commented):
Hello John, thanks for the contributions (again)!

Our e2e doesn't cover pipeline testing - I can approve if you have done e2e testing on your end. It's also a good idea to contribute a pattern [here](https://github.com/aws-samples/cdk-eks-blueprints-patterns) - something like "multi-account-construct" that shows how we can use this flag.

shapirov103:
@jyidiego thank you for your contribution. 
Do you mind adding a small paragraph to the docs on this topic (docs/pipelines.md)?

Also I suppose we can merge this, however let's at least add a unit test on it to the test/stack.test.ts. We have one test on pipelines, we can add another or modify existing.

shapirov103:
This will be merged but the fix based on our standards with the commit is in #410

shapirov103 (approved):
Merged with the follow-up in #410

PR #410, opened 2022-06-09
Kms fix

youngjeong46:
*Issue #, if available:*

#384 

*Description of changes:*

Added some additional items from PR #388:
- Unit test
- Usage ID tagging for anyone that leverages multi-account deployment.

### DIFF
--- a/lib/pipelines/code-pipeline.ts
+++ b/lib/pipelines/code-pipeline.ts
@@ -182,13 +182,18 @@ export class CodePipelineBuilder implements StackBuilder {
 export class CodePipelineStack extends cdk.Stack {
 
     static readonly USAGE_ID = "qs-1s1r465k6";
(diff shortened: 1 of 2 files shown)


lib/pipelines/code-pipeline.ts
```diff
@@ -172,13 +182,18 @@ export class CodePipelineBuilder implements StackBuilder {
 export class CodePipelineStack extends cdk.Stack {
 
     static readonly USAGE_ID = "qs-1s1r465k6";
+    static readonly USAGE_ID_MULTI_ACCOUNT = "qs-1s1r465f2";
```
shapirov103, line 185: Where is that usage tracking ID from? I do recall we had a range, but once a new identifier is leveraged we need to communicate it to the quickstart team. I assume it is leveraged in reports.

shapirov103 (approved):
Looks great, LGTM

PR #424, opened 2022-06-23
Testing my first pull request

Eli1123:
*Issue #, if available:* 401

*Description of changes: I added a test case to catch when the length of the blueprint props id (managed node group id name) exceeds 63 characters.


my-blueprints/.gitignore
```diff
@@ -0,0 +1,8 @@
+*.js
```
shapirov103, line 1: Please drop `my-blueprints` folder entirely. Unless you have a clear reasoning we need it.

package.json
```diff
@@ -12,22 +12,24 @@
...
-        "jest": "^27.4.7",
-        "ts-jest": "^27.1.1",
+        "jest": "^27.5.1",
+        "ts-jest": "^27.1.5",
         "ts-node": "^10.4.0",
         "typedoc": "^0.22.15",
         "typescript": "~4.5.4"
     },
     "dependencies": {
+        "@aws-quickstart/eks-blueprints": "^1.0.1",
```
shapirov103, line 32: remove this line, why do we need it?

test/newNameLength.test.ts
```diff
@@ -0,0 +1,67 @@
...
+const addOns: Array<blueprints.ClusterAddOn> = [
+  new blueprints.addons.MetricsServerAddOn(),
+  new blueprints.addons.ClusterAutoScalerAddOn(),
+  new blueprints.addons.VpcCniAddOn(),
+];
+
+//This makes an array result to hold pre made default blueprints without the build yet added for testing purposes.
+function getConstrainstDataSet(): BlueprintBuilder[] {
+  const result: BlueprintBuilder[] = [];
+  const blueprint1 = blueprints.EksBlueprint.builder().addOns(...addOns);
```
shapirov103, line 21: please add specific data variations for each blueprint builder returned. They are identical atm.

test/newNameLength.test.ts
```diff
@@ -0,0 +1,67 @@
...
+//below you build each example testing a given exception you want to see thrown.
+If I end up needing a for each loop for this here is a draft. 
+var i = 0;
+getConstrainstDataSet().forEach(e => {expect(e.build(app, 'blueprint'))).toThrow(...);});
+*/
+
+//Below is my first test to deal with a node group name being longer than 63 characters 
+
+test("Testing length error!", () => {
+  //try {
```
shapirov103, line 43: please remove commented code (and anything superfluous)

test/newNameLength.test.ts
```diff
@@ -0,0 +1,67 @@
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import * as blueprints from '../lib';
+import { kMaxLength } from 'buffer';
```
shapirov103, line 4: make sure `make lint` passed. Unused imports, etc.

shapirov103 (changes requested):
a few comments

lib/addons/helm-addon/index.ts
```diff
@@ -23,4 +25,20 @@ export abstract class HelmAddOn implements spi.ClusterAddOn {
...
     }
-}
\ No newline at end of file
+}
+//below is my addon approach method for when I get too it. Leaving for later
+function generalAddonConstraintsValidation(props: HelmAddOnProps){
+    try{
+        const test = z.string().max(addonConstraints.addonNameMaxLength);
+        console.log(props.name + "-=-=-=-=-=")
+        test.parse(props.name);
```
shapirov103, line 34: Please address the general GitHun actions warnings.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -54,9 +53,16 @@ export class EksBlueprintProps {
      * Control Plane log types to be enabled (if not passed, none)
      * If wrong types are included, will throw an error.
      */
-    readonly enableControlPlaneLogTypes?: string[];
+    readonly enableControlPlaneLogTypes?: enabledControlPlaneLogTypes[];
 }
 
+export enum enabledControlPlaneLogTypes {
```
shapirov103, line 59: enum is a data type. Should follow the noun semantics. It does NOT represent enabled log types, they are just available control plane log types. So should be ControlPlaneLogTypes (starting with upper case). 

For more information on naming conventions please read: https://google.github.io/styleguide/tsguide.html

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -281,4 +289,47 @@ export class EksBlueprint extends cdk.Stack {
             });
         }
     }
-}
\ No newline at end of file
+}
+function generalConstraintsValidation(blueprintProps: EksBlueprintProps) {
```
shapirov103, line 293: Same feedback on the naming: verb, general => input

lib/addons/helm-addon/index.ts
```diff
@@ -23,4 +25,20 @@ export abstract class HelmAddOn implements spi.ClusterAddOn {
         const chart = {...this.props, ...{ values, wait, timeout, createNamespace} };
         return kubectlProvider.addHelmChart(chart);
     }
-}
\ No newline at end of file
+}
+//below is my addon approach method for when I get too it. Leaving for later
+function generalAddonConstraintsValidation(props: HelmAddOnProps){
```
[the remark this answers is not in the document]
shapirov103: Not addressed?

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,34 @@
+import { z, ZodError, ZodIssueCode } from "zod";
+
+export interface Constraint {
```
shapirov103, line 3: just FYI: add jsdoc to each exported type, interface and function

test/newNameLength.test.ts
```diff
@@ -26,10 +27,10 @@ const longName = 'eks-blueprint-with-an-extra-kind-of-very-long-name-that-is-def
 
 test("Testing ID length error!", () => {
   expect(() => getConstraintsDataSet().at(1)?.build(app, longName))
-    .toThrow!('Managed Node Groups ID must be no more than 63 characters long!');
+    .toThrow(z.ZodIssueCode.too_big);
```
shapirov103, line 30: We need to see the actual field in the error message. Let's update the test in a way that will include validation of the field. You can pass RegExp to the toThrow() to match the string.

test/newNameLength.test.ts
```diff
@@ -2,6 +2,7 @@ import 'source-map-support/register';
 import * as cdk from 'aws-cdk-lib';
```
shapirov103, line 2: Let's rename this file to `contraints-validation.test.ts`

my-blueprints/.gitignore
Eli1123: I removed it now

package.json
Eli1123: I removed it it was not needed

test/newNameLength.test.ts
Eli1123: I have it that way for testing purposes and would remove duplicates when I am done. I just like having a few there in the situation I just needed one or two more to test.

test/newNameLength.test.ts
Eli1123: I am now being sure to check that

lib/addons/helm-addon/index.ts
Eli1123: I have gone through them now I believe.

lib/stacks/eks-blueprint-stack.ts
Eli1123: That is a very helpful link I will be sure to reference it in the future. Also I have fixed that now.

lib/stacks/eks-blueprint-stack.ts
Eli1123: Fixed now.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -281,4 +289,47 @@ export class EksBlueprint extends cdk.Stack {
...
+        
+        if(blueprintProps.name != undefined)
+        testMin.parse(blueprintProps.name)
+    } catch (e) {
+        throw new Error('Managed Node Groups name must be no less than 1 character long!')
+    }
+}
+
+var generalConstraints = {
+    blueprintIDMax: 63,
```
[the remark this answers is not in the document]
Eli1123: We have addressed this with our generic approach.

package.json
```diff
@@ -12,17 +12,18 @@
         "lint": "npx eslint . --ext .js,.jsx,.ts,.tsx"
     },
     "devDependencies": {
+        "@babel/preset-typescript": "^7.17.12",
```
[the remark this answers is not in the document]
Eli1123: I think this came from trying to get something installed to help with what I was doing and I did not need it after all. Now I will be sure to look back on properly removing them if I am to do something similar in the future.

test/newNameLength.test.ts
shapirov103: Looks like the name stayed the same.

test/newNameLength.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const addOns: Array<blueprints.ClusterAddOn> = [
+  new blueprints.addons.MetricsServerAddOn(),
+  new blueprints.addons.ClusterAutoScalerAddOn(),
+  new blueprints.addons.VpcCniAddOn(),
+];
+
+function getConstraintsDataSet(): BlueprintBuilder[] {
+  const result: BlueprintBuilder[] = [];
+  const blueprint1 = blueprints.EksBlueprint.builder().addOns(...addOns).name("eks-blueprint-with-an-extra-kind-of-very-long-name-that-is-definitely-not-going-to-work");
```
shapirov103, line 17: You have a constant `longName` - should it be used here?

test/newNameLength.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  result.push(blueprint1);
+  result.push(blueprint2);
+
+  return result;
+}
+
+const longName = 'eks-blueprint-with-an-extra-kind-of-very-long-name-that-is-definitely-not-going-to-work';
+
+test("Testing ID length error!", () => {
```
shapirov103, line 28: We should not access the array with random access. The point of having the array was to iterate over it and validate each permutation. We discussed that since error message are different in each case you want to return an array of tuples `[blueprint, expectedError]`

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -54,9 +53,22 @@ export class EksBlueprintProps {
      * Control Plane log types to be enabled (if not passed, none)
      * If wrong types are included, will throw an error.
      */
-    readonly enableControlPlaneLogTypes?: string[];
+    readonly enableControlPlaneLogTypes?: ControlPlaneLogTypes[];
 }
 
+const blueprintPropsContraints: ConstraintsType<Partial<EksBlueprintProps>> = {
```
shapirov103, line 59: should not this be `ContraintsType<EksBlueprintProps>`? Why partial again?

test/newNameLength.test.ts
Eli1123: I just changed it this morning and have not committed yet

lib/stacks/eks-blueprint-stack.ts
Eli1123: Not sure, might have been when trying to fix the issue that putting Partial around Record in ConstraintsType fixed and I never removed it.

test/newNameLength.test.ts
Eli1123: Ya this was the last thing I was trying to figure out yesterday. I was trying to find a way to extract the "Message" parameter from the ZodError but was not having luck with that. I did find .too_big worked to a degree but not what I am looking for. 
Like how below the custom error messages I have. If you see ZodErrors its within there but not having luck extracting it.
`z.string() .min(this.min ?? 0, { message: "Managed Node Groups " + key + " must be no less than 1 characters long!" })
`
I figured if I get that to work I can easily have the errors send out without the need for try-catch blocking each section to throw the errors.

And I am not quite sure how RegExp can help here? I have a little Regex experience from Java which from a google RegExp is another way to say it. But Regexp passed into toThrow to match the string? Doesn't toThrow simply check if the error thrown matches the given string? Why would you want to match inside of that?

lib/addons/helm-addon/index.ts
```diff
@@ -23,4 +25,20 @@ export abstract class HelmAddOn implements spi.ClusterAddOn {
         const chart = {...this.props, ...{ values, wait, timeout, createNamespace} };
         return kubectlProvider.addHelmChart(chart);
     }
-}
\ No newline at end of file
+}
+//below is my addon approach method for when I get too it. Leaving for later
+function validateInputAddonConstraints(props: HelmAddOnProps){
```
shapirov103, line 30: Why is this function here? We should drop it  and call the generic function from the contraint utils

lib/addons/helm-addon/index.ts
```diff
@@ -23,4 +25,20 @@ export abstract class HelmAddOn implements spi.ClusterAddOn {
...
+        const test = z.string().max(addonConstraints.addonNameMaxLength);
+        console.log(props.name + "-=-=-=-=-=");
+        test.parse(props.name);
+    }
+    catch(e){
+        throw new Error('Addon Names must be no more than 128 characters long!');
+    }
+}
+
+const addonConstraints = {
```
shapirov103, line 41: I assume scratchpad code? let's drop it and replace with the data structure that defines constraints with our constraints framework.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -54,9 +53,22 @@ export class EksBlueprintProps {
...
+    readonly enableControlPlaneLogTypes?: ControlPlaneLogTypes[];
 }
 
+const blueprintPropsContraints: ConstraintsType<EksBlueprintProps> = {
+    
+    id: new StringConstraint(1,63),
+    name: new StringConstraint(1,63)
+};
+
+export enum ControlPlaneLogTypes {
```
shapirov103, line 65: minor: Name of the type should be singular. `ControlPlaneLogType`

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const app = new cdk.App();
+
+const addOns: Array<blueprints.ClusterAddOn> = [
+  new blueprints.addons.MetricsServerAddOn(),
+  new blueprints.addons.ClusterAutoScalerAddOn(),
+  new blueprints.addons.VpcCniAddOn(),
+];
+
+function getConstraintsDataSet(): [BlueprintBuilder, ZodError][] {
```
shapirov103, line 17: Let's extract the tuple to a type to make it cleaner, the way I showed initially, e.g. 

```
type BlueprintBuilderError = [ BlueprintBuilder, ZodError ];

function getConstraintsDataSet(): BlueprintError[] { ...

File: test/constraints-validation.test.ts [hunk 22]
```diff
@@ -0,0 +1,73 @@
...
+  let result: [BlueprintBuilder, ZodError][] = [];
+  const blueprint1 = blueprints.EksBlueprint.builder().addOns(...addOns).id(longName);
+  
+  const blueprint1ZodError = new z.ZodError([{
```
shapirov103, line 21: You can use `ZodTooBigIssue` as the data type rather that a free from object. 

```typescript
const issue : Partial<ZodTooBigIssue> = {

}
```

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+test("For Each loop test.", () => {
+  getConstraintsDataSet().forEach((ex) => {
+    try{
+     ex[0].build(app, ex[0].props.id!);//be sure all tests built have a id defined before this!
+     expect(true).toBe(false);
+     
+    } catch(e) {
+      if(e instanceof z.ZodError){ //we are testing for code, maximum/minimum, and type
+        const eMax = (e.issues[0] as any).maximum;
```
shapirov103, line 56: 1. There is no need to have expression like `e.issues[0] as any` all over the code. Just extract a variable once. 
2. You can use a generic object comparison function to compare that all fields in object ex[1].issues[0] are equals to e.issues[0]. See more here: https://dmitripavlutin.com/how-to-compare-objects-in-javascript/#3-shallow-equality

shapirov103 (commented):
Some comments added, please review when you get a chance.

lib/addons/helm-addon/index.ts
Eli1123: That was from before we did the generic approach and I never deleted it, good point.

lib/addons/helm-addon/index.ts
Eli1123: Also from before we changed to the generic way and I haven't gone back to this file yet. I'll be sure to fix up this file and remove the old stuff.

lib/stacks/eks-blueprint-stack.ts
Eli1123: Is it singular since its called one at a time like ControlPlaneLogType.api and not a list of things after the dot? Even though it holds an array of various types and many can be called, but only one at a time?

test/constraints-validation.test.ts
Eli1123: Oh yes that is a cleaner way to do it.

test/constraints-validation.test.ts
Eli1123: I figured there must be a better way to do that after I did it, just got it functional first, now for some fine tuning. I'll be sure to look into those things and make it better.

test/constraints-validation.test.ts
Eli1123: After looking into this I think it is just as much code, and if not just slightly different. After adjusting my original solution to be more efficient I think its fine how it is.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -42,6 +43,66 @@ export interface GenericClusterProviderProps extends eks.ClusterOptions {
...
 
+export const managedNodeGroupContraints: ConstraintsType<ManagedNodeGroup> = {
+    /**
+     * id can be no less than 1 character long, and no greater than 63 characters long.
+     */
+    id: new StringConstraint(1, 63),
+    /**
+     * minSize has a maximum of 100 nodes per node group and can be as little as 0.
+     */
+    minSize: new NumberConstraint(0, 100),
```
shapirov103, line 54: TODO: set the correct constraint and reference to the source, e.g. AWS account limits.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -98,7 +159,13 @@ export class ClusterBuilder {
 export class GenericClusterProvider implements ClusterProvider {
 
     constructor(readonly props: GenericClusterProviderProps) {
-        assert(!(props.managedNodeGroups && props.managedNodeGroups.length > 0 
+
+        validateConstraints(props, genericClusterPropsContraints, GenericClusterProvider.name);
```
shapirov103, line 163: Move line 163 - 168 to a separate method `validateInput` (similar to what we have in `EKSBlueprintStack`)

Eli1123: That is a good call, much cleaner now.

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    code: "too_big",
+    maximum: 63,
+    type: "string",
+    inclusive: true,
+    path: [],
+    message: "Big"
+  }]);
+
+  const blueprint2 = blueprints.EksBlueprint.builder().addOns(...addOns).enableControlPlaneLogTypes(ControlPlaneLogType.authenticator).name(longName);
+  const blueprint2ZodError = new z.ZodError([{
```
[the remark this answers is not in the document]
Eli1123: That was one of my main goals for today after I fix what I made yesterday. I had enough time to make the validations, and such but today planned to refine that, get it functional. Then get its test cases and refine and make functional that part.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -54,9 +53,23 @@ export class EksBlueprintProps {
...
 
+const blueprintPropsContraints: ConstraintsType<EksBlueprintProps> = {
+    
+    id: new StringConstraint(1,63),
+    name: new StringConstraint(1,63)
+};
+
+export enum ControlPlaneLogType {
+    
+    api = 'api',
```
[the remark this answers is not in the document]
Eli1123: So like this if I am understanding it right? `export const enum CONTROL_PLANE_LOG_TYPE {
`

lib/addons/helm-addon/index.ts
```diff
@@ -1,26 +1,66 @@
...
+};
+
 export abstract class HelmAddOn implements spi.ClusterAddOn {
 
     props: HelmAddOnProps;
 
     constructor(props: HelmAddOnProps) {
         this.props = cloneDeep(props); // avoids polution when reusing the same props across stacks, such as values
+
+        validateConstraints([props], helmAddonPropsContraints, HelmAddOn.name);
```
shapirov103, line 55: why are you passing props as [props]? Your code is expected to handle arrays or individual objects, correct?

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+            z.number()
+                .gte(this.min ?? 1, { message: key + " (" + identifier + ": " + value + ") must be no less than " + this.min + " node groups!" })
+                .lte(this.max ?? 3, { message: key + " (" + identifier + ": " + value + ") must be no more than " + this.max + " node groups!" })
+                .parse(value.length);
+    }
+}
+
+export type ConstraintsType<T> = Partial<Record<keyof T, Constraint>>;
+
+export function validateConstraints<T>(object: any, constraints: ConstraintsType<T>, context: string) {
```
shapirov103, line 70: You can use a multi-arg support here similar to what we do [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/stacks/eks-blueprint-stack.ts#L116)

This enables you to pass either individual object or an array. The only caveat is  that ...arg must be last in the declaration.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -255,23 +269,22 @@ export class EksBlueprint extends cdk.Stack {
...
             result.add(key, value);
         }
 
         return result;
     }
-
-
     private validateInput(blueprintProps: EksBlueprintProps) {
         const teamNames = new Set<string>();
+        validateConstraints([blueprintProps], blueprintPropsContraints, EksBlueprintProps.name);//this .name gives the class name! Good to know
```
shapirov103, line 287: we should drop explicit array. Just validateConstrains(blueprintProps, ..._);

shapirov103 (commented):
Few comments. Also expecting the docs/internal/input-validations-framework-readme.md to have some content.

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+const longName = 'eks-blueprint-with-an-extra-kind-of-very-long-name-that-is-definitely-not-going-to-work';
+
+const addOns: Array<blueprints.ClusterAddOn> = [
+  new blueprints.addons.MetricsServerAddOn(),
+  new blueprints.addons.ClusterAutoScalerAddOn(),
+  new blueprints.addons.VpcCniAddOn(),
+];
+
+type Executeable = () => void;
```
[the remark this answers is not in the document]
Eli1123: I think that would be wise I am not known for my spelling abilities

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  result.push([() => new blueprints.addons.MetricsServerAddOn({ name: longName }), tooBigString63]);
+  result.push([() => new blueprints.addons.MetricsServerAddOn({ repository: "improper.website" }), notUrlError]);
+
+  result.push([() => manyAutoScalingGroup("Name"), tooBigNumber10]);
+  result.push([() => singleErrorInArray("Name", 345), tooBigNumber100]);
+
+  return result;
+}
+
+function compareObjectFields(object1: any, object2: any) {
```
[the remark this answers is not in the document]
Eli1123: Fair I definitely need more practice on proper naming.

lib/addons/helm-addon/index.ts
Eli1123: Oops thanks for catching that, that was from testing but forgot to change it in helm like I did the others.

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,221 @@
...
+type Executeable = () => void;
+
+type DataError = [Executeable, ZodError];
+
+function getConstraintsDataSet(): DataError[] {
+  let result: [Executeable, ZodError][] = [];
+  const blueprint1 = blueprints.EksBlueprint.builder().addOns(...addOns).id(longName);
+  const blueprint2 = blueprints.EksBlueprint.builder().addOns(...addOns).enableControlPlaneLogTypes(CONTROL_PLANE_LOG_TYPE.authenticator).name("").id("id name");
+
+  const tooBigNumber100 = new z.ZodError([{
```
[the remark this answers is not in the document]
Eli1123: Dang I did not realize I nested all those functions! I believe I have it formatted properly now and it looks much better.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1 @@
+## Here?
```
shapirov103, line 1: Empty file?

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -42,6 +43,85 @@ export interface GenericClusterProviderProps extends eks.ClusterOptions {
...
+     * fargateProfileNames can be no less than 1 character long, and no greater than 63 characters long.
+     * https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+     */
+    fargateProfileName = new StringConstraint(1, 63);
+}
+
+export class GenericClusterPropsConstraints implements ConstraintsType<GenericClusterProviderProps> {
+    /**
+     * managedNodeGroups max size is 10 managed node groups per EKS cluster, and as little as 0.
+     * https://aws.amazon.com/blogs/containers/eks-managed-node-groups/#:~:text=Additionally%2C%20there%20is%20a%20limit,on%20a%20given%20EKS%20cluster
```
shapirov103, line 115: This is dated: 18 NOV 2019
Things changed since then. Only use official AWS documentation for reference. E.g. we can have by default 450 nodes per node group. And even that is a soft limit. I would relax this constraint and use a hard limit. Needs more research.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -42,6 +43,85 @@ export interface GenericClusterProviderProps extends eks.ClusterOptions {
...
+     */
+    managedNodeGroups = new ArrayConstraint(0, 10);
+    /**
+    * autoscalingNodeGroups max size is 10 managed node groups per EKS cluster, and as little as 0.
+    * https://aws.amazon.com/blogs/containers/eks-managed-node-groups/#:~:text=Additionally%2C%20there%20is%20a%20limit,on%20a%20given%20EKS%20cluster
```
shapirov103, line 120: same comment: please use official AWS documentation and only apply hard limits (as opposed to soft limits). We can discuss.

lib/addons/helm-addon/index.ts
```diff
@@ -1,26 +1,66 @@
...
+
+    /**
+    * release can be no less than 1 character long, and no greater than 63 characters long.
+    * https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+    */
+    release = new StringConstraint(1, 63);
+
+    /**
+    * repository can be no less than 1 character long, and no greater than 63 characters long. It also must follow a URL format.
+    * https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
```
shapirov103, line 37: This is wrong reference and max length is not 63. It can be a regular URL to a git repo. I suspect it could be at least 4KB long, however must follow a valid URL format.

lib/addons/helm-addon/index.ts
```diff
@@ -1,26 +1,66 @@
...
+    */
+    repository = new UrlStringConstraint(1, 63);
+
+    /**
+    * version can be no less than 1 character long, and no greater than 63 characters long.
+    * https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
```
shapirov103, line 43: Version is not governed by kubernetes object names. It is specific to helm. Please take a look at what constraints are applied to the version field.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -3,11 +3,12 @@ import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as eks from "aws-cdk-lib/aws-eks";
 import { Construct } from "constructs";
 import { ClusterInfo, ClusterProvider } from "../spi";
-import { setPath, valueFromContext, Writeable } from "../utils";
+import { ArrayConstraint, ConstraintsType, NumberConstraint, setPath, valueFromContext, Writeable } from "../utils";
```
shapirov103, line 6: The rule that we use: if the number of objects in the import is more than 3 (which is the case here: 6) you should use wildcard import e.g. `import * as utils from `

shapirov103 (changes requested):
Looking better on the code structure. Some minor comments on the code side.
However, the documentation and URL references that you used, as well as the constraints themselves are a concern. Please take a look at the comments. 
Also, we cannot use blog references for decisions. Those date fast.

mkdocs.yml
```diff
@@ -3,7 +3,7 @@ repo_name: "aws-quickstart/cdk-eks-blueprints"
 repo_url: "https://github.com/aws-quickstart/cdk-eks-blueprints"
 docs_dir: "docs"
 theme:
-  name: material
+  name: mkdocs
```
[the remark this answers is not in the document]
Eli1123: If I do that I get `ERROR    -  Config value: 'theme'. Error: Unrecognised theme name: 'material'. The available installed themes are: readthedocs, mkdocs` when I run make mkdocs. Is that okay?

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+const notUrlError = new z.ZodError([{
+  validation: "url",
+  code: "invalid_string",
+  message: "",
+  path: []
+}]);
+
+const longName = 'eks-blueprint-with-an-extra-kind-of-very-long-name-that-is-definitely-not-going-to-work';
+const blueprintBase = blueprints.EksBlueprint.builder().addOns(...addOns);
```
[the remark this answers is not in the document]
Eli1123: Is that because longName is not really related to blueprintBase directly?

docs/internal/input-validations-framework-readme.md
Eli1123: When I was talking to you before you left you mentioned a readme and I made this and was like is this name good? Andy you were like yes, but as I went on I realized what is this for? Do we not need it since the make mkdocs does the documentation for us? Or do I need to also make a readme here?

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    if (object1[key] !== object2[key]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+test("For Each loop test.", () => {
```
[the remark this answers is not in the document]
Eli1123: Testing the naming constraints for various fields in the following: EksBlueprint, Fargate, AutoScalingGroup or arrays of them, and HelmAddons. 

Would this name suffice?

lib/addons/helm-addon/index.ts
Eli1123: Would this be the correct reference?https://docs.aws.amazon.com/connect/latest/APIReference/API_UrlReference.html

Seems the name of the URL variable can be: Minimum length of 1. Maximum length of 4096.

While the URL itself can be: Length Constraints: Minimum length of 0. Maximum length of 4096.

lib/cluster-providers/generic-cluster-provider.ts
Eli1123: I believe this link should suffice: https://docs.aws.amazon.com/eks/latest/userguide/service-quotas.html

And I would add a comment along the lines of "managedNodeGroups soft limit is 450 managed node groups per EKS cluster, and as little as 0. But we multiply that by a factor of 5 to 2250 for the situations of hard limit requests made"

Eli1123: While the examples of nodes per managed node group would be: 
"nodes per node group has a soft limit of 450 node, and as little as 0. But we multiply that by a factor of 5 to 2250 in case of situations of a hard limit request being accepted, and as a result the limit would be raised".

docs/internal/input-validations-framework-readme.md
shapirov103: Yes, we need this file. You will provide general description of the framework for internal contributors, how to use it, what types of constraints exist. With respect to the constraints that you defined, you will just include links to the generated mkdocs API docs.

lib/addons/helm-addon/index.ts
shapirov103: This one does not have to come from AWS site, it is a general URL format governed by RFC, e.g. https://datatracker.ietf.org/doc/html/rfc1738

mkdocs.yml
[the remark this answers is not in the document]
Eli1123: Okay I just installed material with pip install mkdocs-material and it works fine now

test/constraints-validation.test.ts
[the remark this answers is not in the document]
Eli1123: Would this suffice as the test text? 
Given that specific validation constraints have been done prior. When creating EKS Blueprints, ClusterProviders, and Helm Add-ons with invalid attributes, the tests will fail with the expected errors.

lib/addons/helm-addon/index.ts
Eli1123: From my research there is no actual limit on it. But different browsers set different character limits. So that amazon link I believe should suffice since its the limits for amazon and being 4096 covers most things since lots of browsers limit to like 2000 something. Would you agree?

test/constraints-validation.test.ts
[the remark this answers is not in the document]
shapirov103: yes (although generally it is a single sentence)

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
+# Description
+
+This is an internal readme to explain how to use the validation constraints framework to detect
+errors at build time rather than run time.
+
+## What can you use with the framework
+
+You will find in the constraints-utils.ts file the generic calls for the following classes if you properly import them from the constraints-utils.ts file.
```
shapirov103, line 8: what do you mean by "generic calls"? Just state something like:


# How to use the framework

The framework implementation is located in the utils/contraints-utils.ts module: 

```typescript
import * from 'utils/constraints-utils`;
```

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+## generic class of validations constraints-utils.ts
```
shapirov103, line 10: Which class are you referring to here? It is file. You can refer to it as "constraints framework" or "constraints module".

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+This class holds the following classes and function(s) to validate constraints.
+
+## StringConstraint
+
+Called with: new StringConstraint(minValue, maxValue);
```
shapirov103, line 16: Replace "called with" with "Constructor:"
applies elsewhere.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+If given number falls outside of these inclusive bounds throws detailed Zod error.
+
+## ArrayConstraint
+
+new utils.ArrayConstraint(minValue, maxValue);
+
+If given array length falls outside of these inclusive bounds throws detailed Zod error.
+
+## validateConstraints Function
```
shapirov103, line 38: Replace with: 
## Framework entry point

`validateConstraints<T>(...)` is the entry point to use the framework. This function can validate either a single object or an array of objects against the provided constraints.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+This is a generic function to take in three parameters:
+
+    constraints: the class type you specify to define key constraints.
+
+    context: context name of whats being tested for detailed error purposes.
+
+    object: The object in question being validated.
+
+## How to use the constraints-utils.ts class?
```
shapirov103, line 50: Please add examples of constraints definition and how to call the entry point with the defined constraints.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+You need two things when utilizing constraints-utils.ts
+
+    -A class with specified keys assigned to given constraints.
+
+    -Calling validateConstraints with right parameters in the right constructor.
```
shapirov103, line 56: Add a Limitations section and mention that at the moment constraints can be defined for flat object. Nested structures will require individual validations.

lib/stacks/eks-blueprint-stack.ts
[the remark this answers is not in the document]
shapirov103: No, this comment was specific to enum values, not enum name itself. Please keep the original name of the enum, and replace values (constants) inside the enum with the CONSTANT_CASE.

lib/addons/helm-addon/kubectl-provider.ts
```diff
@@ -45,7 +45,7 @@ export interface HelmChartConfiguration {
 }
 
 /**
- * Extends helm chart configurtion (repo/chart) with deployment parameters: values, timeout and wait.
+ * Extends helm chart configuration (repo/chart) with deployment parameters: values, timeout and wait.
```
shapirov103, line 48: :+1:  thank you for this

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -3,11 +3,12 @@ import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as eks from "aws-cdk-lib/aws-eks";
 import { Construct } from "constructs";
 import { ClusterInfo, ClusterProvider } from "../spi";
-import { setPath, valueFromContext, Writeable } from "../utils";
+import * as utils from "../utils";
 import * as constants from './constants';
 import { AutoscalingNodeGroup, ManagedNodeGroup } from "./types";
 import assert = require('assert');
-
+import { StringConstraint, validateConstraints } from '../utils';
```
shapirov103, line 10: Why are these here? You already imported wildcard from ../utils on line 6.

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export class StringConstraint implements Constraint {
+
+    constructor(readonly min?: number, readonly max?: number) { }
+
+    validate(key: string, value: any, identifier: string) {
+
+        if (value != undefined)
+            z.string()
+                .min(this.min ?? 0, { message: key + " (" + identifier + ": " + value + ") must be no less than " + this.min + " characters long!" })
```
shapirov103, line 15: Please replace ! with "." - no need to yell :)

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export class ArrayConstraint implements Constraint {
+
+    constructor(readonly min?: number, readonly max?: number) { }
+
+    validate(key: string, value: any, identifier: string) {
+
+        if (value != undefined)
+            z.number()
+                .gte(this.min ?? 1, { message: key + " (" + identifier + " of length: " + value.length + ") must be no less than " + this.min + " node groups!" })
```
shapirov103, line 61: message could be using the back ticks format, e.g. 
\``${key} (${identifier} of length ...`\`

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+                .lte(this.max ?? 3, { message: key + " (" + identifier + " of length: " + value.length + ") must be no more than " + this.max + " node groups!" })
+                .parse(value.length);
+    }
+}
+
+export type ConstraintsType<T> = Partial<Record<keyof T, Constraint>>;
+
+export function validateConstraints<T>(constraints: ConstraintsType<T>, context: string, ...object: any) {
```
shapirov103, line 69: Please add jsdoc to each exported structure: type, function

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    version: KubernetesVersion.V1_21,
+    fargateProfiles: {
+      "fp1": {
+        fargateProfileName: fargateProfileName,
+        selectors: [{ namespace: "serverless1" }]
+      }
+    }
+  });
+}
+function getConstraintsDataSet(): DataError[] {
```
shapirov103, line 126: line break above

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  let result: [Executable, ZodError][] = [];
```
shapirov103, line 128: 4 spaces is the indentation that we use

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    if (object1[key] !== object2[key]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+test("So given that specific validation constraints have been done prior, when creating EKS Blueprints, ClusterProviders, and Helm Add-ons with invalid attributes, the tests will fail with the expected errors.", () => {
```
shapirov103, line 173: Please remove "So": Given validation constraints are defined, when creating

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,190 @@
...
+  getConstraintsDataSet().forEach((ex) => {
+    try {
+      ex[0]();
+      expect(true).toBe(false);
+    } catch (e) {
+      console.log(e);
```
shapirov103, line 179: Do we need this console.log?

test/constraints-validation.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      if (e instanceof z.ZodError) {
+        const thrownError = (e.issues[0] as any);
+        const customError = (ex[1].issues[0] as any);
+        expect(compareIssues(thrownError, customError)).toBe(true);
+      }
+      else {
+        console.log("Not a Zod Error it passed: " + e);
```
shapirov103, line 186: Fail here and remove the log message.

lib/utils/index.ts
```diff
@@ -11,3 +11,4 @@ export * from './yaml-utils';
 export * from './cluster-utils';
 export * from './vpc-utils';
 export * from './registry-utils';
+export * from './constraints-utils';
```
shapirov103, line 14: Ideally in alphabetical order please

docs/internal/readme-internal.md
```diff
@@ -54,4 +54,7 @@ For external contributors:
    1. Clear description of the feature
    2. Test coverage
    3. Validation instructions
-   
\ No newline at end of file
+
+## Validations framework link to readme
+
+Path: docs/internal/input-validations-framework-readme.md
```
shapirov103, line 60: Please add a link here, not just path text. It should be clickable.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+## StringConstraint
+
+Called with: new StringConstraint(minValue, maxValue);
+
+If given string length falls outside of these inclusive bounds throws detailed Zod error
+
+## UrlStringConstraint
+
+Called with new UrlStringConstraint(minValue, maxValue);
```
shapirov103, line 22: Please add a link to the generated API doc for each referenced class.

shapirov103 (changes requested):
Some comments for you. I like the structure and how it is shaping up, a LOT better. 

Also, please remove get-pip.py file - I warned you about it.

lib/cluster-providers/generic-cluster-provider.ts
Eli1123: When I get imports I use the quick fix so I don't manually have to type it, and I think it split it up into two for some reason and I did not notice. I should be sure to review my imports in the future if I do that.

lib/utils/constraints-utils.ts
Eli1123: OH its the ` not ' or " thats why that was not working when I tried it before!

test/constraints-validation.test.ts
Eli1123: I'm sorry what do you mean? Like 4 spaces before let? I just used the Format Document feature and it does it like this. Or do you mean in the space between function and let perhaps?

test/constraints-validation.test.ts
Eli1123: No I forgot to remove that and I noticed it right after I pushed it.

test/constraints-validation.test.ts
shapirov103: Scratch that, it was an aberration in the GitHub view - your tab size looked like it was "2". We adopted tab size "4" and spaces instead of tabs.

test/constraints-validation.test.ts
Eli1123: Is there a certain way you want me to have it fail? Like throw and error? Or use something like fail();

docs/internal/readme-internal.md
Eli1123: I did not know you could do that in a readme

test/constraints-validation.test.ts
shapirov103: Jest does not have a clear fail() method like normal test framework do. Throw an error with description about why it is a failure.

docs/internal/input-validations-framework-readme.md
Eli1123: I meant like the following things as in the rest of the read me, I guess that was not very clear. Would listing it out be okay like StringConstarint, UrlStringConstraint, NumberConstraint, ArrayConstraint, and the validateConstraints function

shapirov103: you can call it something like "supported contraints":

test/constraints-validation.test.ts
shapirov103: Of course, you can commit as many times as you want.  Once you are ready for review, in the PR Reviewers section, click the "Re-request review" button in front of my name.

lib/utils/constraints-utils.ts
Eli1123: /**
 * @type {ConstraintType<T>}
 */
 
 do you mean something like this?
 and for a function like this?
 /**
 * @type {void}
 */
 If it did not have a return type? But the void becomes the return type?

shapirov103: If you type /** and hit enter above the function it will give you the template. 
E.g.

```
/**
 * <description goes here>
 * @param constraints <describe what it is>
 * @param context <describe what it is>
 * @returns 
 */
```

Use the same approach for types. There is no specific template for types, just put the description in the comments.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,75 @@
...
+
+If given array length falls outside of these inclusive bounds throws detailed Zod error.
+
+## validateConstraints Function
+
+validateConstraints<T>(constraints: ConstraintsType<T>, context: string, ...object: any) is the entry point to use the framework. This function can validate either a single object or an array of objects against the provided constraints.
+
+## How to use the constraints-utils.ts
+
+You need two things when utilizing constraints-utils.ts and the following examples are from ['here'](/lib/stacks/eks-blueprint-stack.ts)
```
shapirov103, line 53: This sentence does not read right, it is combining two unrelated things. 

When you say "you need two things" either you tell what they are or below you have a numbered list. E.g. first you need to define constraints object  and second ...

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,88 @@
+import { z } from "zod";
+
```
shapirov103, line 2: jsdocs (/***/) comments around each exported member. Applies to each exported member on this file: type, function, interface.

shapirov103 (commented):
Looking good. Minor stuff left (this time really minor).

docs/internal/input-validations-framework-readme.md
Eli1123: Oops just changed what was below that, changed it to `Validations file constraints-utils.ts`

lib/addons/helm-addon/index.ts
```diff
@@ -1,26 +1,66 @@
 import { Duration } from "aws-cdk-lib";
 import { Construct } from "constructs";
 import * as spi from "../..";
-import { cloneDeep } from "../../utils";
+import { cloneDeep, ConstraintsType, StringConstraint, UrlStringConstraint, validateConstraints } from "../../utils";
```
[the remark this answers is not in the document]
Eli1123: Oh thought you said more than 6

lib/utils/constraints-utils.ts
Eli1123: Oh interfaces also

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,92 @@
...
+
+The constraints framework implementation is located in the utils/constraints-utils.ts module
+
+```typescript
+import * from 'utils/constraints-utils';
+```
+
+## What can you use with the framework
+
+You will find in the constraints framework the generic calls for the following classes if you properly import them from the constraints framework.
```
shapirov103, line 11: There are no generic calls in the framework. A call is an invocation of a specific function by the client (e.g. any code outside of the framework). 

You can state: the constraints framework provides a set of generic classes and interfaces as well as the invocation framework to validate constraints on arbitrary objects.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Validations file constraints-utils.ts
+
+This class holds the supported constraints and function(s) to validate constraints defined below in the rest of this document.
```
shapirov103, line 15: It is not a class.

docs/internal/input-validations-framework-readme.md
```diff
@@ -0,0 +1,92 @@
...
+
+Example with two keys: 
+
+```typescript
+export class BlueprintPropsConstraints implements ConstraintsType<EksBlueprintProps> {
+    id = new StringConstraint(1, 63);
+    name = new StringConstraint(1, 63);
+```
+
+Second you need to call validateConstraints.
```
shapirov103, line 82: validateConstraints is code so should be in ``.

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * This is interface for the constraints needed to test asking for the key name of the object, the value being tested, and context for detailed Zod errors.
+ */
+export interface Constraint {
+    validate(key: string, value: any, identifier: string): any;
+}
+
+export class StringConstraint implements Constraint {
```
shapirov103, line 10: "each exported member" - documented

lib/utils/constraints-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+        if (value != undefined)
+            z.number()
+                .gte(this.min ?? 1, { message: `${key} (${identifier} of length: ${value.length}) must be no less than ${this.min} node groups.` })
+                .lte(this.max ?? 3, { message: `${key} (${identifier} of length: ${value.length}) must be no more than ${this.max} node groups.` })
+                .parse(value.length);
+    }
+}
+
+/**
+ * This is responsible for grabbing all the given keys and excluding any not specified to be tested later.
```
shapirov103, line 71: The type that derives from a generic input structure, retaining the keys. Enables to define mapping between the input structure keys and constraints.

shapirov103 (commented):
Docs comments mostly

docs/internal/input-validations-framework-readme.md
Eli1123: Like this?

Second you need to call 
```typescript 
validateConstraints();
```

Example:

```typescript
validateConstraints(new BlueprintPropsConstraints, EksBlueprintProps.name, blueprintProps);
```

lib/utils/constraints-utils.ts
Eli1123: Ah so also classes, just anything exported then really.

docs/internal/input-validations-framework-readme.md
shapirov103: Second, you need to call `validateConstraints` function: 
(note: punctuation, formatting)

shapirov103 (approved):
LGTM

PR #431, opened 2022-07-05
Correct key-value of elasticSearch

vumdao:
*Issue #, if available:*

*Description of changes:*
Unfortunately the key of region is `awsRegion` instead of `region` like others
Check https://github.com/aws/eks-charts/blob/master/stable/aws-for-fluent-bit/values.yaml#L119


shapirov103 (approved):
@vumdao Thank you for catching it! LGTM
will be published with the next batch of changes (1.1.0)

Issue #419, opened 2022-06-20
EBS CSI Driver - deployed with incorrect trust relationship


daveschmidt86:
### Describe the bug

Update (6/21//2022):

In further troubleshooting, it was discovered that the Addon is deploying an incorrect trust relationship with the OIDC sub as:

`system:serviceaccount:kube-system:aws-ebs-csi-driver`

When I manually update this to match the ebs-csi documentation (https://docs.aws.amazon.com/eks/latest/userguide/csi-iam-role.html) it works, here is the correct sub:

`system:serviceaccount:kube-system:ebs-csi-controller-sa`

EBS CSI Addon Version: 1.6.2
K8 Cluster version: 1.21

When attempting to create a new volume, PVC is reporting the following auth error:

"failed to provision volume with StorageClass "ebs-sc": rpc error: code = Internal desc = Could not create volume "pvc-93e1b143-5f66-4ddd-85fb-6bd293e4d866": could not create volume in EC2: WebIdentityErr: failed to retrieve credentials caused by: AccessDenied: Not authorized to perform sts:AssumeRoleWithWebIdentity status code: 403, request id:"

using example Dynamic Provisioning manifests from: https://docs.aws.amazon.com/eks/latest/userguide/ebs-sample-app.html

### Expected Behavior

PVC and volume should have created 

### Current Behavior

Auth Error assuming Web Identity

### Reproduction Steps

Stood up new cluster, specified k8 version 1.21, ebs csi addon version 1.6.2.

Cluster and default gp2 strorage class verified and created by default.

Deploy example Dynamic Provisioning manifests from this doc: https://docs.aws.amazon.com/eks/latest/userguide/ebs-sample-app.html

### Possible Solution

Modify the sub in the OIDC trust relationship for the ebs-csi service account from:

`system:serviceaccount:kube-system:aws-ebs-csi-driver`

to this:

`system:serviceaccount:kube-system:ebs-csi-controller-sa`

### CDK CLI Version

2.20

### EKS Blueprints Version

1.0.1

### Node.js Version

16

### Environment details (OS name and version, etc.)

MACOS

elamaran11:
This issue is fixed now with this merge, Please validate and close the issue.

rstrazza:
When will `1.0.5` be released? Assuming this fix would be part of it. I'm experiencing the same error.

Btw, thanks for the workaround, [daveschmidt86](https://github.com/daveschmidt86)!

[...]

PR #433, opened 2022-07-07
Fixing EBS CSI Driver Addon and coreAddon Modules

elamaran11:
*Issue #, if available:* #419 

*Description of changes: 

1. Standardization of the `core-addon` module to allow service account name as additional parameter and using the same to create service account for the addon using this model. 
2. Added a dependency to `CfnAddon` to avoid race condition while deployment of the addons.
3. Fixing Issue Related to EBS CSI Driver - deployed with incorrect trust relationship in `ebs-csi-driver` Addon.
4. Adding SA Name to pass as parameter to `CoreAddon` on `coredns`, `kube-proxy` and `vpc-cni` Addons.
 Also 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice

### DIFF
--- a/lib/addons/coredns/index.ts
+++ b/lib/addons/coredns/index.ts
@@ -8,7 +8,8 @@ export class CoreDnsAddOn extends CoreAddOn {
     constructor(version?: string) {
         super({
             addOnName: "coredns",
-            version: version ?? "v1.8.4-eksbuild.1"
+            version: version ?? "v1.8.4-eksbuild.1",
+            saName: "coredns"
         });
     }
 }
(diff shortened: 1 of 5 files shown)


lib/addons/coredns/index.ts
```diff
@@ -8,7 +8,8 @@ export class CoreDnsAddOn extends CoreAddOn {
     constructor(version?: string) {
         super({
             addOnName: "coredns",
-            version: version ?? "v1.8.4-eksbuild.1"
+            version: version ?? "v1.8.4-eksbuild.1",
+            saName: "coredns"
```
shapirov103, line 12: not strictly necessary for anything other than EBS CSI add-on, since none of the rest pass the policies for the IRSA account.

shapirov103 (approved):
LGTM, as long as e2e test passes

PR #457, opened 2022-07-30
AWSPrivateCAIssuerAddon Added

codemato:
*Issue #, if available:*

*Description of changes:*
This addon will install [aws-privateca-issuer](https://github.com/cert-manager/aws-privateca-issuer/)

AWS ACM Private CA is a module of the AWS Certificate Manager that can setup and manage private CAs.
The AWS PrivateCA Issuer plugin acts as an addon to cert-manager that signs certificate requests using ACM Private CA.

Since its an addon to cert-manager, for Installing AWS ACM Private CA Addon, You must install [cert-manager](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/addons/cert-manager.md) Addon first


lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,95 @@
+// lib/kubevious_addon.ts
+import { Construct } from 'constructs';
+import { ManagedPolicy } from "aws-cdk-lib/aws-iam";
+import merge from "ts-deepmerge";
+import { HelmAddOn, HelmAddOnUserProps, HelmAddOnProps } from "../helm-addon";
+import { dependable } from '../../utils';
```
shapirov103, line 6: Let's group all things that you import from utils in a single import. 
`import { dependable, setPath etc. } from "../../utils";`

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    //Create Service Account with IRSA
+    const cluster = clusterInfo.cluster;
+    let values: Values = populateValues(this.options);
+    values = merge(values, this.props.values ?? {});
+    
+    if (this.options.createServiceAccount === false ) {
+      //Create Service Account with IRSA
+      const opts = { name: this.options.serviceAccountName, namespace: this.options.namespace };
+      const sa = cluster.addServiceAccount(this.options.serviceAccountName!, opts);
+      const acmPCAPolicy = ManagedPolicy.fromAwsManagedPolicyName("AWSCertificateManagerPrivateCAFullAccess");
```
shapirov103, line 65: Is full access the recommended level of permissions for this add-on based on least privilege principle?

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ClusterInfo, Values } from "../../spi";
+import { createNamespace } from "../../utils";
+/**
+ * User provided options for the Helm Chart
+ */
+export interface AWSPrivateCAIssuerAddonProps extends HelmAddOnUserProps {
+    /**
+     * Specifies whether a service account should be created
+     */
+    createServiceAccount?: boolean;
```
shapirov103, line 17: it sounds like it is a similar issue as in cert-manager. We should not approach it as a helm chart, this is an add-on. So createServiceAccount - true should indicate whether CDK should create a service account and false is to delegate this feature to the helm chart. 
Let's make it consistent with the cert-manager and other add-ons.

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      const namespace = createNamespace(this.options.namespace! , cluster);
+      sa.node.addDependency(namespace);
+      
+      const chart = this.addHelmChart(clusterInfo, values);
+      chart.node.addDependency(sa);
+      return Promise.resolve(chart);
+
+    } else {
+      //Let aws-pca-issuer Create Service account for you. This is controlled by flag helmOptions.createServiceAccount 
+      const chart = this.addHelmChart(clusterInfo, values);
```
shapirov103, line 77: Let's not duplicate the same logic to add helm chart for `if` and `else` block just because we need to add a dependency on sa. 
Let's create a function scope variable for service account. E.g.

```
deploy() {
    let sa: ServiceAccount | undefined = null;
   
    if(createServiceAccount) {
        sa = createServiceAccount();
        ...
    }

   const chart = this.addHelmChart(...);
   if(sa) {
       <add dependency>
   }

}
```

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,95 @@
...
+import merge from "ts-deepmerge";
+import { HelmAddOn, HelmAddOnUserProps, HelmAddOnProps } from "../helm-addon";
+import { dependable } from '../../utils';
+import { setPath } from '../../utils/object-utils';
+import { ClusterInfo, Values } from "../../spi";
+import { createNamespace } from "../../utils";
+/**
+ * User provided options for the Helm Chart
+ */
+export interface AWSPrivateCAIssuerAddonProps extends HelmAddOnUserProps {
```
shapirov103, line 13: Are there any configuration options from the helm chart that would make sense to promote to the root level?

shapirov103 (commented):
Looks great, a few comments.

lib/addons/aws-privateca-issuer/index.ts
codemato: done

lib/addons/aws-privateca-issuer/index.ts
codemato: yes.. makes sense.. make it more unopinionated, I have made it as a user input same like keda

lib/addons/aws-privateca-issuer/index.ts
codemato: yes.. agree.. made it same like keda, If user provides the Roles, then cdk will create service account

lib/addons/aws-privateca-issuer/index.ts
codemato: updated the code

lib/addons/aws-privateca-issuer/index.ts
codemato: These are the available values
https://github.com/cert-manager/aws-privateca-issuer/blob/main/charts/aws-pca-issuer/values.yaml
I couldnt find anything which required to be promoted to root level

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * The name of the service account to use. If createServiceAccount is true, a serviceAccountName is generated.
+     */
+    serviceAccountName?: string;
+    /**
+     * An array of Managed IAM Policies which Service Account needs for IRSA Eg: irsaRoles:["AWSCertificateManagerPrivateCAFullAccess"]. If not empty
+     * Service Account will be Created by CDK with IAM Roles Mapped (IRSA). In case if its empty, Service Account will be created with out IAM Roles
+     */   
+     irsaRoles?: string[];
```
shapirov103, line 23: Let's call them what they are: iamPolicies. The fact they are attached to the role for IRSA is our implementation details. We only create a single role behind the scenes.

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,103 @@
...
+const defaultProps: HelmAddOnProps & AWSPrivateCAIssuerAddonProps = {
+  name: "blueprints-aws-pca-issuer-addon",
+  chart: "aws-privateca-issuer",
+  namespace:"aws-pca-issuer",
+  version: "1.2.2",
+  release: "aws-pca-issuer",
+  repository:  "https://cert-manager.github.io/aws-privateca-issuer",
+  values: {},
+  serviceAccountName: "aws-pca-issuer",
+  irsaRoles: []
```
shapirov103, line 38: I think it would be beneficial to add the AWSCertificateManagerPrivateCAFullAccess by default and allow customers to override. If they pass policies, we use those. If they don't we apply the default which can be used to instantiate the add-on with no parameters.

lib/addons/aws-privateca-issuer/index.ts
```diff
@@ -0,0 +1,103 @@
...
+ * @param sa - Service Account Object
+ * @param irsaRoles - Array  of Managed IAM Policies
+ */
+ function setRoles(sa:ServiceAccount, irsaRoles: string[]){
+  irsaRoles.forEach((policyName) => {
+      const policy = ManagedPolicy.fromAwsManagedPolicyName(policyName);
+      sa.role.addManagedPolicy(policy);
+    });
+}
+
```
shapirov103, line 101: minor: removing trailing line breaks.

shapirov103 (commented):
Looks awesome! A few tactical/API shape comments. 
Also do you mind adding it to the example blueprint so that I can kick off e2e test for verification.
@codemato

lib/addons/aws-privateca-issuer/index.ts
codemato: done

lib/addons/aws-privateca-issuer/index.ts
codemato: done

shapirov103 (approved):
LGTM
will run e2e then merge

PR #461, opened 2022-08-08
Feature/amp addon tech debt

elamaran11:
*Issue #, if available:* AMP Addon Tech Debts and Improvements.

*Description of changes:*
This PR is to improve Tech Debts on AMP Addon to make it consistent with XRAY and CW Addon on minor tech debts.

### DIFF
--- a/docs/addons/amp-addon.md
+++ b/docs/addons/amp-addon.md
@@ -14,7 +14,7 @@ For more information on the add-on, please review the [user guide](https://docs.
 
 This add-on can used with four different patterns :
 
-Pattern # 1 : Simple and Easy - Using all default property values. This pattern creates a new AMP workspace with default property values such as `workspaceName`, `tags`, `namespace` and deploys an ADOT collector in the `default` namespace with `deployment` as the mode to remote write metrics to AMP workspace.
+Pattern # 1 : Simple and Easy - Using all default property values. This pattern creates a new AMP workspace with default property values such as `workspaceName`, `namespace` with no tags on the AMP workspace and deploys an ADOT collector in the `default` namespace with `deployment` as the mode to remote write metrics to AMP workspace. 
 
 ```typescript
 import * as cdk from 'aws-cdk-lib';
--- a/docs/addons/amp-addon.md
+++ b/docs/addons/amp-addon.md
@@ -1,18 +1,18 @@
 # Amazon Managed Service for Prometheus (AMP)) Amazon EKS Add-on
 
-Amazon Managed Service for Prometheus is a Prometheus-compatible monitoring and alerting service that makes it easy to monitor containerized applications and infrastructure at scale. This Addon deploys an AWS Distro for OpenTelemetry (ADOT) Collector for Amazon Managed Service for Prometheus (AMP) which receives OTLP metrics from the application and Prometheus metrics scraped from pods on the cluster and remote writes the metrics to AMP remote write endpoint. You can change the mode to Daemonset, StatefulSet, and Sidecar depending on your deployment strategy.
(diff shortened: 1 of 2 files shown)


docs/addons/amp-addon.md
```diff
@@ -54,14 +54,15 @@ const addOn = new blueprints.addons.AmpAddOn({
...
 
 const blueprint = blueprints.EksBlueprint.builder()
   .addOns(addOn)
   .build(app, 'my-stack-name');
 ```
-Pattern # 3 : Passing on AMP Remote Write Endpoint URL of an existing AMP workspace to be used to remote write metrics. This pattern does not create an AMP workspace. Deploys an ADOT collector on the namespace specified in `namespace` with `deployment` as the mode to remote write metrics to AMP workspace of the URL passed as input. This pattern ignores any other property values passed if `prometheusRemoteWriteURL` is present.
+Pattern # 3 : Passing on AMP Remote Write Endpoint URL of an existing AMP workspace to be used to remote write metrics. This pattern does not create an AMP workspace. Deploys an ADOT collector on the namespace specified in `namespace` with name in `name` and `deployment` as the mode to remote write metrics to AMP workspace of the URL passed as input. This pattern ignores any other property values passed if `prometheusRemoteWriteURL` is present.
 
 ```typescript
 import 'source-map-support/register';
```
shapirov103, line 68: at some point we will need to clean up all of the source-map-support in the examples. I am unclear why it started in the first place.

shapirov103 (commented):
This will need to be rebased after CW is merged. In general, please branch off upstream main rather than your main branch, that will reduce the noise in the PR and need to rebase.

docs/addons/amp-addon.md
elamaran11: Sounds good.

elamaran11:
> This will need to be rebased after CW is merged. In general, please branch off upstream main rather than your main branch, that will reduce the noise in the PR and need to rebase.
Sounds good, Will follow this.

lib/addons/amp/index.ts
```diff
@@ -116,23 +113,11 @@ export class AmpAddOn implements ClusterAddOn {
             }
             cfnWorkspace = new aps.CfnWorkspace(cluster.stack, 'MyAMPWorkspace', cfnWorkspaceProps);/* all optional props */ 
             const URLEndpoint = cfnWorkspace.attrPrometheusEndpoint;
```
shapirov103, line 115: Since we are addressing tech debt:

all variables are lower camel case. regardless of the acronyms. const is just a shortcut for readonly variables.

lib/addons/amp/index.ts
```diff
@@ -116,23 +113,11 @@ export class AmpAddOn implements ClusterAddOn {
             }
             cfnWorkspace = new aps.CfnWorkspace(cluster.stack, 'MyAMPWorkspace', cfnWorkspaceProps);/* all optional props */
```
shapirov103, line 114: Let's come up with a better ID than the hello world example of 'MyAMPWorkspace'. I am thinking something like workspaceName + "-amp-workspace".

lib/addons/amp/index.ts
```diff
@@ -97,10 +97,7 @@ export class AmpAddOn implements ClusterAddOn {
             tags: defaultProps.workspaceTag
         };
 
-        if (this.ampAddOnProps.prometheusRemoteWriteURL) {
-            remoteWriteEndpoint = this.ampAddOnProps.prometheusRemoteWriteURL;
-        }
-        else {
+        if (typeof(this.ampAddOnProps.prometheusRemoteWriteURL) == 'undefined') {
```
shapirov103, line 100: The previous logic checks both for undefined and null. This only checks for undefined. I think we need to handle both. Difference:

Null is passed like this:
```
{
    prometheusRemoteWriteURL: null
}
```

Undefined is passed by not specifying property at all.

shapirov103 (commented):
Looks a lot better! A few tactical comments. The biggest is handling tags and ID.

lib/addons/amp/index.ts
elamaran11: Fixed.

lib/addons/amp/index.ts
elamaran11: Fixed.

shapirov103 (approved):
LGTM

PR #488, opened 2022-08-24
JupyterHub Addon Support

youngjeong46:
*Issue #, if available:*

*Description of changes:*

Adds support for JupyterHub as an add-on.


docs/addons/jupyterhub.md
```diff
@@ -0,0 +1,66 @@
+# JupyterHub Add-on
+
+Karpenter add-on is based on the [JupyterHub](https://github.com/jupyterhub/jupyterhub) project that supports a multi-user Hub to spawn, manage, and proxy multiple instances of single user Jupyter notebook server.
```
shapirov103, line 3: Karpenter?

docs/addons/jupyterhub.md
```diff
@@ -0,0 +1,66 @@
...
+hub            ClusterIP      172.20.171.28   <none>                                                                    8081/TCP       26m
+proxy-api      ClusterIP      172.20.31.32    <none>                                                                    8001/TCP       26m
+proxy-public   LoadBalancer   172.20.14.210   xxxxxxxx-1234567890.us-west-2.elb.amazonaws.com   80:32733/TCP   26m
+```
+
+You can log into the JupyterHub portal by accessing the Load Balancer endpoint in any browser. 
+
+![JupyterHub Login](./../assets/images/jupyterhub-login-page.png)
+
+A default `admin` username with `admin` as password is created initially. Once logged in, you should be able to access the main portal page.
```
shapirov103, line 64: Do we provide a way to supply a password (e.g. via secrets store add-on) so that this could be leveraged for production use cases?

lib/addons/jupyterhub/index.ts
```diff
@@ -0,0 +1,52 @@
...
+        super({...defaultProps, ...props});
+        this.options = this.props;
+    }
+
+    deploy(clusterInfo: ClusterInfo): Promise<Construct> {
+        const cluster = clusterInfo.cluster;
+        let values = this.options.values ?? {};
+
+        // Create Namespace
+        const ns = createNamespace(JUPYTERHUB, cluster, true, true);
```
shapirov103, line 43: Is deploying to an existing namespace not an option?

lib/addons/jupyterhub/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    version: '1.2.0',
+    chart: JUPYTERHUB,
+    release: RELEASE,
+    repository: 'https://jupyterhub.github.io/helm-chart/',
+};
+
+/**
+ * Implementation of the JupyterHub add-on
+ */
+export class JupyterHubAddOn extends HelmAddOn {
```
shapirov103, line 29: Based on my understanding how the notebooks work it is seems to be a stateful app that relies on storage to store notebooks. Don't we need a dependency on the csi storage driver e.g. for EBS?

lib/addons/jupyterhub/index.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+ type JupyterHubAddOnProps = HelmAddOnUserProps;
+
+const JUPYTERHUB = 'jupyterhub';
+const RELEASE = 'blueprints-addon-jupyterhub';
+
+/**
+ * Defaults options for the add-on
+ */
+const defaultProps: HelmAddOnProps = {
```
shapirov103, line 17: It would make sense to allow specifying storage class based on [this](https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-storage.html#type-of-storage-provisioned).
Also how much [size](https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-storage.html#size-of-storage-provisioned) is allocated.

shapirov103, line 17: Is there anything we can offer for user management other than admin/admin e.g. based on [this](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html).

shapirov103 (commented):
@youngjeong46 looks like a great start! 
Is there any we can add to the props from the schema for the helm chart: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/schema.yaml? 

We could convert it to the typescript object, e.g. like it is done in [this example](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/container-insights/values.ts) if that is the direction you want to take. I have this generator locally at the moment, we can give it a shot.

docs/addons/jupyterhub.md
youngjeong46: That is probably better for a patterns - I talked to Jake, and the best course of action is to keep this as simple, and specify the custom configurations as part of an overall pattern. This applies to all other configurations as you noted below. Thoughts?

lib/addons/jupyterhub/index.ts
youngjeong46: There is a cognito integration that maps as a pattern - was going to add for patterns as commented above.

youngjeong46: Addressed - this makes use of EBS volume with the CSI addon

lib/addons/jupyterhub/index.ts
youngjeong46: The storage can be either EBS or EFS, which makes this dependency a bit more difficult to enforce.

docs/addons/jupyterhub.md
youngjeong46: Addressed

lib/addons/jupyterhub/index.ts
youngjeong46: Once I can add attributes for EFS as data store, I will implement the EBS or EFS CSI addon as a dependency for jupyterhub

docs/addons/jupyterhub.md
shapirov103: As long as there is a way to use this pattern with no manual intervention afterwards we should good. The docs can be updated to show how to pass the password, ideally aligning with our security best practices.

lib/addons/jupyterhub/index.ts
shapirov103: Not addressed? Why namespace is hardcoded?

lib/addons/jupyterhub/index.ts
shapirov103: That works

shapirov103 (commented):
Close to merge:

1. Let's address the admin pwd with a doc example
2. Let's create an issue for the patterns to create an example with cognito integration,

lib/addons/jupyterhub/index.ts
youngjeong46: addressed.

lib/addons/jupyterhub/index.ts
youngjeong46: Resolved as EBS added - EFS will be added as a followup in a future PR.

docs/addons/jupyterhub.md
youngjeong46: So it sounds like we can supply a password as shown [here](https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/authentication.html#allowed-users-admin-users), but it's not meant for any production use. 

In the end we probably need to introduce other ways of authentication (i.e. OAuth or Cognito as stated below). I can  either:

1/ include this as a way to provide the password (using secrets) and alert that this is not good for production use case OR
2/ Provide OAuth or other authentication method

lib/addons/jupyterhub/index.ts
youngjeong46: addressed.

docs/addons/jupyterhub.md
shapirov103: oauth works, I assume it should be good for cognito or any other oidc/oauth2 provider.

youngjeong46: done. Left a doc link with examples of various IdP set up, as the code uses a generic oauth2 authenticator class within JupyterHub. Will create an issue for pattern with Cognito. It may require some code change, which will be a fast follow if needed.

shapirov103 (commented):
@youngjeong46 ready to merge (after e2e).
Do you want to add a reference to the ./mkdocs.yml ? otherwise it won't show up in the navigation pane.

youngjeong46:
> @youngjeong46 ready to merge (after e2e). Do you want to add a reference to the ./mkdocs.yml ? otherwise it won't show up in the navigation pane.

will address this.

examples/blueprint-construct/index.ts
```diff
@@ -100,7 +100,14 @@ export default class BlueprintConstruct {
...
                 irsaRoles: ["CloudWatchFullAccess", "AmazonSQSFullAccess"]
             }),
-            new blueprints.addons.AWSPrivateCAIssuerAddon()
+            new blueprints.addons.AWSPrivateCAIssuerAddon(),
+            new blueprints.addons.JupyterHubAddOn({
+                ebsConfig: {
+                    storageClass: "gp2",
+                    capacity: "4Gi",
+                },
+                enableIngress: true,
```
shapirov103, line 109: @youngjeong46 if delete fails consistently on ENI removal (or load balancer removal), I suspect it is because it happens asynchronously, so from the cluster perspective the resource is dropped, but when VPC is attempted to be deleted, it is still cleaning it up.
I suggest we set enableIngress to false here to get past the issue.

shapirov103:
> end to end tests failed. A maintainer can provide more details.

Failed on the istio bug, that just got merged. Rerunning.

shapirov103 (approved):
LGTM

Issue #487, opened 2022-08-19
 (feat: addons Karpenter) requirements incorrectly Modelled and/or named.


javydekoning:
### Describe the feature

In Karpenter yaml:

```
  requirements:
    - key: "node.kubernetes.io/instance-type"
      operator: In
      values: ["m5.large", "m5.2xlarge"]
    - key: "topology.kubernetes.io/zone"
      operator: In
      values: ["us-west-2a", "us-west-2b"]
    - key: "kubernetes.io/arch"
      operator: In
      values: ["arm64", "amd64"]
    - key: "karpenter.sh/capacity-type" # If not included, the webhook for the AWS cloud provider will default to on-demand
      operator: In
      values: ["spot", "on-demand"]
```

In EKS Blueprints:

```ts
    provisionerSpecs?: {
        'node.kubernetes.io/instance-type'?: string[],
        'topology.kubernetes.io/zone'?: string[],
        'kubernetes.io/arch'?: string[],
        'karpenter.sh/capacity-type'?: string[],
    }
```

We use the correct naming for `taints`, therefore I think we should rename `provisionerSpecs` to `requirements`. Also, we should remodel this and not hardcode to the `in` operator here: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/f7fba04f2488556366779e56ebb88d860747aa5a/lib/addons/karpenter/index.ts#L180 

### Use Case

Customers might want to use other `operators` or `keys`. 

### Proposed Solution

- rename (this will be a breaking change!)
- remodel so users can use key, operator and values.

Below are some options (pseudo code):

```ts
requirements?: [{
    key: string, 
    operator: "In" | "NotIn" ,
    values: string[],
}]
```

The key itself could be an enum like:

```ts
enum Karpenter.Requirement {
  InstanceType = "node.kubernetes.io/instance-type",
  Zone = "topology.kubernetes.io/zone",
  Arch = "kubernetes.io/arch",
  CapacityType = "karpenter.sh/capacity-type",
}
```

We could also model the requirements into a It's own class. That way, we can use TS to do type-checking and leverage:

- `@aws-cdk/aws-ec2 » InstanceType`
- Default zones to `cdk.Stack.of(this).availabilityZones`
- `@aws-cdk/aws-ec2 » InstanceArchitecture`
- CapacityType

Something like:

```
export interface RequirementOpts {
  operator: "In" | "NotIn" ,
}

[...]

Issue #486, opened 2022-08-19
(feat: addons Karpenter) support consolidation in Karpenter 0.15


javydekoning:
### Describe the feature

Karpenter 0.15 introduces a new configuration. 

```yaml
spec:
  # Enables consolidation which attempts to reduce cluster cost by both removing un-needed nodes and down-sizing those
  # that can't be removed.  Mutually exclusive with the ttlSecondsAfterEmpty parameter.
  consolidation:
    enabled: true
```

This needs to be added to https://github.com/aws-quickstart/cdk-eks-blueprints/blob/f7fba04f2488556366779e56ebb88d860747aa5a/lib/addons/karpenter/index.ts

I suggest we default to `True` if not provided. 

### Use Case

consolidation which attempts to reduce cluster cost by both removing un-needed nodes and down-sizing those that can't be removed.  Mutually exclusive with the ttlSecondsAfterEmpty parameter.

### Proposed Solution

Something like: 

```ts
consolidation?: {
    'enabled': boolean,
}

const consolidation = this.options.consolidation || { consolidation: { enabled: true }};
```

### Acknowledgements

- [x] I may be able to implement this feature request

### CDK version used

N/A

### EKS Blueprints Version

1.2.0

### Node.js Version

N/A

### Environment details (OS name and version, etc.)

N/A

youngjeong46:
@javydekoning thanks for the submission. We haven't up-ed our support of Karpenter to 0.15, so it will be in a future PR with some of the other ongoing Karpenter issues (your others included.)

youngjeong46:
I have a PR in flight that takes care of this issue (and upgrade to 0.16 support). Your other issue, as I mentioned in the ticket, will have to be a followup due to breaking changes.

youngjeong46:
Addressed (and upgraded to 0.16 support) Per PR #489. Closing the issue.

Issue #477, opened 2022-08-15
KarpenterAddOn: Can't add taints property


go4real:
### Describe the bug

Can't add taints property due to a type error.

Error log message is:
Property '0' is missing in type '{ key: string; value: string; effect: string; }[]' but required in type '[{ key: string; value: string; effect: string; }]'.

### Expected Behavior

Taints property applied without error.

### Current Behavior

Can't add taint property due to the type error

### Reproduction Steps

    
    const karpenterAddonProps = {
      taints: [{
        key: "workload",
        value: "service",
        effect: "NoSchedule",  
      }],
      provisionerSpecs: {
        'node.kubernetes.io/instance-type': ['m5.large', 'm5a.large', 'm5.2xlarge'],
        'topology.kubernetes.io/zone': ['us-west-2a', 'us-west-2b'],
        'kubernetes.io/arch': ['amd64'],
        'karpenter.sh/capacity-type': ['spot','on-demand'],
      },
      subnetTags: {
        "aws-cdk:subnet-type": "Private"
      },
      securityGroupTags: {
        "kubernetes.io/cluster/dev-blueprint": "owned",
      },
    }
    
    const vpcCniAddOn = new blueprints.VpcCniAddOn("v1.10.1-eksbuild.1");
    const defaultKarpenterAddOn = new blueprints.KarpenterAddOn(karpenterAddonProps);
    const addOns: Array<blueprints.ClusterAddOn> = [ vpcCniAddOn, defaultKarpenterAddOn ];

### CDK CLI Version

2.37.1 (build f15dee0)

### EKS Blueprints Version

1.2.0

### Node.js Version

v16.16.0

### Environment details (OS name and version, etc.)

Cloud9 - AL2

youngjeong46:
@go4real we have a PR in flight that will resolve the issue. I will notify when it's released.

youngjeong46:
Addressed per PR #489.

PR #489, opened 2022-08-25
Karpenter 0.16 upgrade, consolidation support and hotfixes

youngjeong46:
*Issue #, if available:*
#477, #486, #487

*Description of changes:*
- Karpenter support up to 0.16
- consolidation support (for 0.15 and above, also mutually exclusive with ttlSecondsAfterEmpty parameter)
- weight support (for 0.16 and above)
- Type miscast fix on taints
- Provisioner Requirement property was remodeled match provisioner field name and to allow multiple operators. Since this is a breaking change, upgrade path has been documented in the documentation.


docs/addons/karpenter.md
```diff
@@ -10,17 +10,15 @@ Karpenter works by:
...
 1. This add-on depends on [VPC CNI](vpc-cni.md) Add-on for cni support.
 
 ***VPC CNI add-on must be present in add-on array*** and ***must be in add-on array before the Karpenter add-on*** for it to work, as shown in below example. Otherwise will run into error `Assertion failed: Missing a dependency for VpcCniAddOn`.
 
 2. There is no support for utilizing both Cluster Autoscaler **and** Karpenter. Therefore, any addons list that has both will result in an error `Deploying <name of your stack> failed due to conflicting add-on: ClusterAutoscalerAddOn.`.
 
-## Prerequisite
-
-(If using Spot), EC2 Spot Service Linked Role should be created. See [here](https://docs.aws.amazon.com/batch/latest/userguide/spot_fleet_IAM_role.html) for more details.
+3. (If using Spot), EC2 Spot Service Linked Role should be created. See [here](https://docs.aws.amazon.com/batch/latest/userguide/spot_fleet_IAM_role.html) for more details.
```
shapirov103, line 21: Do you think customers will benefit from doing this automatically with blueprints? I understand the lifecycle for this is account, rather than a cluster, interested in your thoughts on it. If a pattern in the patterns repo with spot and karpenter is useful, let's create an issue for it so that one of us could address.

docs/addons/karpenter.md
```diff
@@ -32,12 +30,12 @@ import * as blueprints from '@aws-quickstart/eks-blueprints';
 const app = new cdk.App();
 
 const karpenterAddonProps = {
-  provisionerSpecs: {
-    'node.kubernetes.io/instance-type': ['m5.2xlarge'],
-    'topology.kubernetes.io/zone': ['us-east-1c'],
-    'kubernetes.io/arch': ['amd64','arm64'],
-    'karpenter.sh/capacity-type': ['spot','on-demand'],
-  },
+  requirements: [
```
shapirov103, line 33: Is it a breaking change, e.g if customer already using Karpenter previous version, will this break it?

lib/addons/karpenter/index.ts
```diff
@@ -85,11 +108,27 @@ export class KarpenterAddOn extends HelmAddOn {
...
         const subnetTags = this.options.subnetTags || {};
         const sgTags = this.options.securityGroupTags || {};
         const taints = this.options.taints || [];
         const amiFamily = this.options.amiFamily;
+        const version = this.options.version!;
+        const ttlSecondsAfterEmpty = this.options.ttlSecondsAfterEmpty || null;
+        const weight = this.options.weight;
+        
+        // Weight only available with v0.16.0 and later
+        if (semver.lt(version, '0.16.0')){
```
shapirov103, line 121: Nice to add this check!

shapirov103 (commented):
@youngjeong46 great work! A couple of comments, to understand the repercussions.

docs/addons/karpenter.md
youngjeong46: I can't put a number around it - but I saw a conversation where around half of the the spot customers on EKS consider Karpenter, so generally it's a pattern we should consider. I'm happy to create an issue on the patterns repo and include this check - this might be revised as a troubleshooting note on this repo (i.e. if you select spot, you may run into errors if you don't do this)

docs/addons/karpenter.md
youngjeong46: If they are using the previous addon version, this will break it. I can add notes to ensure the changes and an upgrade path, although I'm not sure it is sustainable to keep documenting every breaking change.

shapirov103: Upgrade path could be 1/ remove add-on and deploy 2/ add the add-on back and deploy. However, I am unsure what operation impact is going to be when karpenter is removed: I hope all nodes will be running as is.

allamand: If the addon remove the provisioner karpenter will remove the nodes

shapirov103: @youngjeong46 based on the above, we need an upgrade path.

youngjeong46: would it suffice to document the changes in the "requirement" prop? The values would remain the same, it's just the prop type that changed. 

As for version upgrades specifics (like 0.14 -> 0.15 or 0.15 -> 0.16) is documented in the official Karpenter site which is listed under troubleshooting at the bottom.

docs/addons/karpenter.md
youngjeong46: I created the issue on the patterns repo.

docs/addons/karpenter.md
shapirov103: Yes, and let's document and mention in the PR description so that it gets to the release notes, e.g., in the description of changes adding a bullet about upgrade path documentation

shapirov103 (approved):
LGTM

PR #491, opened 2022-08-26
Fixing CW Addon and AMP Addon to emit Custom Metrics.

elamaran11:
@shapirov103  As per our discussion, Submitting this PR to allow all CW metrics to flow to CloudWatch Console.


lib/addons/cloudwatch-adot-addon/collector-config-cloudwatch.ytpl
```diff
@@ -338,13 +338,7 @@ spec:
           enabled: true
         dimension_rollup_option: NoDimensionRollup
         parse_json_encoded_attr_values: [Sources, kubernetes]
-        metric_declarations:
```
shapirov103, line 341: Agree that hard-coding to just some dimensions does not make sense. However, what will be the path now to limit it? I see this applicable to every single production use case. 

I suggest make it as a configurable set of dimensions to the users which are part of the properties.

lib/addons/cloudwatch-adot-addon/index.ts
```diff
@@ -12,7 +12,7 @@ import { KubectlProvider, ManifestDeployment } from "../helm-addon/kubectl-provi
  */
 
 /**
- * Configuration options for add-on.
+ * Configuration options for CloudWatch Adot add-on.
  */
 export interface CloudWatchAdotAddOnProps {
```
shapirov103, line 17: Let's add dimensions here.

shapirov103 (commented):
@elamaran11 definitely a step in the right direction. 
Let's address production use cases, which is a must for the platform view of this. Since we already have replaceable parameters handling, it should not be too much effort, please let me know if you would like to discuss.

lib/addons/cloudwatch-adot-addon/collector-config-cloudwatch.ytpl
elamaran11: Fixed for custom metrics.

lib/addons/cloudwatch-adot-addon/index.ts
elamaran11: Done.

lib/addons/cloudwatch-adot-addon/index.ts
```diff
@@ -42,10 +47,11 @@ export const enum cloudWatchDeploymentMode {
 /**
  * Defaults options for the add-on
  */
-const defaultProps = {
+const defaultProps: CloudWatchAdotAddOnProps = {
     deploymentMode: cloudWatchDeploymentMode.DEPLOYMENT,
     namespace: 'default',
-    name: 'adot-collector-cloudwatch'
+    name: 'adot-collector-cloudwatch',
+    metricsNameSelectors: ['apiserver_request_.*', 'container_memory_.*', 'container_threads', 'otelcol_process_.*']
```
shapirov103, line 54: default selection seems to make sense, but I don't see anything other otelcol_process* metrics in cloudwatch. Expected to see pod and container memory for each container in all namespaces.

shapirov103 (commented):
Getting there. I would to address to use cases with this add-on:

1. Customer chooses to send all metrics to CloudWatch. At the moment I see a single namespace for such metrics is created "ContainerInsights/Prometheus". It only has 6 metrics which are useless for the application and cluster monitoring (all otel_processpr metrics only).
2. Customer chooses to specify a subset of metrics. I would to get yelb metrics for example. This may be a change in the pattern, but I need a clear path how customers can set it.

lib/addons/cloudwatch-adot-addon/index.ts
shapirov103: Other than EKS_Cluster, EKS_Namespace and EKS_PodName are there any other dimensions that customer will be interested in? I see metrics selectors but as it stands dimensions are fixed.

lib/addons/cloudwatch-adot-addon/collector-config-cloudwatch.ytpl
```diff
@@ -311,7 +344,7 @@ spec:
             operations:
               - action: add_label
                 new_label: EKS_Cluster
-                new_value: ${CLUSTER_NAME}
```
shapirov103, line 314: Why did it deploy at all before, lol.

lib/addons/cloudwatch-adot-addon/index.ts
```diff
@@ -29,7 +29,17 @@ export interface CloudWatchAdotAddOnProps {
...
+    /**
+     * Metrics name selectors to write to CloudWatch.
+     * @default "['apiserver_request_.*', 'container_memory_.*', 'container_threads', 'otelcol_process_.*']"
+     */
+    metricsNameSelectors?: string[];
+    /**
+     * Pod Labels to select your application pods emitting custom metrics.
+     * @default 'frontend|downstream(.*)'
+     */
+    podLabels?: string;
```
shapirov103, line 42: We should call out that this is podLabelRegex

docs/addons/cloudwatch-adot-addon.md
```diff
@@ -29,7 +29,7 @@ const blueprint = blueprints.EksBlueprint.builder()
   .build(app, 'my-stack-name');
 ```
 
-Pattern # 2 : Overriding property value for different deployment Modes. This pattern deploys an ADOT collector on the namespace specified in `namespace`, name specified in `name` with `daemonset` as the mode to CloudWatch console. Deployment mode can be overridden to any of these values - `deployment`, `daemonset`, `statefulset`, `sidecar`.
+Pattern # 2 : Overriding property value for different deployment Modes. This pattern deploys an ADOT collector on the namespace specified in `namespace`, name specified in `name` with `daemonset` as the mode to visualize metrics in CloudWatch console. Deployment mode can be overridden to any of these values - `deployment`, `daemonset`, `statefulset`, `sidecar`. You can pass required metrics including custom metrics and required pod labels of applications emitting custom metrics to visualize using `metricsNameSelectors`, `podLabels` as parameters as shown below.
```
shapirov103, line 32: I just had a specreq when customer was asking about deploying it for fargate. I assume the sidecar mode is for Fargate, it would help to call it out here.

shapirov103 (commented):
Looks good, i have only a couple of comments with respect to dimensions and some minor naming and doc comments.

lib/addons/cloudwatch-adot-addon/index.ts
shapirov103, line 42: Can you be more specific what this means? Are these labels on pods? If so, why would we add frontend|downstream here? Choices on the demo apps should not drive the default that we apply.

lib/addons/cloudwatch-adot-addon/collector-config-cloudwatch.ytpl
elamaran11: This is was never a problem. Prebiously we used env var name and now we are rendering directly because `"{{clusterName}}"` is rendered in the code.

docs/addons/cloudwatch-adot-addon.md
elamaran11: Yes that is correct. `Sidecar` is supported mainly for Fargate.

lib/addons/cloudwatch-adot-addon/index.ts
elamaran11: This is fixed in code and docs.

lib/addons/cloudwatch-adot-addon/index.ts
```diff
@@ -42,10 +53,12 @@ export const enum cloudWatchDeploymentMode {
 /**
  * Defaults options for the add-on
  */
-const defaultProps = {
+const defaultProps: CloudWatchAdotAddOnProps = {
     deploymentMode: cloudWatchDeploymentMode.DEPLOYMENT,
     namespace: 'default',
-    name: 'adot-collector-cloudwatch'
+    name: 'adot-collector-cloudwatch',
+    metricsNameSelectors: ['apiserver_request_.*', 'container_memory_.*', 'container_threads', 'otelcol_process_.*', 'ho11y*'],
```
shapirov103, line 60: we should drop ho11y from here and check out meaningful default metrics that customers want to see. if apiserver requests is an edge case. 
I would rather prefer metrics such as:

node * (CPU, memory, disk, threads)
pod (CPU, memory, threads)
alb ingress (request count, latency)
nginx (whatever they provide ootb, ideally number of http requests, and latency)

I am sure just based on your experience you can come up with a few really meaningful metrics that customers will need.

elamaran11: @shapirov103 I have dropped `ho11y` in the metric selectors of the addon but this can be passed from our pattern or any specific customer metric the customer is looking for. We are already getting useful metrics with all `metricNameSelectors` we have already. We will get more metrics once we have `kube-state metrics` and `node-exporter` addons. Please let me know if yu have any questions.

shapirov103:
@elamaran11 I submitted a fix for the istio issue and will merge this after e2e pass. One thing that I discovered is that AMP add-on does not work in us-west-1, which was odd to discover. It fails with an obscure error, we will need to think about updating our docs and error messages.

elamaran11:
> @elamaran11 I submitted a fix for the istio issue and will merge this after e2e pass. One thing that I discovered is that AMP add-on does not work in us-west-1, which was odd to discover. It fails with an obscure error, we will need to think about updating our docs and error messages.

@shapirov103 Please check this [link](https://docs.aws.amazon.com/prometheus/latest/userguide/what-is-Amazon-Managed-Service-Prometheus.html) and yes as of today AMP is supported in only 10 regions. `us-west-1` is not suported today. I have updated the docs page with the supported regions link.

elamaran11:
@shapirov103 I have added a comment to the docs page for CW Addon for the kubectl layers issue as referenced in this issue https://github.com/aws/aws-cdk/issues/20263#issuecomment-1252910571

shapirov103 (approved):
LGTM

PR #501, opened 2022-09-15
Fixed the istio-system namespace missing from dependency graph

RadekDvorak:
*Description of changes:*
The istio-system namespace resource was omitted from CF dependency graph. That is why deleting the stack caused trouble as the namespace was deleted before other resources placed there. 


shapirov103 (approved):
LGTM 
Thank you for your contribution - great catch!

Issue #482, opened 2022-08-17
(cluster-providers): How to configure a cluster secrets encryption


pflorek:
### Describe the bug

How do I configure a `SecretsEncryptionKey`in the scope of the EksBlueprint stack?

With the plain AWS CDK EKS module I did the following:

```typescript
    const secretsEncryptionKey = new kms.Key(stack, "Key", {
      removalPolicy: RemovalPolicy.RETAIN,
      alias: "my-cluster",
    });


    this.cluster = new eks.Cluster(stack, "Cluster", {
      secretsEncryptionKey: secretsEncryptionKey,
      ...
    });

```

What's the recommended way to do so?

### Expected Behavior

I can configure the `secretsEncryptionKey` in the scope of the EksBlueprint stack

### Current Behavior

I have to create another stack upfront, then pass the `secretEncryptionKey` to the ClusterProvider

### Reproduction Steps

```typescript
    const secretsEncryptionKey = new kms.Key(stack, "Key", {
      removalPolicy: RemovalPolicy.RETAIN,
      alias: "my-cluster",
    });


    this.cluster = new eks.Cluster(stack, "Cluster", {
      secretsEncryptionKey: secretsEncryptionKey,
      ...
    });

```

### Possible Solution

Create a resource provider?
Don't create a stack by `EksBlueprint.builder().build()`, instead create a construct

### CDK CLI Version

2.37.1

### EKS Blueprints Version

1.2.0

### Node.js Version

18.7.0

### Environment details (OS name and version, etc.)

MacOS

shapirov103:
@pflorek 
I will take a look at registering resource providers for resources that are either created or looked up for cluster creation. 

At the moment you can apply standard OO extension mechanism:

1. Create a custom cluster provider by extending GenericClusterProvider.  E.g. `EncryptedClusterProvider extends GenericClusterProvider (props: EncryptedClusterProviderProps extends GenericClusterProviderProps)`
2. Override method `internalCreateCluster`
3. Create a new clusterOptions object by cloning the `clusterOptions` parameter and setting the `secretEncryptionKey` to the desired value.  You can scope to look up or create new kms key. 

You can control the shape of your cluster entirely with this method, however agree with the premise for this issue and will address.

pflorek:
```typescript
import * as blueprints from "@aws-quickstart/eks-blueprints";
import { RemovalPolicy } from "aws-cdk-lib";
[...]

PR #516, opened 2022-10-03
feat(resource-provider): cluster secrets encryption key

pflorek:
fixes https://github.com/aws-quickstart/cdk-eks-blueprints/issues/482

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/cluster-providers/generic-cluster-provider.ts
+++ b/lib/cluster-providers/generic-cluster-provider.ts
@@ -7,14 +7,16 @@ import * as utils from "../utils";
 import * as constants from './constants';
 import { AutoscalingNodeGroup, ManagedNodeGroup } from "./types";
 import assert = require('assert');
+import {IVpc} from "aws-cdk-lib/aws-ec2";
+import {IKey} from "aws-cdk-lib/aws-kms";
 
 export function clusterBuilder() {
     return new ClusterBuilder();
 }
 
 /**
- * Properties for the generic cluster provider, containing definitions of managed node groups, 
- * auto-scaling groups, fargate profiles. 
+ * Properties for the generic cluster provider, containing definitions of managed node groups,
+ * auto-scaling groups, fargate profiles.
  */
 export interface GenericClusterProviderProps extends eks.ClusterOptions {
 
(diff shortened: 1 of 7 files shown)


shapirov103 (commented):
Looks great!
Can you add an e2e test here: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts#L154

I assume a case when Kms key is created could be a good candidate.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -189,10 +191,10 @@ export class GenericClusterProvider implements ClusterProvider {
             "Mixing managed and autoscaling node groups is not supported. Please file a request on GitHub to add this support if needed.");
     }
 
-    /** 
-     * @override 
+    /**
+     * @override
      */
-    createCluster(scope: Construct, vpc: ec2.IVpc): ClusterInfo {
+    createCluster(scope: Construct, vpc: IVpc, secretsEncryptionKey: IKey): ClusterInfo {
```
shapirov103, line 197: while VPC is a mandatory parameter, secretsEncryptionKey seems to be optional since the default `KmsKeyProvider` (no-arg constructor) returns undefined. Do you see any issue making it an optional parameter here, e.g. `secretEncryptionKey?`?

shapirov103 (commented):
Looks really good, seriously! One minor comment before merging.

lib/resource-providers/kms-key.ts
```diff
@@ -0,0 +1,50 @@
...
+    const id = context.scope.node.id;
+    let key = undefined;
+
+    if (this.aliasName) {
+      key = kms.Key.fromLookup(context.scope, `${id}-kms-key`, {
+        aliasName: this.aliasName,
+      });
+    }
+
+    if (!key) {
```
pflorek, line 41: This will create a new Key if lookup returns null

lib/cluster-providers/generic-cluster-provider.ts
pflorek: > while VPC is a mandatory parameter, secretsEncryptionKey seems to be optional since the default `KmsKeyProvider` (no-arg constructor) returns undefined. Do you see any issue making it an optional parameter here, e.g. `secretEncryptionKey?`?

Should we make the KMS Key Provider optional and not default?

lib/resource-providers/kms-key.ts
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * @param aliasName The alias name to lookup an existing KMS Key in the deployment target, if omitted a key will be created.
+   * @param kmsKeyProps The key props used
+   */
+  public constructor(aliasName?: string, kmsKeyProps?: kms.KeyProps) {
+    this.aliasName = aliasName;
+    this.kmsKeyProps = kmsKeyProps;
+  }
+
+  provide(context: ResourceContext): kms.IKey {
```
pflorek, line 31: @shapirov103 behaves like VpcProvider either lookup or create

shapirov103: Not sure what I saw yesterday night, I should stop reviewing close to midnight.

lib/cluster-providers/generic-cluster-provider.ts
shapirov103: You are right, it is consistent.

shapirov103 (approved):
LGTM

lib/resource-providers/kms-key.ts
pflorek: > Not sure what I saw yesterday night, I should stop reviewing close to midnight.

🚀

Issue #517, opened 2022-10-09
[Question] aws-cdk version of blueprints causes codebuild failed


vumdao:
### Describe the feature

aws-cdk version of blueprints (v2.43.0) does not match with the latest one v.2.45.0 which caused codebuild failed
```
+ aws-cdk@2.45.0
30 | added 1 package from 1 contributor in 3.48s
31 |  
32 | [Container] 2022/10/09 02:48:57 Phase complete: INSTALL State: SUCCEEDED
33 | [Container] 2022/10/09 02:48:57 Phase context status code:  Message:
34 | [Container] 2022/10/09 02:48:57 Entering phase PRE_BUILD
35 | [Container] 2022/10/09 02:48:57 Phase complete: PRE_BUILD State: SUCCEEDED
36 | [Container] 2022/10/09 02:48:57 Phase context status code:  Message:
37 | [Container] 2022/10/09 02:48:57 Entering phase BUILD
38 | [Container] 2022/10/09 02:48:57 Running command yarn install --frozen-lockfile
39 | yarn install v1.22.19
40 | [1/4] Resolving packages...
41 | [2/4] Fetching packages...
42 | [3/4] Linking dependencies...
43 | [4/4] Building fresh packages...
44 | Done in 40.46s.
45 |  
46 | [Container] 2022/10/09 02:49:39 Running command npx projen build
47 | 👾 build » default \| ts-node --project tsconfig.dev.json .projenrc.ts
48 | yarn install v1.22.19
49 | [1/4] Resolving packages...
50 | [2/4] Fetching packages...
51 | [3/4] Linking dependencies...
52 | [4/4] Building fresh packages...
53 | success Saved lockfile.
54 | Done in 13.73s.
55 | 👾 build » post-compile » synth:silent \| cdk synth -q
56 | ⨯ Unable to compile TypeScript:
57 | src/eks-blueprints/builder.ts(31,13): error TS2345: Argument of type 'Cluster' is not assignable to parameter of type 'Cluster \| PromiseLike<Cluster>'.
58 | Type 'import("/codebuild/output/src259311067/src/node_modules/@aws-quickstart/eks-blueprints/node_modules/aws-cdk-lib/aws-eks/lib/cluster").Cluster' is not assignable to type 'import("/codebuild/output/src259311067/src/node_modules/aws-cdk-lib/aws-eks/lib/cluster").Cluster'.
59 | The types of 'vpc.publicSubnets' are incompatible between these types.
[...]

PR #519, opened 2022-10-10
fix(deps): allow aws-cdk-lib greater or equal than 2.43.0

pflorek:
https://github.com/aws-quickstart/cdk-eks-blueprints/issues/517

*Issue #, if available:*

*Description of changes:*


shapirov103:
@pflorek please check out the discussion thread in #517. I am open to discuss it, but fixed CDK version was a conscious choice.

Provided your npm version is great than 8.1.4 you can just leverage package level overrides to use a different version, but it becomes very easy for us to 1/ know version was used and 2/ see if that version was something that we tested with.

Issue #518, opened 2022-10-10
(feat: addons Karpenter) allow setting ttlSecondsUntilExpired


javydekoning:
### Describe the feature

The provisioner spec allows setting `ttlSecondsUntilExpired`

```
# If omitted, the feature is disabled and nodes will never expire.  If set to less time than it requires for a node
# to become ready, the node may expire before any pods successfully start.
ttlSecondsUntilExpired: 2592000 # 30 Days = 60 * 60 * 24 * 30 Seconds;
```

### Use Case

I want my nodes to have a maximum lifetime so they are automatically launched from newer AMI's.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

N/A

### EKS Blueprints Version

1.3.0

### Node.js Version

N/A

### Environment details (OS name and version, etc.)

N/A

youngjeong46:
@javydekoning This has been implemented as part of 1.4 release. Will close the issue for now, but feel free to try and re-open the issue if you see any problems.

PR #527, opened 2022-11-02
Karpenter addon to allow setting ttlSecondsUntilExpired

youngjeong46:
*Issue #, if available:*
#518 

*Description of changes:*
Minor Karpenter addon change to allow setting ttlSecondsUntilExpired as requested per above issue.

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -60,6 +60,13 @@ interface KarpenterAddOnProps extends HelmAddOnUserProps {
     }
 
     /**
+     * If omitted, the feature is disabled and nodes will never expire.  
+     * If set to less time than it requires for a node to become ready, 
+     * the node may expire before any pods successfully start.
+     */
+    ttlSecondsUntilExpired?: number,
+
+    /**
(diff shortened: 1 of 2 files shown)


lib/addons/karpenter/index.ts
```diff
@@ -202,6 +210,7 @@ export class KarpenterAddOn extends HelmAddOn {
                         subnetSelector: subnetTags,
                         securityGroupSelector: sgTags,
                     },
+                    ttlSecondsUntilExpired: ttlSecondsUntilExpired,
```
shapirov103, line 213: did you validate that null setting does not break the install? I saw cases when null not permitted, however lack of key is OK.

shapirov103 (commented):
Looks good, one comment/clarification before merging.

lib/addons/karpenter/index.ts
youngjeong46: Yes - just validated and it works as intended (no value provided -> null -> no custom helm value provided)

shapirov103 (approved):
LGTM

PR #537, opened 2022-11-14
JupyterHub Addon: 2.0 release, EFS Persistent Storage, and custom notebook stack support

youngjeong46:
*Issue #, if available:*

*Description of changes:* 
- Updated JupyterHub addon to support EFS as a persistent storage
- Revisions to provide custom ingress host and annotation parameters
- Provide options to select Jupyter-maintained notebook stack
- Provide latest helm chart version support (2.0.0)


lib/addons/jupyterhub/index.ts
```diff
@@ -1,23 +1,42 @@
...
         capacity: string,
     }
 
+    /**
+     * Configuration necessary to use EFS as Persistent Volume
+     * @property {cdk.RemovalPolicy} removalPolicy - Removal Policy for EFS (DESTROY, RETAIN or SNAPSHOT)
+     * @property {string} pvcName - Name of the Volume to be used for PV and PVC
+     * @property {string} capacity - Storage Capacity (in Mi or Gi)
+     */
+    efsConfig?: {
```
shapirov103, line 34: I am unclear if both efs and ebs configs are defined what happens? Something takes precedence?
It is worth to define this behavior and call it out in the docs.

lib/addons/jupyterhub/index.ts
```diff
@@ -67,64 +105,206 @@ export class JupyterHubAddOn extends HelmAddOn {
         this.options = this.props as JupyterHubAddOnProps;
     }
 
-    @dependable('EbsCsiDriverAddOn')
+    @dependable('EbsCsiDriverAddOn' || 'EfsCsiDriverAddOn')
```
shapirov103, line 108: You cannot define dependency with a logic OR. What do you think the result is when a boolean operator is applied to strings? Tbh, I would expect a compile error, but may be ts just returns one of the expressions.

You can pass both of the addons separated by `,`  or one of the addons. 

We can extend the dependable decorator to accept a comparison custom function like `(a)=> return a == 'A' or a.equals('B')`

shapirov103 (commented):
@youngjeong46 awesome PR, really love the amount of new features and effort you put into it. I added a few mostly programming style comments, nothing major. 
If addressed this week,  I will include in the release this week, otherwise after RIV.

examples/blueprint-construct/index.ts
```diff
@@ -20,7 +20,7 @@ export interface BlueprintConstructProps {
 export default class BlueprintConstruct {
     constructor(scope: Construct, props: cdk.StackProps) {
 
-        HelmAddOn.validateHelmVersions = true;
+        HelmAddOn.validateHelmVersions = false;
```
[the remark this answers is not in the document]
youngjeong46: This was turned off for testing - addressed.

lib/addons/jupyterhub/index.ts
youngjeong46: Line 117 does an assert check to make sure that only one is defined. Otherwise we get an error. 

I will make the changes on the documentation to note this behavior.

lib/addons/jupyterhub/index.ts
youngjeong46: Hmm - I just tested this and it only checks the first one (EBS). I will make a change on the decorator to accept a comparison.

lib/addons/jupyterhub/index.ts
```diff
@@ -67,64 +105,206 @@ export class JupyterHubAddOn extends HelmAddOn {
...
+            const storageClass = this.options.ebsConfig.storageClass;
+            const ebsCapacity = this.options.ebsConfig.capacity;
+            setPath(values, "singleuser.storage", {
+                "dynamic": { "storageClass": storageClass },
+                "capacity": ebsCapacity
+            });
+        } 
+        
+        // Persistent Storage Setup for EFS
+        if (this.options.efsConfig) {
```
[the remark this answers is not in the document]
youngjeong46: Same as above, the check on L117 should prevent entering this if statement if EBS storage is there.

lib/addons/jupyterhub/index.ts
shapirov103: You can also create two methods in the addons: addEbsStorage and addEbsStorage each with its own decorator. That would be the easiest of each storage method translated in a construct.

You can also add a manual validation using `clusterInfo.getScheduledAddOn` to test.

Decorator while convenient and handy, is the least direct approach: you will still be able to define ebsConfig add dependency on EFS driver and fail.

lib/addons/jupyterhub/index.ts
```diff
@@ -67,64 +105,230 @@ export class JupyterHubAddOn extends HelmAddOn {
...
\ No newline at end of file
+    /**
+     * This is a helper function to create EBS persistent storage
+     * @param {ClusterInfo} clusterInfo - Cluster Info
+     * @param {string} values - Helm Chart Values
+     * @param {string} ebsConfig - EBS Configurations supplied by user
+     * @returns
+     */
+    protected addEbsStorage(clusterInfo: ClusterInfo, values: any, ebsConfig: any){
+        const ebsAddon = 'EbsCsiDriverAddOn';
```
shapirov103, line 205: minor: you don't need this constant. You can safely import the addon and use `getScheduledAddOn(EbsCsiDriverAddOn.name)`.
Example: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/f51c11337607ccd3c0bc98295ab2ce84a77d322d/lib/addons/secrets-store/csi-driver-provider-aws-secrets.ts#L164

youngjeong46: addressed.

shapirov103:
@youngjeong46 did you test the cluster delete flow? e2e failed after hanging for 2.5 hours. 
Some error messages:
```Custom::AWSCDK-EKS-KubernetesResource | jupyterhub-namespace-struct/Resource/Default (jupyterhubnamespacestruct3DC2568A) CloudFormation did not receive a response from your Custom Resource. Please check your logs for requestId [1630bb80-0c53-4a81-aad9-155ddebb8720]. If you are using the Python cfn-response module, you may need to update your Lambda function code so that CloudFormation can attach the updated version.
```

I will kick it off again to see if it persists.

shapirov103:
Update: some ENIs were left over, those provisioned for ELB app/k8s-jupyterh-jupyterh-ce1484d0a6/67f13f7899b03689.
For e2e test we may need to supply a config that does expose ingress (e.g. just a clusterIP service). We apply a similar approach to the nginx ingress addon for the same reasons.

youngjeong46:
@shapirov103 I disabled ingress on the addon. Let's run another e2e and see if it passes.

shapirov103 (approved):
LGTM

PR #539, opened 2022-11-17
Allow passing in `values` to `external-dns`

praveenperera:


shapirov103 (approved):
@praveenperera thank you for catching and fixing this defect! LGTM

Issue #490, opened 2022-08-25
[ADDON] - KNative 


5herlocked:
### Describe the feature

Add support for Kubernetes powered event-driven applications with KNative via an add-on.

### Use Case

Include KNative as a supported add-on through the eks blueprints.

### Proposed Solution

AddOn module which can consume the KNative manifest and apply it to the EKS Cluster.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.37.1

### EKS Blueprints Version

1.2.0

### Node.js Version

16.17.0

### Environment details (OS name and version, etc.)

Amazon Linux 2 - Cloud9

5herlocked:
Attaching pull request #541

PR #541, opened 2022-11-18
Feature: KNative Operator addon

5herlocked:
*Issue #, if available:* #490

*Description of changes:* Add support for the KNative operator v1.8 through a blueprint add-on. Included test ensures that the blueprint is defined after adding the KNative AddOn.
Currently draft as it includes a copy of the operator yaml in the source, will update it with a remote download soon, also need to fix issues with Istio not deploying properly before KNative. Need to also add in the required lines of code to the BlueprintConstruct module for a showcase.


shapirov103 (commented):
@5herlocked thank you for your contribution, it is a great start!

A few comments on the format:

1. Please exclude all redundant files like .DS_Store from the PR (this should be in gitignore). 
2. Can you provide pros and cons for using YAML installation vs KNative operator? It seems the operator is more tailored for production cases and might be easier to support?
3. If there is a helm chart that would be beneficial compared to yaml which will require manual effort to maintain newer versions.
4. The e2e test will need to be updated. We will need to decide whether the blueprint [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts) can be extended with knative for testing. 
5. Docs on usage: e.g. under docs/addons

5herlocked:
Latest update, Moved to remove YAML format that'll get update by KNative (the end user will still have to manually update versions).

1. Done - Removed .DS_STORE @17a44a2
2. YAML installation of the eventing, serving and functions CRDs have custom YAML that aren't parsed by the js-yaml library that is used with CDK-EKS. KNative recommends using the operator or the custom YAML installations for production grade workloads and since we can't use one, I'm using the other. It will definitely be easier to support as we would only need to install the Operator to get full functionality of KNative. While with Eventing, Serving, and Functions installed separately, you would need to maintain parity between the three versions while updating KNative.
3. There is currently no Helm Chart. For explanations I would refer to this long running github issue: https://github.com/knative/docs/issues/311
4. Sounds good!
5. Will add them soon! Is there anything specific that needs to be in there?

shapirov103:
> 5. Will add them soon! Is there anything specific that needs to be in there?
./docs/addons/index.md
./docs/addons/knative.md
./mkdocs.yml (for navigation).

5herlocked:
Updated docs, have tested Knative deployment with it's dependencies (istio) and it deploys and functions well.

youngjeong46 (commented):
A couple of minor comments.

test/knative-addon-validation.test.ts
```diff
@@ -0,0 +1,20 @@
...
+        new IstioControlPlaneAddOn(),
+        new KNativeOperator(),
+    ];
+
+    const blueprint = await blueprints.EksBlueprint.builder()
+        .account('123456789').region('us-east-1')
+        .addOns(...addons)
+        .build(app, 'knative-stack');
+
+    expect(blueprint).toBeDefined();
```
shapirov103, line 19: Blueprint will be defined no matter what. How about expecting blueprint to have a resource of a particular type with the name knative? At least it will protect from someone deleting the body of the addon (do nothing on deploy).

lib/addons/knative-operator/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * 
+     * @default 'https://github.com/knative/operator/releases/download/knative'
+     */
+    base_url?: string;
+}
+
+const defaultProps = {
+    name: 'knative-operator',
+    namespace: 'default',
```
shapirov103, line 37: How common it is to place the operator into the default namespace vs a custom namespace like knative?
If a customer defines the namespace to be custom, how do you see they will approach namespace creation, provided the add-on has no logic to do it?

lib/addons/knative-operator/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        // Load External YAML: https://github.com/knative/operator/releases/download/knative-v1.8.1/operator.yaml
+        const doc = loadExternalYaml(
+            this.knativeAddOnProps.base_url + `-${this.knativeAddOnProps.version}/operator.yaml`
+        ).slice(0, 26); // the last element is null
+
+        const kubectlProvider = new KubectlProvider(clusterInfo);
+
+        const statement = kubectlProvider.addManifest(
+            { manifest: doc, values: {}, name: 'knative-operator', namespace: 'default' }
```
shapirov103, line 64: why namespace `default` is hardcoded here? Customers are allowed to pass a different one from what I observe.

shapirov103 (commented):
A few comments, getting close.

lib/addons/knative-operator/index.ts
5herlocked: Based on my deduction and brief experimentation, the Knative Operator needs to be installed to the default namespace so the associated CRDs can be used in any other namespace.

I know with the YAML deployment, you have to deploy the CRDs per namespace and the Operator avoids that repetition.

lib/addons/knative-operator/index.ts
5herlocked: Changed the behavior though I am going to double-check and test how installing the operator in a non-default namespace affects deployment of Knative in other namespaces.

shapirov103 (approved):
LGTM

PR #546, opened 2022-11-24
Enable awsproxy extension in XRay ADOT Addon

RobKenis:
When using the OpenTelemetry collector with the javaagent, in combination with Sampling Rules in Cloudwatch or Xray, configuration is as follows:
https://aws-otel.github.io/docs/getting-started/remote-sampling

By default, the proxy will run on port 2000.

*Issue #, if available:*

*Description of changes:*
- Enable awsproxy in adot collector. When using CloudWatch Sampling Rules, the awsproxy needs to be enabled. It handles authentication to AWS, so the containers using the javaagent don't have to.

### DIFF
--- a/lib/addons/xray-adot-addon/collector-config-xray.ytpl
+++ b/lib/addons/xray-adot-addon/collector-config-xray.ytpl
@@ -6,6 +6,10 @@ metadata:
 spec:
   mode: "{{deploymentMode}}"
   serviceAccount: adot-collector
+  ports:
+  - name: awsproxy
+    targetPort: 2000
+    port: 2000
   config: |
     receivers:
       otlp:
@@ -20,7 +24,11 @@ spec:
       awsxray:
         region: "{{awsRegion}}"
 
+    extensions:
+      awsproxy:
+
     service:
+      extensions: [awsproxy]
       pipelines:
         traces:
           receivers: [otlp]


shapirov103:
@elamaran11 could you please review this PR when you get a chance? I am hesitant to _always_  add the awsproxy extension, it could a config option, but that will require some rework.

elamaran11:
@RobKenis @shapirov103 Im not supportive to add this feature this at this point. This is not officiall supported in [ADOT XRay AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/configure-xray.html). Please check the reference to [collector-config-xray.yaml](https://github.com/aws-observability/aws-otel-community/blob/master/sample-configs/operator/collector-config-xray.yaml) and this does not support `awsproxy` today. I would recommend working on this with [`aws-otel-community`](https://github.com/aws-observability/aws-otel-community) and once they are able to approve and merge this change, we can work towards this change. Also even in future this should be an option given to user to add and should not be a mandate to use `awsproxy`, this change is looking like it has to be a mandate. At this point i dont approve to merge this change as this might break the current functioning of the addon.

elamaran11 (commented):
We _cant_ approve this at this state. the `aws-otel-community` should support this change in the `collector-config-xray` to proceed. Also this can be mandatory as added here. This should be an option to the user to configure or ignore.

RobKenis:
@elamaran11 I understand that this option should not be forced upon users, so making it optional is the best way to go. However, as per this documentation https://aws-otel.github.io/docs/getting-started/remote-sampling we can enable the proxy by using the config in `OpenTelemetryCollector`. Our current solution is to install the addon and overwrite the `OpenTelemetryCollector` manifest after the installation is done. So the documentation supports this feature, but the blueprint doesn't. Can't we make the addon more flexible to adding extensions? Also, is https://aws-otel.github.io/docs/getting-started/remote-sampling official AWS documentation?

elamaran11:
@RobKenis As discussed above, This is not officially supported in [ADOT XRay AWS documentation](https://docs.aws.amazon.com/eks/latest/userguide/configure-xray.html). Please check the reference to [collector-config-xray.yaml](https://github.com/aws-observability/aws-otel-community/blob/master/sample-configs/operator/collector-config-xray.yaml) and this does not support awsproxy today. I would recommend working on this with [aws-otel-community](https://github.com/aws-observability/aws-otel-community) and once they are able to approve and merge this change, we can work towards this change. Once they approve this, we can find a way to allow external file for configuration vs modifying the existing one.

shapirov103:
@elamaran11 can you confirm if awsproxy extension can be added as a separate extensions, e.g. a private add-on that maybe depends on xray ADOT add-on? If so, I suggest posting that approach here and closing the PR.

elamaran11:
@RobKenis If Remote sampling is a feature you are looking for, I would recommend submitting a new PR with a new Addon. The new add-on can have an approach to overwrite the OpenTelemetryCollector custom resource after the XRay Adot addon is installed. We will take a look in to your PR, Run some tests and approve the same.

elamaran11:
Closing this Issue.

PR #547, opened 2022-11-25
Feature/aws controller for k8s

elamaran11:
*Issue #, if available:*

*Description of changes:*

New Addon to create ACK Controller for respective AWS Service supported by ACK Community.


lib/addons/ack/index.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * User provided option for the Helm Chart
+ */
+export interface AckAddOnProps extends HelmAddOnUserProps {
+    
+    /**
+     * Name of the ack controller
+     * @default iam-chart
+     */
+    name?: string;
```
shapirov103, line 17: Please remove all redundant attributes that are already declared in the HelmAddOnUserProps. Applies to name, chart, version, release, repository.

lib/addons/ack/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * To Create Namespace using CDK. This should be done only for the first time.
+     */    
+    createNamespace?: boolean;
+}
+
+/**
+ * Default props to be used when creating the Helm chart
+ */
+const defaultProps: HelmAddOnProps & AckAddOnProps = {
```
shapirov103, line 52: Why `HelmAddOnProps &` and not just AckAddOnProps? I suggest keeping just AckAddOnProps, unless you see a good reason to do otherwise.

lib/addons/ack/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * populateValues populates the appropriate values used to customize the Helm chart
+ * @param helmOptions User provided values to customize the chart
+ */
+function populateValues(helmOptions: AckAddOnProps, awsRegion: string): Values {
+  const values = helmOptions.values ?? {};
+  setPath(values, "aws.region", awsRegion);
+  setPath(values,"serviceAccount.create", false);
+  setPath(values,"serviceAccount.name", helmOptions.chart);
```
shapirov103, line 106: You are setting serviceAccount.name to helmOptions.chart - looks wrong. Should be set to the name that you generated. 
Has it worked for you in tests?

examples/blueprint-construct/index.ts
```diff
@@ -74,6 +75,19 @@ export default class BlueprintConstruct {
...
             new blueprints.addons.KubeProxyAddOn(),
             new blueprints.addons.OpaGatekeeperAddOn(),
+            new blueprints.addons.AckAddOn({
+                skipVersionValidation: true
+            }),
+            new blueprints.addons.AckAddOn({
+                skipVersionValidation: true,
+                name: "rds-chart",
+                chart: "rds-chart",
+                version: "v0.1.1",
```
shapirov103, line 85: is this the latest version?

examples/blueprint-construct/index.ts
```diff
@@ -74,6 +75,19 @@ export default class BlueprintConstruct {
             new blueprints.addons.CoreDnsAddOn(),
             new blueprints.addons.KubeProxyAddOn(),
             new blueprints.addons.OpaGatekeeperAddOn(),
+            new blueprints.addons.AckAddOn({
+                skipVersionValidation: true
+            }),
+            new blueprints.addons.AckAddOn({
+                skipVersionValidation: true,
+                name: "rds-chart",
+                chart: "rds-chart",
```
shapirov103, line 84: The blueprint should make it easy for the customers to use add-on. They don't need to know the oci repo and convention between AWS service name, chart name, repo, etc. See my other comment on the design of the addon props.

lib/addons/ack/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import { ManagedPolicy } from 'aws-cdk-lib/aws-iam';
+import { Construct } from 'constructs';
+import merge from "ts-deepmerge";
+import { ClusterInfo, Values } from "../../spi";
+import { createNamespace, setPath } from "../../utils";
+import { HelmAddOn, HelmAddOnProps, HelmAddOnUserProps } from "../helm-addon";
+/**
+ * User provided option for the Helm Chart
+ */
+export interface AckAddOnProps extends HelmAddOnUserProps {
```
shapirov103, line 11: We need to design it a bit better. I suggest just letting customers pass the service name, like iam or rds as one field and we apply the logic to inject it to the repo, chart, release, sa. 

We should also supply an enum of supported service names:

```typescript
export enum AckServiceName {
   IAM = "iam",
   RDS = "rds" // etc.
}
```

Another thing that would be really helpful is managed policy mapping with each service, e.g. IAM -> IAMFullAccess

shapirov103, line 11: Another issue is that this add-on will need to have a unique identifier per each service. I assume this may collapse if more than one AckAddOn is added. 
For example I need to add a dependency on IAM ACK addon,  but I have both IAM and RDS in the blueprint. How can we differentiate them? There is an ID field for each add-on for that which should be populated: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/spi/addon-contracts.ts#L18

docs/addons/ack-addon.md
```diff
@@ -0,0 +1,115 @@
...
+
+> Pattern # 1 : This installs AWS Controller for Kubernetes for IAM ACK Controller. This uses all default parameters for installation of the IAM Controller.
+
+```typescript
+import * as cdk from 'aws-cdk-lib';
+import * as blueprints from '@aws-quickstart/eks-blueprints';
+
+const app = new cdk.App();
+
+const addOn = new blueprints.addons.AckAddOn({
```
shapirov103, line 19: Please drop skipVersionValidation - the fix is merged now, it will skip OCI. Please let me know if you see any issues.

docs/addons/ack-addon.md
```diff
@@ -0,0 +1,115 @@
...
+  skipVersionValidation: true,
+  name: "rds-chart",
+  chart: "rds-chart",
+  version: "v0.1.1",
```
shapirov103, line 40: is version specific to every ACK chart?

docs/addons/ack-addon.md
```diff
@@ -0,0 +1,115 @@
...
+## Configuration Options
+
+- `name`: Name of the ACK Chart
+- `chart`: Chart Name of the ACK Chart
+- `version`: Version of the ACK Chart
+- `release`: Release Name of the ACK Chart
+- `repository`: Repository URI of the specific ACK Chart
+- `managedPolicyName`: Policy Name required to be added to the IAM role for that ACK
+- `createNamespace`: (boolean) This should be false if you are using for the second time
+- `values`: Arbitrary values to pass to the chart. Refer to the cert-manager [Helm Chart documentation](https://artifacthub.io/packages/helm/cert-manager/cert-manager) for additional details. It also supports all standard helm configuration options ( for Eg: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/addons/index.md#standard-helm-add-on-configuration-options)
```
shapirov103, line 85: Please drop the reference to the cert-manager here.

shapirov103 (commented):
Love the PR and the effort put into it, including the doc. 

I added a few comments, some minor but a couple are design level. Happy to discuss.

lib/addons/ack/index.ts
elamaran11: Done.

lib/addons/ack/index.ts
elamaran11: This wont work and this line `super({...defaultProps, ...props});` fail.

lib/addons/ack/index.ts
elamaran11: I added a new variable for `saName`.

examples/blueprint-construct/index.ts
elamaran11: Yes. You can check the ECR link https://gallery.ecr.aws/aws-controllers-k8s/ for all 47 charts.

lib/addons/ack/index.ts
elamaran11: This will make it very difficult to maintain it as ACK constantly adds one service after the other. Currently 47 services are supported today. Adding and maintaining is each month is something we dont want to do. Lets talk about it.

examples/blueprint-construct/index.ts
elamaran11: This will make it very difficult to maintain it as ACK constantly adds one service after the other. Currently 47 services are supported today. Adding and maintaining is each month is something we dont want to do. Lets talk about it.

lib/addons/ack/index.ts
elamaran11: Lets talk about this too. I added both `iam` and `ec2` and it worked.

docs/addons/ack-addon.md
elamaran11: Done

docs/addons/ack-addon.md
elamaran11: Yes there is one specific version to each chart check this link - https://gallery.ecr.aws/aws-controllers-k8s/

docs/addons/ack-addon.md
elamaran11: Done.

shapirov103 (commented):
Mostly minor comments.

PR #551, opened 2022-12-05
Minor JupyterHub fixes

youngjeong46:
*Issue #, if available:* None

*Description of changes:* Unit test added for jupyterhub

### DIFF
--- a/lib/addons/jupyterhub/index.ts
+++ b/lib/addons/jupyterhub/index.ts
@@ -107,7 +107,8 @@ export class JupyterHubAddOn extends HelmAddOn {
         super({...defaultProps, ...props});
         this.options = this.props as JupyterHubAddOnProps;
     }
-
+    
+    @dependable(AwsLoadBalancerControllerAddOn.name)
     deploy(clusterInfo: ClusterInfo): Promise<Construct> {
         const cluster = clusterInfo.cluster;
         let values = this.options.values ?? {};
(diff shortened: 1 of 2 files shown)


lib/addons/jupyterhub/index.ts
```diff
@@ -176,6 +175,8 @@ export class JupyterHubAddOn extends HelmAddOn {
         setPath(values, "ingress.enabled", enableIngress);
 
         if (enableIngress){
+            const dep = clusterInfo.getScheduledAddOn(AwsLoadBalancerControllerAddOn.name);
```
shapirov103, line 178: is this related to the order of how dependencies are added? I am a bit unclear to why change from dependable to scheduled addon?

lib/addons/jupyterhub/index.ts
```diff
@@ -108,7 +108,6 @@ export class JupyterHubAddOn extends HelmAddOn {
         this.options = this.props as JupyterHubAddOnProps;
     }
 
-    @dependable(AwsLoadBalancerControllerAddOn.name)
     deploy(clusterInfo: ClusterInfo): Promise<Construct> {
```
shapirov103, line 111: Please address the github actions warning: dependable is defined but never used.

shapirov103 (commented):
A couple of small comments

lib/addons/jupyterhub/index.ts
youngjeong46: I just realized this change is not necessary. Will remove this. (still need to run dependable decorator for NLB)

lib/addons/jupyterhub/index.ts
youngjeong46: see above.

shapirov103 (approved):
LGTM

PR #571, opened 2023-01-16
Allow multiple ExternalDNS addons

praveenperera:
rationale: needed to work with having both a internal and external loadbalancer: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/public-private-route53.md

*Description of changes:*

Uses the chart name to create service account and as the id for the deployment on CDK

### DIFF
--- a/lib/addons/external-dns/index.ts
+++ b/lib/addons/external-dns/index.ts
@@ -46,7 +46,7 @@ export class ExternalDnsAddOn extends HelmAddOn {
         const cluster = clusterInfo.cluster;
         const namespace = this.options.namespace ?? this.options.name;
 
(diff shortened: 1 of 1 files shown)


shapirov103 (commented):
@praveenperera thank you for your contribution! 
I am a bit curious on your use case, so a couple of questions: 

1. why do you need several add-ons in the same blueprint?  
You can pass multiple hosted zone to a single one, so it must be something else. 

2. Are you using separate namespaces/ service accounts for them or reuse the namespace?

The changes make sense, I can merge once I understand it better.

shapirov103:
Makes sense, approved, will merge provided e2e tests pass.

shapirov103 (approved):
LGTM

Issue #554, opened 2022-12-08
[ADDON] - AWS Batch on EKS


youngjeong46:
### Describe the feature

An addon that lets users leverage EKS for AWS Batch Jobs.

### Use Case

AWS Batch on Amazon EKS is a managed service for scheduling and scaling batch workloads into existing EKS clusters. AWS Batch doesn't create, administer, or perform lifecycle operations of your EKS clusters on your behalf. AWS Batch orchestration scales up and down nodes managed by AWS Batch and run pods on those nodes.

### Proposed Solution

Implementation using addon, and a team to execute jobs.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.51.1

### EKS Blueprints Version

1.5.2

### Node.js Version

v16.16.0

### Environment details (OS name and version, etc.)

OSX 12.6.1

youngjeong46:
Completed as of 1.6.0 Release. Closing.

PR #578, opened 2023-01-20
AWS Batch on EKS Addon

youngjeong46:
*Issue #, if available:* 
#554 

*Description of changes:*
Implementation of AWS Batch on EKS using an add-on.


shapirov103 (approved):
Test

shapirov103 (commented):
test

shapirov103 (changes requested):
test

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
+import { assertEC2NodeGroup } from "../../cluster-providers";
+import { ClusterInfo, Values } from "../../spi";
+import { ApplicationTeam, TeamProps } from "../team";
+import { KubectlProvider, ManifestDeployment } from "../../addons/helm-addon/kubectl-provider";
+import { loadYaml, readYamlDocument } from "../../utils/yaml-utils";
+import { Construct } from "constructs";
```
shapirov103, line 6: Please address github action warning `Construct` is defined but never used as well as other warnings elsewhere.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+const NAMESPACE = 'aws-batch';
+
+/**
+ * Interface to define an AWS Batch on EKS team
+ */
+export interface BatchEksTeamProps extends TeamProps {
+  /**
+   * Name of the team
+   */
+  name: string,
```
shapirov103, line 24: why name and namespace repeated here? they are inherited from the TeamProps, let's drop them.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Namespace name for AWS Batch
+   */
+  namespace: string
+
+  /**
+   * Allocation strategies for EKS Compute environment
+   */
+  allocationStrategy?: 'BEST_FIT_PROGRESSIVE' | 'SPOT_CAPACITY_OPTIMIZED'
```
shapirov103, line 34: May be an enum would work better here. At least a spot to add js comment to describe what it means, as well as no strings.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  jobQueueName: string
+
+  priority: number,
```
shapirov103, line 38: what could it be? Can be negative or 1 billion? Is there a range?

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+   * @param {BatchEksTeamProps} props the Batch team definition {@link BatchEksTeamProps}
+   */
+  constructor(props: BatchEksTeamProps) {
+    super({...defaultProps, ...props});
+    this.batchTeam = props;
+  }
+
+  setup(clusterInfo: ClusterInfo): void {
+    const cluster = clusterInfo.cluster;
+    const allocStr = this.batchTeam.allocationStrategy || 'BEST_FIT_PROGRESSIVE';
```
shapirov103, line 67: why this check?  You set default explicitly, can it ever be unset?

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const AwsBatchAddOn = clusterInfo.getProvisionedAddOn('AwsBatchAddOn');
```
shapirov103, line 69: minor: all variables start with lower-case.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    if (AwsBatchAddOn === undefined) {
+      throw new Error("AwsBatchAddOn must be deployed before creating AWS Batch on EKS team.");
+    }
+
+    // Set AWS Batch namespace and necessary RBACs
+    const ns = this.setBatchEksNamespace(clusterInfo, this.batchTeam.namespace);
+
+    // Create compute environment
+    const computeEnv = this.setComputeEnvironment(clusterInfo, this.batchTeam.namespace, ns, allocStr);
```
shapirov103, line 79: should not computeEnv depend on awsBatchAddon to make sure that one is provisioned ahead of the team?

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+      name: 'aws-batch-rbacs',
+      namespace: namespace,
+      manifest,
+      values
+    };
+
+    const kubectlProvider = new KubectlProvider(clusterInfo);
+    const statement = kubectlProvider.addManifest(manifestDeployment);
+
+    const batchEksNamespace = createNamespace(namespace, clusterInfo.cluster, true, true);
```
shapirov103, line 124: why not put the namespace in the yptl file to start with, rather than deal with wiring dependency/order for execution here?

lib/teams/aws-batch/aws-batch-on-eks-team.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    // Need to create instance profile for the nodegroup role
+    const instanceProfile = new iam.CfnInstanceProfile(cluster, 'ng-role-instance-profile',{
+      instanceProfileName: ngRole,
+      roles: [ngRole]
+    })
+
+    const batchComputeEnv = new batch.CfnComputeEnvironment(cluster, "batch-eks-compute-environment", {
+      type: 'MANAGED',
+      computeEnvironmentName: 'My-Eks-CE',
```
shapirov103, line 144: why this name is hardcoded to `my-eks-ce`?

lib/addons/aws-batch-on-eks/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    const cluster = clusterInfo.cluster;
+
+    // Create the service role used by AWS Batch on EKS 
+    const batchOnEksSlr = new CfnServiceLinkedRole(cluster.stack, 'BatchServiceRole', {
+      awsServiceName: 'batch.amazonaws.com',
+    });
+
+    //Init the service role as IRole because `addRoleMapping` method does not
+    //support the CfnServiceLinkedRole type
+    const batchEksServiceRole: IRole = Role.fromRoleArn(
```
shapirov103, line 19: You are using Role.fromArn which assumes the role already exists with the name `AWSServiceRoleForBatch`. What creates that role? I don't see you specifying that name anywhere.

shapirov103 (commented):
@youngjeong46 Valuable PR for sure! 
Added a few comments/questions, might take a few cycles to complete the review.
Please also add the docs.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: whole number. i have not tested the limits of what it could be (and it's certainly something not documented anywhere). I'll put in some assert statements along the way.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: Fixed with export enum, and some description of each choice.

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed

lib/teams/aws-batch/aws-batch-on-eks-team.ts
youngjeong46: addressed by adding a team prop for compute environment name

lib/addons/aws-batch-on-eks/index.ts
youngjeong46: When the Compute Environment is created, it creates the role with that exact name. However, compute environment is currently created inside the teams and not in the addon. I'll need to think of a way to address this.

youngjeong46: So when we create the service role (line 13-15), it will automatically generate the exact ARN: `arn:aws:iam::<acccountid>:role/aws-service-role/batch.amazonaws.com/AWSServiceRoleForBatch` (no suffix is allowed for service role names)

The reason we truncate between `role` and `AWSServiceRoleForBatch` is because there is an issue with aws-auth if we use the full role ARN. See note at the bottom of Step 1 [here](https://docs.aws.amazon.com/batch/latest/userguide/getting-started-eks.html#getting-started-eks-step-1) and github issue on this [here](https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/268).

examples/blueprint-construct/index.ts
```diff
@@ -105,7 +105,7 @@ export default class BlueprintConstruct {
             new blueprints.addons.KarpenterAddOn({
                 requirements: [
                     { key: 'node.kubernetes.io/instance-type', op: 'In', vals: ['m5.2xlarge'] },
-                    { key: 'topology.kubernetes.io/zone', op: 'NotIn', vals: ['us-west-2c']},
+                    { key: 'topology.kubernetes.io/zone', op: 'In', vals: ['us-west-2a','us-west-2b','us-west-2c']},
```
shapirov103, line 108: Intentional change to include all azs in karpenter?

examples/blueprint-construct/index.ts
```diff
@@ -115,11 +115,11 @@ export default class BlueprintConstruct {
                 securityGroupTags: {
                     "kubernetes.io/cluster/blueprint-construct-dev": "owned",
                 },
-                taints: [{
-                    key: "workload",
-                    value: "test",
-                    effect: "NoSchedule",
-                }],
+                // taints: [{
```
shapirov103, line 118: Why commented out?

lib/addons/aws-batch-on-eks/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Construct } from "constructs";
+
+const BATCH = 'aws-batch';
+
+export class AwsBatchAddOn implements ClusterAddOn {
+  deploy(clusterInfo: ClusterInfo): Promise<Construct> {
+    const cluster = clusterInfo.cluster;
+
+    // Create the service role used by AWS Batch on EKS 
+    const batchOnEksSlr = new CfnServiceLinkedRole(cluster.stack, 'BatchServiceRole', {
```
shapirov103, line 13: This role is created once per account. If it is already created, will this fail? I was not able to tell by looking at `CfnServiceLinkedRole` CDK docs, it appears that it may fail. As we discussed, if the role is only expected to be created once per account, we could supply a flag to control this behavior.

shapirov103 (commented):
@youngjeong46 looks great, just a couple of last comments. I will kick off testing proactively, if you address the SLR question I will run it again.

examples/blueprint-construct/index.ts
```diff
@@ -5,7 +5,7 @@ import { PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import { Construct } from "constructs";
 import * as blueprints from '../../lib';
 import { AckServiceName, HelmAddOn } from '../../lib';
-import { EmrEksTeamProps } from '../../lib/teams';
+import { EmrEksTeamProps, BatchEksTeamProps, BatchAllocationStrategy, BatchEnvType } from '../../lib/teams';
```
shapirov103, line 8: @youngjeong46 we use wildcard import if more than 3 objects are imported. This seems redundant since we `import * as teams` on line 10. You should be able to use `teams.BatchEksTeamProps` and such.

youngjeong46: the import on line 10 is from the `../teams` as in the custom teams from examples. This import is from our standard library of teams. Regardless your comment stands and I will use wildcards.

examples/blueprint-construct/index.ts
youngjeong46: changed from local testing. Will restore.

examples/blueprint-construct/index.ts
youngjeong46: commented from local testing. Will restore.

shapirov103 (approved):
lgtm

PR #582, opened 2023-01-23
[do not merge] Create dedicated IAM role for VpcCniAddOn

### LINKED ISSUE #434: (VPC CNI Addon): (VPC CNI Addon not creating required IAM Roles and not annotating the Service Account for the plugin to work)
### Describe the bug

VPC CNI Addon is not working as per the documentation https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html
1. It does create the required IAM roles for IRSA to work.
2. It does not annotate the SA with right IAM Role.
3. Currently the plugin is not working as expected.

### Expected Behavior

VPC CNI Plugin should work as expected in the documentation - https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html

### Current Behavior

VPC CNI Addon is not working as per the documentation :
1. It does create the required IAM roles for IRSA to work.
2. It does not annotate the SA with right IAM Role.
3. Currently the plugin is not working as expected.

### Reproduction Steps

Stand up new cluster, specified k8 version 1.21, vpc-cni addon preferably use the https://github.com/aws-samples/cdk-eks-blueprints-patterns/tree/main/lib/multi-team-construct

### Possible Solution

Rework on the VPC CNI plugin to get the addon working as expected.

### Additional Information/Context

_No response_

### CDK CLI Version

2.29.0

### EKS Blueprints Version

1.0.4

### Node.js Version

18.4.0

Amazon Linux.

### Other information

_No response_

Feder1co5oave:
Create dedicated IAM role for VpcCniAddOn with policy attached, according to recommendation from https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html

Remove the default 'AmazonEKS_CNI_Policy' from the node role used for the managed node group, since it is no longer necessary.

Fixes #434.

Local tests (`npm test`) run successfully, but a deploy on **an existing cluster built with eks-blueprints 1.5.4 fails when trying to replace the managed node group**:

```
Resources:
[~] AWS::EKS::Nodegroup EKS/Nodegroup-EKS-ng-ng EKSNodegroupEKSngng84084AE6 replace
 └─ [~] NodeRole (requires replacement)
     └─ [~] .Fn::GetAtt:
         └─ @@ -1,4 +1,4 @@
            [ ] [
            [-]   "EKSNodegroupEKSngngNodeGroupRole4B998ECE",
            [+]   "NodeGroupRoleB2FF2978",
            [ ]   "Arn"
            [ ] ]
```

Result of changeset apply:
```
3:43:54 PM | UPDATE_FAILED        | AWS::EKS::Nodegroup                   | EKSNodeg...useEKSngng84084AE6
Resource handler returned message: "NodeGroup already exists with name EKS-ng and cluster name EKS (Service: Eks, Status Code: 409, Request ID: 0486189b-6ec8-46f7
-a047-85f0323443c0)" (RequestToken: 86f57db1-4ab2-23b3-034f-bac46d98009e, HandlerErrorCode: AlreadyExists)
```

Does this block this code change from getting merged? I'm afraid it will bring errors to users' builds.
**Alternatively, one could merge only the creation of the custom role for the VPC CNI addon, and then change the [default permission policies given to the node role in CDK](https://github.com/aws/aws-cdk/blob/cb7367965608eb733f7ae03a06419bba4b21c713/packages/%40aws-cdk/aws-eks/lib/managed-nodegroup.ts#L378-389) to the minimum required as specified in the docs.**

### DIFF
--- a/lib/addons/vpc-cni/iam-policy.ts
+++ b/lib/addons/vpc-cni/iam-policy.ts
@@ -1,0 +1,35 @@
+import { PolicyDocument } from "aws-cdk-lib/aws-iam";
+
+export function getVpcCniDriverPolicyDocument() : PolicyDocument {
+    return PolicyDocument.fromJson({
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+                "Effect": "Allow",
+                "Action": [
+                    "ec2:AssignPrivateIpAddresses",
+                    "ec2:AttachNetworkInterface",
+                    "ec2:CreateNetworkInterface",
+                    "ec2:DeleteNetworkInterface",
+                    "ec2:DescribeInstances",
+                    "ec2:DescribeTags",
+                    "ec2:DescribeNetworkInterfaces",
+                    "ec2:DescribeInstanceTypes",
+                    "ec2:DetachNetworkInterface",
+                    "ec2:ModifyNetworkInterfaceAttribute",
+                    "ec2:UnassignPrivateIpAddresses"
+                ],
+                "Resource": "*"
+            },
+            {
+                "Effect": "Allow",
+                "Action": [
+                    "ec2:CreateTags"
+                ],
+                "Resource": [
+                    "arn:aws:ec2:*:*:network-interface/*"
+                ]
+            }
+        ]
+    });
+}
--- a/lib/addons/vpc-cni/index.ts
+++ b/lib/addons/vpc-cni/index.ts
@@ -1,4 +1,5 @@
 import { CoreAddOn } from "../core-addon";
+import { getVpcCniDriverPolicyDocument } from "./iam-policy";
 
 /**
  * Implementation of VpcCni EKS add-on.
@@ -9,7 +10,8 @@ export class VpcCniAddOn extends CoreAddOn {
         super({
             addOnName: "vpc-cni",
             version: version ?? "v1.12.0-eksbuild.1",
-            saName: "vpc-cni"
+            saName: "vpc-cni",
+            policyDocumentProvider: getVpcCniDriverPolicyDocument
         });
     }
 }
--- a/lib/cluster-providers/generic-cluster-provider.ts
+++ b/lib/cluster-providers/generic-cluster-provider.ts
@@ -14,6 +14,7 @@ import * as utils from "../utils";
 import * as constants from './constants';
 import { AutoscalingNodeGroup, ManagedNodeGroup } from "./types";
 import assert = require('assert');
+import * as iam from "aws-cdk-lib/aws-iam";
 
 export function clusterBuilder() {
     return new ClusterBuilder();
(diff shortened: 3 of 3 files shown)


shapirov103 (commented):
@Feder1co5oave looks great, valuable change!

However, with respect to the implementation, at this point we need to make sure there is continuity for existing customers. So the IAM policy for the vpc-cni add-on should be an opt-in. 

With respect to the change in the generic cluster provider: 
1. This change is safe only if customers use the vpc-cni add-on. It will be a breaking change for any blueprint that is not using vpc-cni add-on (for whatever reason). 
2. Would it make sense if the framework makes it easy to pass a node role and we supply a utility method to create a role that has only the required policies?

Feder1co5oave:
Apart from the node group issue, even the first commit doesn't actually work.
The VPC-CNI addon deploys a daemonset `aws-node` which runs via the `aws-node` serviceaccount, created internally by the addon chart.
Currently the blueprint add-on will create a `vpc-cni` serviceaccount, the associated IAM role and policies, and annotate that serviceaccount with the `eks.amazonaws.com/role-arn` annotation. Meanwhile the daemonset will run with the `aws-node` serviceaccount and will not be able to perform the necessary calls to the ec2 API (returns 403 Forbidden). The daemonset will be unhealthy, the VPC CNI addon degraded, and the `cdk deploy` will fail after a painful 1-hour timeout. Any setup scheduled after the VpcCniAddOn will not be applied (e.g. KarpenterAddOn and any teams you added). AwsAuth remains untouched.

If I change the `saName` in `VpcCniAddOn` to `aws-node`, the deploy fails because the `KubernetesManifest` resource of the serviceaccount wants be be `CREATED`, and when the KubectlProvider runs `kubectl create` it will fail with

> Error from server (AlreadyExists): error when creating "/tmp/manifest.yaml": serviceaccounts "aws-node" already exists

I don't know how to proceed from here. Is there any way to annotate an existing serviceaccount without creating it?

I'm not even sure how the addon gets created before the serviceaccount, given the `cfnAddon.node.addDependency(serviceAccount)` in CoreAddOn ?

shapirov103:
the PR points out that there is a defect in the VPC-CNI add-on provisioning incorrect SA name (expected aws-node, but was vpc-cni). we will address it.

shapirov103:
This was address in #636 , [example](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts#L200) was provided with removal of node level CNI permissions and using VPC CNI IRSA.
You will need to opt-in and pass policies to the vpc cni add-on as shown in the example. 

I will keep it for a bit in case of questions, will close shortly.

Feder1co5oave:
@shapirov103 that's great, thank you for fixing this

PR #585, opened 2023-01-26
fix: allow multiple DelegatingHostedZoneProvider instances

aceat64:
*Issue #, if available:* n/a

*Description of changes:*

It is not currently possible to have multiple instances of DelegatingHostedZoneProvider, as the id for `delegationRole` is set to a static string. This causes the following error:

> Error: There is already a Construct with name 'MutableRoleDelegationRole' in EksBlueprint


shapirov103 (approved):
@aceat64 Thank you for catching it, LGTM.

PR #586, opened 2023-01-28
Added KMS part of the IAM policy for the EBS CSI Driver optionally.

sbollers:
*1/Added IAM policy for the EBS CSI Driver to include a conditional KMS policy statement when KMS keys are passed. 2/Modified EBS CSI driver addon constructor to accept an optional kmsKeys array.*

### DIFF
--- a/lib/addons/ebs-csi-driver/iam-policy.ts
+++ b/lib/addons/ebs-csi-driver/iam-policy.ts
@@ -174,7 +174,6 @@ export function getEbsDriverPolicyDocument(partition: string, kmsKeys?: string[]
                 "Resource": kmsKeys
             }
         ];
-        // @ts-ignore
         result.Statement.push(kmsPolicy);
     }
     return PolicyDocument.fromJson(result);
--- a/lib/addons/ebs-csi-driver/iam-policy.ts
+++ b/lib/addons/ebs-csi-driver/iam-policy.ts
@@ -1,4 +1,5 @@
 import { PolicyDocument } from "aws-cdk-lib/aws-iam";
+import * as kms from "aws-cdk-lib/aws-kms";
 
 interface Statement {
   Effect: string;
@@ -13,7 +14,7 @@ interface Statement {
 
 export function getEbsDriverPolicyDocument(
   partition: string,
-  kmsKeys?: string[]
+  kmsKeys?: kms.Key[]
 ): PolicyDocument {
   const result: { Version: string; Statement: Statement[] } = {
     Version: "2012-10-17",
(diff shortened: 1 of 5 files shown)


shapirov103:
@sbollers  can you please take a look at why the nodejs.CI check above is failing?

lib/addons/ebs-csi-driver/iam-policy.ts
```diff
@@ -1,149 +1,182 @@
...
+                    "kms:Encrypt",
+                    "kms:Decrypt",
+                    "kms:ReEncrypt*",
+                    "kms:GenerateDataKey*",
+                    "kms:DescribeKey"
+                ],
+                "Resource": kmsKeys
+            }
+        ];
+        // @ts-ignore
```
youngjeong46, line 177: @sbollers can you remove this line? It should pass the build fail.

youngjeong46 (changes requested):

shapirov103:
@sbollers can you check why the above check is failing?

youngjeong46 (dismissed):

youngjeong46 (commented):
A couple of comments:
1. Can you revise the documentations under `docs/addons/ebs-csi-driver` to mention this feature?
2. Has this been tested? Typically we ask that you use `examples/blueprint-construct/index.ts` to deploy a cluster on your AWS account with this feature incorporated, and ensure that the cluster provisions successfully.

shapirov103:
@aliaksei-ivanou @sbollers: please resolve the merge conflict here, can't merge without it.

aliaksei-ivanou:
> 

Thank you for the guiding on the testing process! The test was successful - the ebs csi managed policy contains the KMS keys-related statements as expected. I also added the interface for EbsCsiDriverProps for convenience.

Also updated the EBS CSI driver documentation as you suggested. Please let me know if it looks ok.

lib/addons/ebs-csi-driver/index.ts
```diff
@@ -1,26 +1,42 @@
...
+ */
+interface IEbsCsiDriverAddOnOptions {
+  /**
+   * Version of the driver to deploy
+   */
+  version?: string;
+  /**
+   * List of KMS keys to be used for encryption
+   */
+  kmsKeys?: string[];
```
aliaksei-ivanou, line 15: @shapirov103 do we really need a list of KMS keys? Don't we need just one to encrypt an EBS volume?

shapirov103: We do need a list, because the cluster may have many teams and each team may require its own encryption key. These keys are only fed to the IAM policy here, so I see no reason to restrict it to just one. And the actual tightening of resource access policies to particular keys (as opposed to wildcard) was recommendation from Steve.

aliaksei-ivanou (dismissed):

shapirov103 (approved):
LGTM

Issue #299, opened 2022-02-27
[FEATURE] Support ArgoCD GitOps deployment for AddOns


starchx:
Currently addons are installed via CDK. Some customers prefer to install and manage the addons via GitOps (ArgoCD).

There are two ways for ArgoCD to deploy and manage addons:

- Install each addon as separated ArgoCD `Application` via CDK

@shapirov103 already has working example in [branch](https://github.com/aws-quickstart/ssp-amazon-eks/tree/feature/argo-for-helm).

- Install each addon using ArgoCD [App of Apps](https://argo-cd.readthedocs.io/en/stable/operator-manual/cluster-bootstrapping/#app-of-apps-pattern) strategy.

Ideally we can provide both options and provide the flexibility to our customers.

PR #588, opened 2023-01-31
feat: support argocd addon management

starchx:
*Issue #, if available:*

https://github.com/aws-quickstart/cdk-eks-blueprints/issues/299

*Description of changes:*

A factory class is introduced to make it easy for users to enable or disable GitOps mode via **cluster builder**:

```
# Enable ArgoCD GitOps with each addon `Application` deployed by CDK:
        blueprints.EksBlueprint.builder()
            .addOns(...addOns)
            .enableGitOps()

# Enable ArgoCD GitOps with all addons deployed via App of Apps:
        blueprints.EksBlueprint.builder()
            .addOns(...addOns)
            .enableGitOpsAppOfApps()
```

Two patterns have been added to show how to use this feature: https://github.com/aws-samples/cdk-eks-blueprints-patterns/pull/59

- ArgoCD GitOps

image

- ArgoCD App of Apps

image

image


lib/addons/helm-addon/kubectl-provider.ts
```diff
@@ -48,6 +48,15 @@ export interface HelmChartConfiguration {
      * Optional values for the helm chart. 
      */
     values?: Values
+
+    /**
+     * Indicate the helm chart provided uses dependency mode (https://helm.sh/docs/helm/helm_dependency/).
+     * Dependency mode is widely used in `aws-samples/ssp-eks-add-ons` repository, for example:
```
shapirov103, line 54: The repo changed a while a ago <s>ssp-eks-add-ons</s> eks-blueprints-add-ons
applies elsewhere

lib/addons/helm-addon/kubectl-provider.ts
```diff
@@ -48,6 +48,15 @@ export interface HelmChartConfiguration {
...
+     * https://github.com/aws-samples/ssp-eks-add-ons/blob/main/add-ons/appmesh-controller/Chart.yaml
+     * Dependency mode requires the chart values to be wrapped within the chart name.
+     * This value is only used when AddOns are deployed via GitOps way: `ArgoGitOpsFactory.enableGitOps()`
+     */
+    dependencyMode?: boolean
```
shapirov103, line 59: Let's specify that the default is `true` and this field is only used to `turn off dependency mode in case customers choose to copy the whole helm chart into their repo`

shapirov103, line 59: I suspect there is a better way of doing it now with Argo. The examples that they provide still contain `dependency` for helm, but I can see that in [this](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#declarative) example `Application` CR was extended to speciify helm attributes for external repo. 
We can table it till upgrade to argo 5.x, an issue would help.

lib/spi/gitops-contracts.ts
```diff
@@ -8,7 +8,7 @@ import { GitRepositoryReference, Values } from ".";
 export interface GitOpsApplication {
     name: string,
     repository?: GitRepositoryReference,
-    namespace: string
```
shapirov103, line 11: please remind me why this change as opposed to default to `default` namespace? Is it because namespace is set in GitOps?

shapirov103 (commented):
@starchx Looks very good! I added a few mostly minor comments/questions on this revision. 
P.S.
It is weird that GitOps mode won't apply to core EKS add-on like vpc-cni, ebs-csi and such, e.g customers won't be able to change versions of such add-ons through GitOps.

shapirov103:
@starchx since we are quite close to finishing this, do you mind taking a look at the above comments and bring it to the finish line?

lib/addons/argocd/argo-gitops-factory.ts
```diff
@@ -0,0 +1,58 @@
+import { Construct } from "constructs";
+import { ArgoCDAddOn } from "../../addons";
+import { HelmChartDeployment } from "../helm-addon/kubectl-provider";
+import { ClusterInfo } from "../../../lib/spi";
+import { KubectlProvider } from '../helm-addon/kubectl-provider';
+import { kebabToCamel } from "../../utils";
+
+const original = KubectlProvider.applyHelmDeployment
```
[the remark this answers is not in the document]
starchx: It was my bad - I didn't run `make lint`. It has been ran and fixed.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -66,6 +66,16 @@ export class EksBlueprintProps {
      * Default is true.
      */
     readonly useDefaultSecretEncryption? : boolean  = true;
+
+    /**
+     * Whether to manage cluster addOns using GitOps.
+     */
+    readonly enableGitOps?: boolean = false;
```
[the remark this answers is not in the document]
starchx: Makes sense - updated.

lib/addons/argocd/argo-gitops-factory.ts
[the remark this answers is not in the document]
starchx: 👍  - fixed.

lib/addons/cert-manager/index.ts
```diff
@@ -23,7 +23,7 @@ export interface CertManagerAddOnProps extends HelmAddOnUserProps {
  * Default props to be used when creating the Helm chart
  */
 const defaultProps: HelmAddOnProps & CertManagerAddOnProps = {
-  name: "blueprints-cert-manager-addon",
```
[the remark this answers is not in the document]
starchx: yeah, I am testing this at the moment:

- creating a cluster with old name `blueprints-cert-manager-addon`
- changing name from `blueprints-cert-manager-addon` to `cert-manager`
- update the cluster again

I will update here once the test is completed.

lib/addons/helm-addon/kubectl-provider.ts
starchx: 👍  - fixed.

lib/addons/helm-addon/kubectl-provider.ts
starchx: doc updated as suggested 👍 

the default is set to `true`, not inside the interface, but in `lib/addons/helm-addon/index.ts`:

`const dependencyMode = this.props.dependencyMode ?? true;`

starchx: I think Argo supports external repo, and there was some other reason [this repo](https://github.com/aws-samples/eks-blueprints-add-ons/) uses dependency mode, couldn't recall the reason though 👀

lib/addons/cert-manager/index.ts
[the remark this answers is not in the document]
starchx: As per my testing, the name is only used as a logical name in the CloudFormation. Notice the `props.name` in the code below.

```
    public static applyHelmDeployment = function (clusterInfo: ClusterInfo, props: HelmChartDeployment): Construct {
        return clusterInfo.cluster.addHelmChart(props.name, {
            repository: props.repository,
```
[addHelmChart](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-eks.Cluster.html#addwbrhelmwbrchartid-options)

**HOWEVER - This causes all the deployed cert-manager in cluster resources got DELETED.**

BEFORE NAME CHANGES:

image

image

AFTER NAME CHANGES:

image

image

Looks like I will have to revert this name change. Let me know if you have any other ideas please :)

lib/spi/gitops-contracts.ts
starchx: @shapirov103 , I had to change this from `string` to `string | undefined`  to match the one defined in this type:
https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/helm-addon/kubectl-provider.ts#L16

I think this might be related to this comment here https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/container-insights/index.ts#L15

starchx:
> It is weird that GitOps mode won't apply to core EKS add-on like vpc-cni, ebs-csi and such, e.g customers won't be able to change versions of such add-ons through GitOps.

@shapirov103 , GitOps mode only applies to those addons extending from `HelmAddOn`, like:

```
export class AwsLoadBalancerControllerAddOn extends HelmAddOn {
```

Core addons are not applied:

```
export class VpcCniAddOn extends CoreAddOn {
```

docs/addons/argo-cd.md
```diff
@@ -120,6 +120,66 @@ blueprint.clone('us-west-2')
...
+
+There are two types of GitOps deployments via ArgoCD depending on whether you would like to adopt the [App of Apps](https://argoproj.github.io/argo-cd/operator-manual/cluster-bootstrapping/#app-of-apps-pattern) strategy:
+
+- CDK deploys the `Application` resource for each AddOn enabled, and ArgoCD deploys the actual AddOn via GitOps based on the `Application` resource. Example:
+
+```typescript
+    import * as blueprints from '@aws-quickstart/eks-blueprints';
+
+    // enable gitops bootstrapping with argocd
+    const prodBootstrapArgo = new blueprints.addons.ArgoCDAddOn({
```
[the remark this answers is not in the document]
starchx: image

lib/addons/cert-manager/index.ts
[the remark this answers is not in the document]
shapirov103: Something is going on with cert-manager, unsure if it is related. We started experiencing failures of cert-manager yesterday. I am curious if provisioning of cert-manager works at all at the moment with blueprints. I understand that name change can cause the previous deployment to be deleted, but new resources should be provisioned. One workaround is to change the namespace of the cert-manager deployment.
There is a chance that namespace termination happens after the add-on with the new name is added.

starchx:
@shapirov103  , added a small change to enable `prune` and `autoHeal` on ArgoCD apps.

docs/addons/argo-cd.md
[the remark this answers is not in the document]
shapirov103: Sounds good. I would rather push this into the main first, then review the workload repo support.

lib/addons/argocd/application.ts
```diff
@@ -1,25 +1,34 @@
...
+
+        const flatValues = dot.dot(normalizedValues);
+
         const nameValues = [];
 
-        for( let key in flatValues) {
-            nameValues.push({ name: key, value: `${flatValues[key]}`});
+        for (let key in flatValues) {
+            // Avoid passing the undefined values and empty objects
+            if (flatValues[key] !== undefined) {
```
shapirov103, line 23: This if statement is a little confusing. You do the same action in both cases, why are you using `else if`? Why not just `if(<all conditions>)`?

Also if you are checking for undefined, what is the purpose of testing for `Object.getPrototypeOf`? Why not just `flatValues[key] === undefined` (or `!==  undefined`? Are you checking for empty arrays?

lib/addons/efs-csi-driver/index.ts
```diff
@@ -89,6 +89,8 @@ function populateValues(helmOptions: EfsCsiDriverProps, clusterName: string,
     setPath(values, "clusterName",  clusterName);
     setPath(values, "controller.serviceAccount.create",  false);
     setPath(values, "controller.serviceAccount.name",  serviceAccountName);
+    setPath(values, "node.serviceAccount.create",  false);
```
shapirov103, line 92: What prompted this change? I recall vaguely there was an issue with the EFS SA, seems like the right change, just curious.

lib/addons/helm-addon/kubectl-provider.ts
shapirov103: I meant in jsdoc using `@default`:

```typescript
/**
* @default true
*/
```

We use it here: https://aws-quickstart.github.io/cdk-eks-blueprints/api/
Also vscode support is great for such cases.

shapirov103: No worries, let's get it in, and I will test with argo 5.x and if refactoring is needed I will apply.

shapirov103 (commented):
@starchx almost there, looks great. I added a few minor comments, we can kick off tests after that. Targeting the next release.

lib/addons/efs-csi-driver/index.ts
starchx: EFS CSI Driver addon provisions two kinds of pods in the cluster: `controller` and `node`. Both of them requires IAM permissions to AWS resources. This line passes the same `k8s service account` to the `node` pods, which was missing, otherwise the addon fails to stabilize.

lib/addons/argocd/application.ts
starchx: Yes, I am filtering out `undefined` and `empty objects`.

After flatten using `dot.dot(deployment.values)`, we will have in `flatValues`:

```
  'appmeshController.serviceAccount.create': false,
  'appmeshController.serviceAccount.name': 'appmesh-controller',
  'appmeshController.tracing.enabled': false,
  'appmeshController.tracing.provider': 'x-ray',
  'appmeshController.tracing.address': undefined,   <<=== This will be filtered out
  'appmeshController.tracing.port': undefined,   <<=== This will be filtered out
  'appmeshController.enable': true,
  'something': {}   <<===  This will be filtered out, although I cannot recall how this was generated in the first place..
```

The logic of `If else` is:

1. If value is not defined, skip.
2. otherwise:
    - If the value is NOT an object, include it
    - If the value is an object, and it is not empty, include it as well

lib/addons/helm-addon/kubectl-provider.ts
starchx: Thanks - added.

docs/addons/argo-cd.md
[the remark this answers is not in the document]
starchx: Agreed. We can test workload repo after this.  - so I will leave them as commented out for now then.

lib/addons/argocd/application.ts
starchx: I added a comment to the code.

Issue #587, opened 2023-01-28
[Karpenter] Deploy fail when more than 2 eks


namjug-kim:
### Describe the bug

If deploy more than 2 eks cluster with carpenter addon, an **"Export with name KarpenterInstanceProfileName is already exported by stack {cdk-eks-blueprints-id}"** error occurs.

### Expected Behavior

Deployment Successful

### Current Behavior

cloudformation stop with error **"Export with name KarpenterInstanceProfileName is already exported by stack {cdk-eks-blueprints-id}"**

### Reproduction Steps

```
const clusterA = blueprints.EksBlueprint.builder()
    .addOns(new blueprints.addons.KarpenterAddOn())
    .build(app, `cluster-a`);

const clusterB = blueprints.EksBlueprint.builder()
    .addOns(new blueprints.addons.KarpenterAddOn())
    .build(app, `cluster-b`);
```

deploy multiple cluster with karpenterAddOn

### CDK CLI Version

2.61.1 (build d319d9c)

### EKS Blueprints Version

1.5.4

### Node.js Version

v19.4.0

### Environment details (OS name and version, etc.)

mac m1

youngjeong46:
@namjug-kim Thank you for submitting the issue yesterday. I have an open PR to address the issue at hand.

youngjeong46:
@namjug-kim The latest npm release (1.6.0) should have this incorporated. Please test and re-open the issue if you still see it linger. Closing it for now.

PR #595, opened 2023-02-02
Addressing multiple EKS role name/instance name issue

youngjeong46:
*Issue #, if available:* #587

*Description of changes:* If deploy more than 2 EKS clusters are deployed in a Stack with Karpenter addons, an "Export with name KarpenterInstanceProfileName is already exported by stack {cdk-eks-blueprints-id}" error occurs.

Minor revision to add random names appended to those role names.


shapirov103 (approved):
LGTM

PR #604, opened 2023-02-28
Adding ability to mark add-ons as control plane (deployed before capacity is added to the cluster)

shapirov103:
Adding ability to mark add-ons as control plane (deployed before capacity is added to the cluster)

*Issue #, if available:*

*Description of changes:*
All core add-ons now have a new `controlPlaneAddOn` attribute. If set to true, will deploy the add-on before any capacity is added to the cluster. 


shapirov103:
@elamaran11 if you merge from this branch, you can deploy VPC CNI to the control plane before any capacity is added. Please let me know if that solves the custom network config issue.

shapirov103:
e2e tests fail on delete. @youngjeong46 I see that there is an NLB provisioned with jupyterhub even thought ingress is set to false, is it by design?

youngjeong46:
> e2e tests fail on delete. @youngjeong46 I see that there is an NLB provisioned with jupyterhub even thought ingress is set to false, is it by design?

The parameter is a bit of a misnomer. Enabling it exposes the front end hub as an ingress controller (using AWS ALB). Not enabling it just exposes the service via Network Load Balancer.

elamaran11 (approved):
LGTM @shapirov103 This worked fine as well along with my VPC CNI addon.

PR #606, opened 2023-03-01
Feature/Jupyterhub: Removed option to expose proxy svc using NLB

youngjeong46:
*Issue #, if available:*

*Description of changes:* If the Ingress option is disabled, the proxy svc will use `ClusterIP` and not `LoadBalancer`, and therefore removing the option to expose using NLB. This will require users to expose the svc using port-forwarding to access via `localhost`.

### DIFF
--- a/lib/addons/jupyterhub/index.ts
+++ b/lib/addons/jupyterhub/index.ts
@@ -176,7 +176,7 @@ export class JupyterHubAddOn extends HelmAddOn {
         }
 
         // Ingress instead of LoadBalancer service to expose the proxy - leverages AWS ALB
-        // If not, then it will leverage AWS NLB
+        // If not, then it will not be exposed
         const enableIngress = this.options.enableIngress || false;
         const ingressHosts = this.options.ingressHosts || [];
         const ingressAnnotations = this.options.ingressAnnotations;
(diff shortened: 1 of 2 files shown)


lib/addons/jupyterhub/index.ts
```diff
@@ -205,13 +205,7 @@ export class JupyterHubAddOn extends HelmAddOn {
...
             assert(!ingressAnnotations, 'Ingress annotations CANNOT be assigned when ingress is disabled');
             assert(!cert, 'Cert option is only supported if ingress is enabled.');
-            setPath(values, "proxy.service", { 
-                "annotations": {
-                    "service.beta.kubernetes.io/aws-load-balancer-type": "nlb",
-                    "service.beta.kubernetes.io/aws-load-balancer-scheme": "internet-facing",
-                    "service.beta.kubernetes.io/aws-load-balancer-nlb-target-type": "ip",
-                }
-            });
+            setPath(values, "proxy.service", {"type": "ClusterIP"});
```
shapirov103, line 208: hard-coded? Why not let customers decide if they need an NLB and modify our e2e test to set this option to false?

shapirov103 (commented):
@youngjeong46 looks good, but let's not hardcode that decision, unless you point me how customers can override it.

shapirov103 (approved):
LGTM

Issue #616, opened 2023-03-07
JupyterHubAddOn: CLUSTERIP requires AwsLoadBalancerControllerAddOn


aliaksei-ivanou:
### Describe the bug

JupyterHubAddOn JupyterHubServiceType.CLUSTERIP has an unnecessary dependency on AwsLoadBalancerControllerAddOn.

### Expected Behavior

Remove AwsLoadBalancerControllerAddOn as a dependency

### Current Behavior

Dependency on AwsLoadBalancerControllerAddOn

### Code:
```
import * as blueprints from "@aws-quickstart/eks-blueprints";
import {
  EksBlueprint,
  GlobalResources
} from "@aws-quickstart/eks-blueprints";
import { RemovalPolicy } from "aws-cdk-lib";
import * as kms from "aws-cdk-lib/aws-kms";
import { Construct } from "constructs";

const gitUrl =
  "https://github.com/aliaksei-ivanou/eks-blueprints-workloads.git";
const targetRevision = "efs-encryption-at-rest";

export default class DataAtRestEncryptionConstruct {
  build(scope: Construct, id: string) {
    const stackId = `${id}-blueprint`;

    const kmsKey = blueprints.getNamedResource(
      blueprints.GlobalResources.KmsKey
    ) as kms.Key;

    EksBlueprint.builder()
      .resourceProvider(GlobalResources.KmsKey, new blueprints.KmsKeyProvider())
      .addOns(
        new blueprints.JupyterHubAddOn({
          serviceType: blueprints.JupyterHubServiceType.CLUSTERIP,
          efsConfig: {
            removalPolicy: RemovalPolicy.DESTROY,
            pvcName: "jupyterhub-efs-pvc",
            capacity: "10Gi"
          },
        }),
      )
      .teams() // add teams here
      .build(scope, stackId);
  }
}
```

### Error:
```
❯ cdk list
DEBUG Core add-on adot is at version v0.66.0-eksbuild.1
DEBUG Core add-on adot is at version v0.66.0-eksbuild.1
Assertion failed: Unable to auto-detect cluster autoscaler version. Applying latest. Provided EKS cluster version: 1.21
DEBUG Core add-on adot is at version v0.66.0-eksbuild.1
Assertion failed: Unable to auto-detect cluster autoscaler version. Applying latest. Provided EKS cluster version: 1.21
DEBUG Core add-on vpc-cni is at version v1.12.2-eksbuild.1
[...]

PR #620, opened 2023-03-08
JupyterHubAddOn: Removing AWS Load Balancer Controller dependency issue for ClusterIP service

youngjeong46:
*Issue #, if available:* #616

*Description of changes:* We are removing AWS Load Balancer Controller Add On dependency when the JupyterHub proxy service is using ClusterIP type (and therefore does not require NLB or ALB).

### DIFF
--- a/lib/addons/jupyterhub/index.ts
+++ b/lib/addons/jupyterhub/index.ts
@@ -1,7 +1,7 @@
 import * as assert from "assert";
 import { Construct } from "constructs";
 import { ClusterInfo } from '../../spi';
-import { createNamespace, dependable, setPath } from '../../utils';
+import { createNamespace, setPath } from '../../utils';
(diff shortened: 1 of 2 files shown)


lib/addons/jupyterhub/index.ts
```diff
@@ -135,7 +135,6 @@ export class JupyterHubAddOn extends HelmAddOn {
         this.options = this.props as JupyterHubAddOnProps;
     }
     
-    @dependable(AwsLoadBalancerControllerAddOn.name)
```
shapirov103, line 138: GitHub action warning: unused import of dependable.

shapirov103 (commented):
looks good, one minor comment.

lib/addons/jupyterhub/index.ts
youngjeong46: addressed.

shapirov103 (approved):
lgtm

Issue #619, opened 2023-03-08
AwsForFluentBitAddOn: Helm chart does not deploy due to unconnected dependencies between namespace, serviceaccount, and helm chart


youngjeong46:
### Describe the bug

When deploying AwsForFluentBitAddOn with all default configurations, we receive a CloudFormation error:

```
Error: b'Release "blueprints-addon-aws-for-fluent-bit" does not exist. Installing it now.\nError: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "blueprints-addon-aws-for-fluent-bit" namespace: "" from "": no matches for kind "ClusterRole" in version "rbac.authorization.k8s.io/v1beta1"\nensure CRDs are installed first, resource mapping not found for name: "blueprints-addon-aws-for-fluent-bit" namespace: "" from "": no matches for kind "ClusterRoleBinding" in version "rbac.authorization.k8s.io/v1beta1"\nensure CRDs are installed first]\n' Logs: /aws/lambda/batch-blueprint-awscdkawseksKubect-Handler886CB40B-iHku6qSuRmVO at invokeUserFunction (/var/task/framework.js:2:6) at processTicksAndRejections (internal/process/task_queues.js:95:5) at async onEvent (/var/task/framework.js:1:365) at async Runtime.handler (/var/task/cfn-response.js:1:1543) (RequestId: 9b9ff205-9229-4981-84d9-db7a267e983f)
```

Upon further inspection, there seems to be no dependencies between namespace, service account, and helm chart that are created as part of the addon.

### Expected Behavior

I expect to deploy the addon successfully.

### Current Behavior

The addon does NOT deploy, with the error produced from CloudFormation as mentioned in the bug report.

### Reproduction Steps

The blueprint that I used to deploy is:

```
import { Construct } from 'constructs';
import * as blueprints from '@aws-quickstart/eks-blueprints'
import { BatchEksTeam } from '@aws-quickstart/eks-blueprints';

export default class BatchOnEKSConstruct {
    build(scope: Construct, id: string, teams: BatchEksTeam[]) {
        
        const stackID = `${id}-blueprint`
        blueprints.EksBlueprint.builder()
            .account(process.env.CDK_DEFAULT_ACCOUNT!)
            .region(process.env.CDK_DEFAULT_REGION!)
            .addOns(
                new blueprints.AwsBatchAddOn(), 
                new blueprints.AwsForFluentBitAddOn({
                    version: '0.1.23',
[...]

PR #621, opened 2023-03-08
AwsForFluentBitAddOn: Bug Fix on Helm Chart dependency

youngjeong46:
*Issue #, if available:* #619

*Description of changes:*
- Added Helm Chart dependency on serviceaccount
- Removed namespace creation (default to `kube-system` which already exists)

### DIFF
--- a/lib/addons/aws-for-fluent-bit/index.ts
+++ b/lib/addons/aws-for-fluent-bit/index.ts
@@ -52,11 +52,6 @@ export class AwsForFluentBitAddOn extends HelmAddOn {
     deploy(clusterInfo: ClusterInfo): Promise<Construct> {
         const cluster = clusterInfo.cluster;
         const namespace = this.options.namespace!;
-        
-        // Create namespace
-        if (this.options.createNamespace) {
-            createNamespace(namespace, cluster, true);
-        }
 
         // Create the FluentBut service account.
         const serviceAccountName = 'aws-for-fluent-bit-sa';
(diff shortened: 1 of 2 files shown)


lib/addons/aws-for-fluent-bit/index.ts
```diff
@@ -45,10 +45,7 @@ export class AwsForFluentBitAddOn extends HelmAddOn {
 
     deploy(clusterInfo: ClusterInfo): Promise<Construct> {
         const cluster = clusterInfo.cluster;
-
-        // Create the FluentBit namespace.
         const namespace = this.options.namespace;
-        createNamespace(this.options.namespace!, cluster, true);
```
shapirov103, line 51: it is going to fail if namespace is custom, not kube-system. let's add a flag to like
```
createNamespace: true | false
```
if true, let's create namespace and add dependencies.

shapirov103 (commented):
Definitely needed fix! 
one comment, also please address github action warning (I assume lint) on used imports.

lib/addons/aws-for-fluent-bit/index.ts
youngjeong46: addressed.

lib/addons/aws-for-fluent-bit/index.ts
```diff
@@ -45,10 +51,12 @@ export class AwsForFluentBitAddOn extends HelmAddOn {
...
         const cluster = clusterInfo.cluster;
-
-        // Create the FluentBit namespace.
-        const namespace = this.options.namespace;
-        createNamespace(this.options.namespace!, cluster, true);
+        const namespace = this.options.namespace!;
+        
+        // Create namespace
+        if (this.options.createNamespace) {
+            createNamespace(namespace, cluster, true);
```
shapirov103, line 58: Missing dependency between the namespace and service account, it will fail intermittently. 
Similar pattern [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/ack/index.ts#L77).

shapirov103 (commented):
Almost there, one more issue.

PR #660, opened 2023-04-25
CDK version change to 2.76.0 from 2.72.0 per 1.7.0 release

youngjeong46:
*Issue #, if available:* 

*Description of changes:*
With 1.7.0 release, CDK version should be documented as 2.76.0 (as was updated in `package.json`). This will revise to reflect changes in all documentations.


shapirov103 (approved):
I cannot image how that happened. thank you for catching, will include in 1.7.1  now.

LGTM

Issue #631, opened 2023-03-26
EKS: tags are not propagated to the cluster


Feder1co5oave:
### Describe the bug

CDK EKS offers the possibility to add tags to the EKS cluster on cluster creation (modification is not yet supported: https://github.com/aws/aws-cdk/issues/19388)
via the `tags` property in the props passed to the `eks.Cluster(scope: Construct, id: string, props: ClusterProps)` constructor.

In the bluprint, the cluster is created in
`GenericClusterProvider.internalCreateCluster()` or
`FargateClusterProvider.internalCreateCluster()`,
which get passed either a 
`GenericClusterProviderProps extends eks.ClusterOptions` or a
`FargateClusterProviderProps extends eks.CommonClusterOptions` props. Neither of those has the `tags` property.

### Expected Behavior

There should be a way to tag the EKS cluster

### Current Behavior

Cluster gets created without any tags.
Even specifying `Tags.of(stack).add('tagKey', 'tagValue')` has no effect, because these tags do not make it into the cluster custom resource `Config`.

### Reproduction Steps

You cannot specify tags for a Blueprint EKS cluster

### Possible Solution

Would it be feasible to add a `tags?: { [key: string]: string }` property to `GenericClusterProviderProps` and 
`FargateClusterProviderProps` ?

### CDK CLI Version

2.66.1

### EKS Blueprints Version

1.6.0

### Node.js Version

v16.19.1

Zetten:
I note it's not yet been triaged, but (entirely selfishly!) I'd suggest fairly high priority. Support for common tags through the stack would be vital for some common business use cases, e.g. for compliance with project/owner/cost allocation tagging requirements.

If I were to deploy an EKS cluster using this I'd have our cloud billing team complaining _very_ quickly that I haven't tagged resources appropriately :smile:

elamaran11:
@Zetten @Feder1co5oave This issue is fixed now. Please check from your end using latest release and let us know if you have any questions. Please close the PR if all is well.

PR #662, opened 2023-04-27
Fix/cluster tags

elamaran11:
*Issue #, if available:* 
#631 

*Description of changes:*

Tags for GenericClusterProvider and FargateClusterProvider.

### DIFF
--- a/docs/addons/grafana-operator.md
+++ b/docs/addons/grafana-operator.md
@@ -1,4 +1,4 @@
-# FluxCD Add-on
+# Grafana-Operator Add-on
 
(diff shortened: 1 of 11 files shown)


lib/cluster-providers/fargate-cluster-provider.ts
```diff
@@ -2,6 +2,7 @@ import { Construct } from "constructs";
 import * as ec2 from "aws-cdk-lib/aws-ec2";
 import * as eks from "aws-cdk-lib/aws-eks";
 import { defaultOptions, GenericClusterProvider } from './generic-cluster-provider';
+import { Tags } from "aws-cdk-lib";
```
shapirov103, line 5: Github actions warning - make lint please

docs/addons/grafana-operator.md
```diff
@@ -1,4 +1,4 @@
-# FluxCD Add-on
```
shapirov103, line 1: good catch 👍

shapirov103 (commented):
@elamaran11 awesome change, only a couple of minor inline comments. 

A bigger question is: it is odd that the change affects only Fargate and Generic cluster providers but not MngClusterProvider or AsgClusterProvider.

I realize that `MngClusterProviderProps` and such already have tags attribute but that was meant for the nodegroup tags, not the cluster. 

If a customer passes tags with either MngClusterProvider or AsgClusterProvider, both node group and cluster will be tagged. This may be counter-intuitive. 

Resolution: 
1. Rename tags clearly and use `clusterTags` vs `nodeGroupTags` to avoid confusion across the board.
2. Use tags consistently across the board, however for Mng and Asg cases document that provided tags will apply both to the nodegroup and the cluster.

shapirov103 (approved):
Looks great, approving. merge pending e2e

PR #674, opened 2023-05-08
Bumping cdk version

rverma-nsl:
Bumping CDK version to make this compatible with projects using latest aws-cdk.


shapirov103:
@rverma-nsl  we set the version fixed as an indication of the version that we tested with. It does not mean that customers cannot use a different version of CDK. You can use `npm` and override constructs to set a  different version of CDK  if you wish as we do [here](https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/package.json#L45).
CDK version can change any day, so by the time we release it may already be different.

shapirov103:
superseded by https://github.com/aws-quickstart/cdk-eks-blueprints/pull/679

PR #681, opened 2023-05-10
VPC Custom CIDR and Flux Kustomization Features

elamaran11:
*Issue #, if available:*

*Description of changes:*


lib/resource-providers/vpc.ts
```diff
@@ -8,11 +8,13 @@ import { ResourceContext, ResourceProvider } from "../spi";
  */
 export class VpcProvider implements ResourceProvider<ec2.IVpc> {
     readonly vpcId?: string;
+    readonly primaryCidr?: string;
```
shapirov103, line 11: This is a risky change for any existing customer. What they used to pass as secondary now all of a sudden becomes primary cidr. I suggest a different approach to prevent such issue:

```
interface VpcProps {
   primaryCidr, 
   secondaryCidr?
   secondarySubnetCidrs?: []
}

class VpcProvider {
    constructor(private id?: string, private vpcProps?: VpcProps)
}
```

shapirov103 (commented):
Looks good, see my comment on VPC refactoring.

Issue #651, opened 2023-04-18
(Add-on) Apache Airflow


youngjeong46:
### Describe the feature

Implementation of Apache Airflow on EKS Blueprints

### Use Case

Apache Airflow is a workflow orchestration tool. This will help build more features for Data on EKS.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.72.0

### Node.js Version

v16.16.0

### Environment details (OS name and version, etc.)

MacOS

youngjeong46:
Feature merged - will close.

Issue #456, opened 2022-07-29
Backstage Addon


blandes:
### Describe the feature

This is a new feature to introduce Backstage Addon to AWS CDK EKS Blueprints.

This Addon deploys Backstage deployment with Aurora/Postgres. https://github.com/vinzscam/backstage-chart

### Use Case

This enables a developer experience for instrumenting your applications running on Amazon EKS. https://backstage.io/

### Proposed Solution

New Backstage Addon that implements a scalable Backstage deployment with Aurora/Postgres.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.33

### EKS Blueprints Version

1.10

### Node.js Version

18.6.0

### Environment details (OS name and version, etc.)

Mac

elamaran11:
This addon is implemented already. Please reachout with a new issue, if you see problems with the addon. Also we will be having a pattern soon in patterns repo;

PR #683, opened 2023-05-12
Database Resource Provider

5herlocked:
*Issue #, if available:* In support of #456; #651

*Description of changes:* Addition of an extensible RDS resource provider - currently only supports Aurora, but will be adding additional support in the coming months as demand changes.

### DIFF
--- a/docs/resource-providers/rds-providers.md
+++ b/docs/resource-providers/rds-providers.md
@@ -1,0 +1,126 @@
+# Relational Database Service Resource Providers
+
+The relational database resource provider specifically tightly couples the lifetime of the RDS
+Resources created with that of the EKS Cluster. It should only be used with a so-called management
+cluster that governs the platform.
+
+To prevent accidental deletion of data while rebuilding EKS clusters, the default retention policy 
+assigned to the RDS instances and clusters is `RetainPolicy.SNAPSHOT` which ensures that while
+the cluster and database is being deleted, it is first backed-up and then deleted.
+
+In the management cluster pattern, this resource provider is like resources provisioned with ACK
+or Crossplane. Removal of the resource from the management cluster by default will drop such
+resources as well.
+
+### CreateRDSInstanceProvider
+Creates an RDS Instance and make it available to the blueprint constructs with the provided name.
+
+This method creates an RDS Instance with the database engine of your choice and in the VPC
+included in the resource context or creates a VPC for you.
+
+The `rdsProps` transparently exposes the underlying RDS Instance properties and will accept and 
+pass them upstream to the RDS instance method creating the database.
+
+Example Implementation without providing a custom VPC:
+```typescript
+const stack = blueprints.EksBlueprint.builder()
+    .resourceProvider(
+        GlobalResources.Rds,
+        new CreateRDSProvider(
+            {
+                rdsProps: {
+                    credentials: Credentials.fromGeneratedSecret('admin'),
+                    engine: DatabaseInstanceEngine.mariaDb({
+                        version: MariaDbEngineVersion.VER_10_3
+                    })
+                },
+                name: "rds-instance-no-vpc"
+            }
+        )
+    )
+    .account("123456789")
+    .region("us-east-1")
+    .build(app, 'rds-instance-no-vpc');
+```
+
+Example Implementation while providing a custom VPC:
+```typescript
+const stack = blueprints.EksBlueprint.builder()
+    .resourceProvider(
+        GlobalResources.Vpc,
+        new VpcProvider(
+            undefined,
+            {
+                primaryCidr: "10.0.0.0/16"
+            },
+        )
+    )
+    .resourceProvider(
+        GlobalResources.Rds,
+        new CreateRDSProvider({
+            rdsProps: {
+                credentials: Credentials.fromGeneratedSecret('admin'),
+                engine: DatabaseInstanceEngine.postgres({
+                    version: PostgresEngineVersion.VER_15_2
+                })
+            },
+            name: 'rds-instance-w-vpc'
+        })
+    )
+    .account("1234567889")
+    .region("us-east-1")
+    .build(app, 'rds-instance-w-vpc');
+
+```
+
+### CreateAuroraClusterProvider
+Creates an RDS Aurora Cluster and makes it available to the blueprint constructs with the provided name.
+
+This method creates an RDS Cluster with the database engine of your choice and in the VPC
+included in the resource context or creates a VPC for you.
+
+The `rdsProps` transparently exposes the underlying RDS Cluster properties and will accept and
+pass them upstream to the RDS cluster method creating the database.
+We recommend using either Aurora Serverless or creating specific reader and writer instances.
+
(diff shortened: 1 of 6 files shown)


youngjeong46:
@5herlocked would love to see this extend to RDS as an addon I'm working on (Apache Airflow) needs RDS for DB.

5herlocked:
Will start working on it this week

5herlocked:
As requested, RDS Instance provider has been built.

Will be adding in test cases promptly.

5herlocked:
@youngjeong46 Give this a try.

It does successfully create an RDS and provides a secret as an ISecret.
Should be pretty straight forward to integrate it into a full deployment.

5herlocked:
@shapirov103 @youngjeong46 Please leave feedback if considered required.
Will fillout additional documentation towards EoW.

5herlocked:
Absolutely, we can use crossplane or ACK. Though it would require a fair bit of setup to ensure the RDS instance/cluster's lifecycle would match that of the EKS cluster.

Right now if I understand it right, the process to setup an RDS cluster/instance supporting an EKS cluster is one of the following:
* EKS Blueprints -> ACK -> Argo Managed RDS YAMLs
* EKS Blueprints -> ACK -> Pre-Existing RDS YAMLs (probably still managed through argo)

With this, we can bypass ACK and directly attach the Lifecycle of a Blueprinted cluster to an RDS instance supporting to make the flow:
* EKS Blueprints -> RDS with the same configurations that you get with the CDK resources `DatabaseInstance`and `DatabaseCluster`

I believe this more closely matches with the spirit of the repository than using ACK, as this considers the RDS instance as a part of the infrastructure required by the cluster and provides the ability to create and delete it as part of the infra pipeline.

elamaran11:
@5herlocked First of all, thank-you for the PR and also response. Having RDS and managing its lifecycle strictly as part of Cluster creation is strictly a No-No. Lifecycle of RDS should be ideally maintained outside of the lifecycle of the EKS cluster creation so even if the cluster is deleted, RDS remains and data persisted in RDS remains. For S3 its a different usecase where you might need to create a bucket for temporary purposes like Airflow usecase. But in case of RDS its a strict anti pattern to manage the RDS creation with EKS Cluster creation. The reason why i suggested to use ACK or CrossPlane is for multi cluster management scenarios where you have a management cluster which creates child cluster and RDS databases. So bottom like is RDSProvider does not align to the spirit of ResourceProvider so i would question on why we needs this, what is your motivation and usecase if the answer is RDS creation with cluster, i wouldn't necessarily agree. Using ACK or CrossPlane with Argo or Flux promotes GitOps which would be a right approach in your case.

shapirov103:
@elamaran11 and myself discussed it. It should be made very clear in the documentation that this addon is coupled to the lifecycle of the cluster (which is the case for any resource tbh). It can only be used with a so-called management cluster that governs the platform. 
It can also be used for illustrative purposes, however it makes sense to check resource retention control to avoid accidental data loss when stack is reprovisioned or dropped. 
In the management cluster pattern, this resource provider is not that different from any resources provisioned with ACK or Crossplane. Removal of the resource from the management cluster by default will drop such resources as well. 
I can see Ela's perspective as more generalizing  - e.g. what if we now add a bunch of different database providers and then customers may misuse it. We need to define an overall strategy for resource providers with the goal of addressing real customer use cases and improving customer experience the best we can.

5herlocked:
@shapirov103 @elamaran11 Will it make sense to, by default set the retention policy to snapshot when deleting as a way of safeguarding customer data?

shapirov103:
@5herlocked let's resolve the conflicts and I agree with the direction of applying a snapshot or any more conservative retention policy to avoid inadvertent data loss. 
We will need a doc under docs/resource-providers (index.md and the rds.md).

shapirov103 (commented):
Please see my comment on the issue on retention and docs.

docs/resource-providers/rds-providers.md
```diff
@@ -0,0 +1,126 @@
+# Relational Database Service Resource Providers
+
+The relational database resource provider specifically tightly couples the lifetime of the RDS
```
shapirov103, line 3: remove 'specifically'

shapirov103: reword "with that of the EKS Cluster"  - sounds confusing to me

docs/resource-providers/rds-providers.md
```diff
@@ -0,0 +1,126 @@
...
+Resources created with that of the EKS Cluster. It should only be used with a so-called management
```
shapirov103, line 4: Let's say that "production usage should be using the centralized/management cluster pattern.

docs/resource-providers/rds-providers.md
```diff
@@ -0,0 +1,126 @@
...
+### CreateRDSInstanceProvider
+Creates an RDS Instance and make it available to the blueprint constructs with the provided name.
+
+This method creates an RDS Instance with the database engine of your choice and in the VPC
+included in the resource context or creates a VPC for you.
+
+The `rdsProps` transparently exposes the underlying RDS Instance properties and will accept and 
+pass them upstream to the RDS instance method creating the database.
+
+Example Implementation without providing a custom VPC:
```
shapirov103, line 24: minor: why implementation is upper case? applies elsewhere

shapirov103 (commented):
@5herlocked great work! a few minor comments on the doc

5herlocked:
Yes, it was as part of the last commit. Please feel free to look it over.

5herlocked:
@elamaran11 feel free to close/merge. It is complete and the tests are inplace.

elamaran11 (commented):
LGTM. @shapirov103 Please check from your end.

sanyer:
Hey @5herlocked how would this work in terms of network connectivity? I don't any security group defined or imported here.

shapirov103 (commented):
@5herlocked have you validated functionally on a cluster? I see this is not included in the e2e test as it stands. If you can attach the blueprint you used to validate and add a screenshot of a command (kubectl or aws cli call) that validates it wokrs, that would help merge it. 

Please also see the question wrt to network connectivity. How do you envision that part - should there be an extension for SG configuration? we can defer it, but it seems like a common requirement from the customers

5herlocked:
Networking configuration is entirely inherited from  the [VPC Resource Provider](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/resource-providers/vpc.ts). 

Additional configurations are can also be transparently passed to the underlying resources, note the interface `CreateRDSInstanceProviderProps.rdsProps`.

Will include deployment screenshots and necessary e2e tests ASAP.

shapirov103:
> @shapirov103 Will include deployment screenshots and necessary e2e tests ASAP.

@5herlocked whenever you get a chance, lets address the above and merge it. The functionality looks really good, I assume there will be more requests around this capability once customers start using it.

5herlocked:
Underlying Database Cluster CDK constructs changed. This is no longer functional.

shapirov103:
Closing for now, we can reopen if needed.

PR #690, opened 2023-05-18
Flux Kustomization Bootstrap Values Redesign and Amp Workspace Redesign

elamaran11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/addons/amp-addon.md
+++ b/docs/addons/amp-addon.md
@@ -34,14 +34,15 @@ Pattern # 2 : Overriding property values for Name and Tags for a custom AMP Work
 
 ```typescript
 import * as cdk from 'aws-cdk-lib';
+import { CfnWorkspace } from 'aws-cdk-lib/aws-aps';
 import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
 const ampWorkspaceName = "blueprints-amp-workspace";
-const ampPrometheusEndpoint = (blueprints.getNamedResource(ampWorkspaceName) as unknown as amp.CfnWorkspace).attrPrometheusEndpoint;
+const ampWorkspace: CfnWorkspace = blueprints.getNamedResource(ampWorkspaceName);
 
 const addOn = new blueprints.addons.AmpAddOn({
-    ampPrometheusEndpoint: ampPrometheusEndpoint,
+    ampPrometheusEndpoint: ampWorkspace.attrPrometheusEndpoint,
     deploymentMode: DeploymentMode.DEPLOYMENT,
     namespace: 'default',
     name: 'adot-collector-amp'
@@ -91,14 +92,15 @@ Pattern # 4 : Overriding property values for different deployment Modes. This pa
 
 ```typescript
 import * as cdk from 'aws-cdk-lib';
+import { CfnWorkspace } from 'aws-cdk-lib/aws-aps';
 import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
 const ampWorkspaceName = "blueprints-amp-workspace";
-const ampPrometheusEndpoint = (blueprints.getNamedResource(ampWorkspaceName) as unknown as amp.CfnWorkspace).attrPrometheusEndpoint;
+const ampWorkspace: CfnWorkspace = blueprints.getNamedResource(ampWorkspaceName);
 
 const addOn = new blueprints.addons.AmpAddOn({
-    ampPrometheusEndpoint: ampPrometheusEndpoint,
+    ampPrometheusEndpoint: ampWorkspace.attrPrometheusEndpoint,
     deploymentMode: DeploymentMode.DAEMONSET,
     namespace: 'default',
     name: 'adot-collector-amp'
--- a/docs/addons/amp-addon.md
+++ b/docs/addons/amp-addon.md
@@ -37,21 +37,11 @@ import * as cdk from 'aws-cdk-lib';
 import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
+const ampWorkspaceName = "blueprints-amp-workspace";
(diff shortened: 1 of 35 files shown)


shapirov103 (commented):
@elamaran11 I only reviewed high level.
About using resource providers and the referencing the elsewhere in the blueprint:

Example of having a correct resource provider [here](https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/lib/security/data-at-rest-encryption/index.ts#L25).  
Example of using a property from the resource provider on some other place, like passing it to an addon [here](https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/lib/security/data-at-rest-encryption/index.ts#L78).

elamaran11:
@shapirov103 PR is ready with all issues fixed.

examples/blueprint-construct/index.ts
```diff
@@ -149,6 +155,17 @@ export default class BlueprintConstruct {
             new blueprints.AwsBatchAddOn(),
             new blueprints.AwsForFluentBitAddOn(),
             new blueprints.FluxCDAddOn(),
+            // new blueprints.FluxCDAddOn({
```
shapirov103, line 158: please remove the commented content, however I think it would quite handy for testing if you moved it to the examples file here: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/examples.ts

That one contains examples all of which deploy to default VPC to avoid address limits. you can run them with `npm run example list|deploy`

docs/addons/amp-addon.md
```diff
@@ -88,30 +94,35 @@ import * as cdk from 'aws-cdk-lib';
 import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
+const ampWorkspaceName = "blueprints-amp-workspace";
+const ampPrometheusEndpoint = (blueprints.getNamedResource(ampWorkspaceName) as unknown as amp.CfnWorkspace).attrPrometheusEndpoint;
```
shapirov103, line 98: This should be changed at least in the docs. 

I suggest this syntax:

```typescript
const ampWorkspace: CfnWorkspace = blueprints.getNamedResource(ampWorkspaceName);

const addOn = new blueprints.addons.AmpAddOn({
    ampPrometheusEndpoint: ampWorkspace.attrPrometheusEndpoint,
 ..
```

I saw that you struggled with the data types. Please pull from my feature/cfn-resource-providers branch first, it has a change that will make it happen.

examples/blueprint-construct/index.ts
```diff
@@ -41,14 +42,19 @@ export default class BlueprintConstruct {
             iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonSSMManagedInstanceCore")
         ]);
 
+        const ampWorkspaceName = "blueprints-amp-workspace";
+        const ampPrometheusEndpoint = (blueprints.getNamedResource(ampWorkspaceName) as unknown as amp.CfnWorkspace).attrPrometheusEndpoint;
```
shapirov103, line 46: Please see my comments in the docs on the style of referencing an attribute from a resource, provided by a resource provider

lib/addons/amp/index.ts
```diff
@@ -60,7 +48,7 @@ export const enum DeploymentMode {
  * Defaults options for the add-on
  */
 const defaultProps = {
-    workspaceName: 'blueprints-amp-workspace',
+    ampPrometheusEndpoint: "https://aps-workspaces.us-east-1.amazonaws.com/workspaces/ws-sample/",
```
shapirov103, line 51: why is this endpoint here hardcoded?

lib/addons/fluxcd/index.ts
```diff
@@ -62,7 +86,7 @@ const defaultProps: HelmAddOnProps & FluxCDAddOnProps = {
   fluxSyncInterval: "5m0s",
   fluxTargetNamespace: "default",
   fluxPrune: true,
-  fluxTimeout: "1m"
+  fluxTimeout: "1m",
```
shapirov103, line 89: remove the last ","  - i know ts does not complain, it hurts engineers across the world

shapirov103 (commented):
Looking very good. Added a few comments, please review.

lib/addons/amp/index.ts
elamaran11: This is to show a sample. See the end it has `ws-sample`.

shapirov103: remove it please from default values.

lib/addons/fluxcd/index.ts
elamaran11: Fixed.

examples/blueprint-construct/index.ts
elamaran11: fixed,

docs/addons/amp-addon.md
elamaran11: Fixed. great suggestion and also on the cfn PR.

examples/blueprint-construct/index.ts
elamaran11: Done.

lib/addons/amp/index.ts
elamaran11: We cant remove that parameter. `ampPrometheusEndpoint` is a mandatory parameter.

lib/addons/amp/index.ts
```diff
@@ -60,7 +48,7 @@ export const enum DeploymentMode {
  * Defaults options for the add-on
  */
 const defaultProps = {
-    workspaceName: 'blueprints-amp-workspace',
+    ampPrometheusEndpoint: "ws-sample-workspace",
```
shapirov103, line 51: Why do we need this attribute here? In the past it was needed to create the workspace, but that approach change.  Is it one of those cases, when customers should really pass the endpoint to the addon to make it work?

shapirov103 (commented):
Looks great, one comment. Almost there.

lib/addons/amp/index.ts
elamaran11: Fixed.

shapirov103 (approved):
LGTM

PR #703, opened 2023-05-26
Apache Airflow update: Removed RDS for now, added doc and refactored codes

youngjeong46:
*Issue #, if available:* #651

*Description of changes:*
- Refactoring code to include full custom helm values for S3, ALB and EFS implementations
- Commented out RDS implementation (as we await for RDS Provider PR to merge)
- Added documentations.

### DIFF
--- a/lib/addons/apache-airflow/index.ts
+++ b/lib/addons/apache-airflow/index.ts
@@ -1,0 +1,302 @@
+import { Construct } from "constructs";
+import * as assert from "assert";
+
+import { ICertificate } from 'aws-cdk-lib/aws-certificatemanager';
+import { IBucket } from 'aws-cdk-lib/aws-s3';
+import { KubernetesManifest } from 'aws-cdk-lib/aws-eks';
+import { PolicyDocument } from 'aws-cdk-lib/aws-iam';
+
+import { HelmAddOnProps, HelmAddOnUserProps } from "../helm-addon";
+import { HelmAddOn } from '../helm-addon/index';
+import { AwsLoadBalancerControllerAddOn } from "../aws-loadbalancer-controller";
+import { EfsCsiDriverAddOn } from "../efs-csi-driver";
+
+import { ClusterInfo } from '../../spi/types';
+import { Values } from "../../spi";
+import { setPath, createNamespace, createServiceAccount } from "../../utils";
+import { IFileSystem } from "aws-cdk-lib/aws-efs";
+
+/**
+ * User provided options for the Helm Chart
+ */
+export interface AirflowAddOnProps extends HelmAddOnUserProps {
+    /**
+     * Enable Load Balancer
+     */
+    enableAlb?: boolean,
+
+    /**
+     * Name of the {@link certificateResourceName} to be used for certificate look up. 
+     * @see {@link ImportCertificateProvider} and {@link CreateCertificateProvider} for examples of certificate providers.
+     */
+    certificateResourceName?: string,
+
+    /**
+     * Names of the S3 Bucket provider named resources (@see CreateS3BucketProvider, @see ImportS3BucketProvider).
+     * S3 Bucket provider is registered as named resource providers with the EksBlueprintProps.
+     */
+    s3Bucket?: string,
+
+    /**
+     * Enable EFS for persistent storage of DAGs
+     */
+    enableEfs?: boolean,
+
+    /**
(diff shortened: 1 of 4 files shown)


lib/addons/apache-airflow/index.ts
```diff
@@ -0,0 +1,302 @@
...
+    constructor(props?: AirflowAddOnProps) {
+        super({...defaultProps, ...props});
+        this.options = this.props as AirflowAddOnProps;
+    }
+
+    deploy(clusterInfo: ClusterInfo): Promise<Construct> {
+        const cluster = clusterInfo.cluster;
+
+        // Create Namespace
+        const ns = createNamespace(this.options.namespace!, cluster, true, true);
```
shapirov103, line 105: is there a case when they want to install into an existing namespace? we can call out in the documentation. Generally, we also add a `createNamespace` attribute

lib/addons/apache-airflow/index.ts
```diff
@@ -0,0 +1,302 @@
...
+
+        return Promise.resolve(chart);
+    }
+}
+
+/**
+ * populateValues populates the appropriate values used to customize the Helm chart
+ * @param helmOptions User provided values to customize the chart
+ */
+function populateValues(clusterInfo: ClusterInfo, ns: KubernetesManifest, helmOptions: AirflowAddOnProps): Values {
```
shapirov103, line 119: Our general approach on reconciling the values passed to the helm chart by the user explicitly and values that we set ourselves is to only set the values for keys that are not supplied by the customers - customer explicit settings takes precedence.  it is generally achieved by merging onto the customer values like [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/appmesh/index.ts#L101)

lib/addons/apache-airflow/index.ts
```diff
@@ -0,0 +1,302 @@
...
+        name: `${sa.serviceAccountName}`
+    });
+    
+    return values;
+}
+
+/**
+ * TODO: Helper function to set up the Metadata DB with RDS
+*/
+// function setUpDatabase(values: Values, dbConfig: {username: string; password: string; dbName: string;} | undefined): Values {
```
shapirov103, line 282: Drop the comments

shapirov103 (commented):
@youngjeong46 looks great, a few comments for you. 

Testing - what are we doing for it? Do you want to add it to the e2e test?

lib/addons/apache-airflow/index.ts
youngjeong46: addressed

lib/addons/apache-airflow/index.ts
```diff
@@ -0,0 +1,302 @@
...
+    // dbConfig?: {
+    //     username: string,
+    //     password: string,
+    //     dbName: string 
+    // }
+    
+    /**
+     * Enable Logging with S3
+     */
+    enableLogging?: boolean,
```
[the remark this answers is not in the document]
youngjeong46: addressed.

lib/addons/apache-airflow/index.ts
```diff
@@ -0,0 +1,302 @@
...
+    enableEfs?: boolean,
+
+    /**
+     * Names of the EFS File System provider named resources (@see CreateEfsFileSystemProvider, @see LookupEfsFileSystemProvider).
+     * EFS File System provider is registered as named resource providers with the EksBlueprintProps.
+     * This is required if EFS is enabled
+     */
+    efsFileSystem?: string,
+
+    /**
```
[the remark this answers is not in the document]
youngjeong46: addressed.

youngjeong46:
> @youngjeong46 looks great, a few comments for you.
> 
> Testing - what are we doing for it? Do you want to add it to the e2e test?

I have added e2e. Still testing a few other stuff, so will address rest of the comments.

shapirov103:
> I have added e2e. Still testing a few other stuff, so will address rest of the comments.
Thank you @youngjeong46 , I will release tonight after e2e.

lib/addons/apache-airflow/index.ts
youngjeong46: This may need to be addressed in the next round of PR - there's a bit of refactoring needed here.

shapirov103: Sounds good. I am going to release now, please continue on this.

shapirov103 (approved):
LGTM

Issue #632, opened 2023-03-27
Can you run this cluster with a single NAT gateway for development clusters?


danielloader:
### Describe the documentation issue

The documentation has no information on running a cluster in a single NAT gateway mode for cross AZ access, is this possible?

### Links

https://aws-quickstart.github.io/cdk-eks-blueprints/

shapirov103:
@danielloader we included this in the docs. Please see "Example implementations" [here](https://aws-quickstart.github.io/cdk-eks-blueprints/resource-providers/#contracts). 

I am closing it now, please reopen in case of any issues.

PR #716, opened 2023-06-06
docs: add clarity for creating own VPC resource provider

zjaco13:
*Issue #, if available:* #632

*Description of changes:*
Added documentation to explain creating VPC Resource Providers and added links to other resource providers

### DIFF
--- a/docs/addons/cluster-autoscaler.md
+++ b/docs/addons/cluster-autoscaler.md
@@ -14,7 +14,7 @@ import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
 
-const addOn = new blueprints.addons.ClusterAutoscalerAddOn();
+const addOn = new blueprints.addons.ClusterAutoScalerAddOn();
 
 const blueprint = blueprints.EksBlueprint.builder()
   .addOns(addOn)
--- a/docs/resource-providers/index.md
+++ b/docs/resource-providers/index.md
@@ -40,7 +40,22 @@ Example implementations:
 class VpcResourceProvider implements ResourceProvider<IVpc> {
     provide(context: ResourceContext): IVpc {
         const scope = context.scope; // stack
-        ...
+        return new Vpc(scope, '<vpc-name>', {
+            availabilityZones: ['us-east-1a', 'us-east-1b', 'us-east-1c'], // VPC spans all AZs
+            subnetConfiguration: [{
+                cidrMask: 24,
+                name: 'private',
+                subnetType: SubnetType.PRIVATE_WITH_EGRESS
+            }, {
+                cidrMask: 24,
+                name: 'public',
+                subnetType: SubnetType.PUBLIC
+            }],
+            natGatewaySubnets: {
+                availabilityZones: ['us-east-1b'] // NAT gateway only in 1 AZ 
+                subnetType: SubnetType.PUBLIC
+            }
+        });
     }
 }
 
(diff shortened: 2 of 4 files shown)


docs/resource-providers/index.md
```diff
@@ -40,7 +40,22 @@ Example implementations:
 class VpcResourceProvider implements ResourceProvider<IVpc> {
     provide(context: ResourceContext): IVpc {
         const scope = context.scope; // stack
-        ...
```
elamaran11, line 43: I would rather add this is as a separate example that over writing on existing one.

docs/addons/cluster-autoscaler.md
```diff
@@ -14,7 +14,7 @@ import * as blueprints from '@aws-quickstart/eks-blueprints';
 
 const app = new cdk.App();
 
-const addOn = new blueprints.addons.ClusterAutoscalerAddOn();
+const addOn = new blueprints.addons.ClusterAutoScalerAddOn();
```
elamaran11, line 17: What changed here?

elamaran11 (commented):
@zjaco13 Thanks for the PR, have few minor comments.

docs/addons/cluster-autoscaler.md
zjaco13: changed the s in ClusterAutoscalerAddOn to uppercase, since that's how you have to call the addon

docs/resource-providers/index.md
```diff
@@ -50,6 +50,31 @@ class DynamoDbTableResourceProvider implements ResourceProvider<ITable> {
     }
 }
 
+/**
+ * Example implementation of a VPC Provider that creates a NAT Gateway 
+ * which is available in all 3 AZs of the VPC while only being in one
+ */
+class OtherVpcResourceProvider implements ResourceProvider<IVpc> {
```
shapirov103, line 57: @zjaco13 this is the framework repo, it is different from examples repo. That means that code here is not an example and must configurable, reusable, etc. 

It also must be tested, we have unit tests and e2e tests. 

There are a couple of ways to approach this task:

1. Move the example resource provider (Let's call CustomVpcResourceProvider) to the patterns dir. It can stay as is but we should still use it somewhere in a pattern to validate. 
2. Move this code to documentation as example code.
3. Improve the existing VPC resource provider to become more configurable, such as accept subnet configuration, availability zones. 

@elamaran11 I am a bit concerned with this example as it won't register the provisioned subnets to be reused with vpc-cni. So if this provider is used customers won't be able to use secondary ENIs. Thoughts?

shapirov103 (commented):
@zjaco13 please see my comment and let me know if you have any questions. @elamaran11 please take a look at my comment as well.

docs/resource-providers/index.md
zjaco13: Currently this is in the resources providers documentation, is there a better place in the docs to put it?

zjaco13: I do think it should be more reusable, but I'm struggling to figure out the best way to go about it.  Would it make sense to define a construct and properties that include these VPC properties?

shapirov103: All looks good. I must have assume it was code that you added to the repo not docs.

shapirov103 (approved):
LGTM. @elamaran11  please add review and merge if all good.

elamaran11 (approved):
LGTM

PR #719, opened 2023-06-09
[Apache Airflow addon] Dependency added for EFS

youngjeong46:
*Issue #, if available:*

*Description of changes:* Dependency for EFS Kubernetes resources (storageclass and PVC) added so that they are provisioned after namespace creation.

### DIFF
--- a/lib/addons/apache-airflow/index.ts
+++ b/lib/addons/apache-airflow/index.ts
@@ -16,6 +16,8 @@ import { Values } from "../../spi";
 import { setPath, createNamespace, createServiceAccount } from "../../utils";
 import { IFileSystem } from "aws-cdk-lib/aws-efs";
 
+import merge from "ts-deepmerge";
+
 /**
  * User provided options for the Helm Chart
  */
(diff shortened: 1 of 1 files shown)


lib/addons/apache-airflow/index.ts
```diff
@@ -217,6 +217,8 @@ function populateValues(clusterInfo: ClusterInfo, ns: KubernetesManifest, helmOp
             overwrite: true,
         });
 
+        efsResources.node.addDependency(ns);
```
shapirov103, line 220: efs storage class is not a namespace resource - created at the global. So adding dependency to the namespace is irrelevant for it but it is revelant for the PVC. 

The error that we saw was: either chart or PVC or both are created before the storage class is created. 

We need the following dependencies:

efs storage class depends efs csi driver - it is a global resource so dependencies on ns, pvc, chart
PVC depends on storage class and ns
chart depends on PVC/storage class

shapirov103 (commented):
@youngjeong46 please see my comment

lib/addons/apache-airflow/index.ts
youngjeong46: Am I interpreting this correctly:

For the chart, we would only need to add Dependency for the PVC - if I implement PVC node dependency on sc and ns, by transitivity, sc and ns would provision before the chart.

shapirov103: Yeah, that seems right. The storage class may require dependency on the EFS CSI driver most likely. I am curious if CSI driver dependency is critical for this case, this will have to be tested.

youngjeong46: Is there a way to add CSI driver as dependency? addons are Construct promises, which do not use IDependable interface.

shapirov103: I am a bit confused with this question since we had this conversation in the past, I attribute it to the case of Mondays. I believe [this](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/c5ad95777e7db129916f03c18b9a07f8be0be3c4/lib/addons/jupyterhub/index.ts#L251) is your code based on the discussion we had.

youngjeong46: Case of the Mondays it is.

shapirov103 (dismissed):
Love it, awesome stuff! Will run e2e, hope everything checks out.

PR #734, opened 2023-06-16
Upgrading XRAY Addon

elamaran11:
*Issue #, if available:*

*Description of changes:*


docs/addons/xray.md
```diff
@@ -22,11 +22,11 @@ const blueprint = blueprints.EksBlueprint.builder()
...
+Once deployed, it allows applications to be instrumented with X-Ray by leveraging the X-Ray SDK.  Examples of such integration can be found on [AWS Docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/deploy_servicelens_CloudWatch_agent_deploy_EKS.html).
 
 ## Functionality
 
-1. Creates the `xray-system` namespace.
-2. Deploys the [`xray-daemon`](https://www.eksworkshop.com/intermediate/245_x-ray/x-ray-daemon/) manifests into the cluster.
-3. Configures Kubernetes service account with IRSA (`AWSXRayDaemonWriteAccess`) for communication between the cluster and the AWS X-Ray service 
+1. Creates the `amazon-cloudwatch` namespace.
+2. Deploys the [`xray-daemon`](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/deploy_servicelens_CloudWatch_agent_deploy_EKS.html) manifests into the cluster.
+3. Configures Node role with policies (`CloudWatchAgentServerPolicy`, `AWSXRayDaemonWriteAccess`) for communication between the cluster and the CloudWatch and X-Ray service.
```
shapirov103, line 31: It looks like this setup deploys CW agent and XRay into the cluster. Why would we do that? Can we drop the agent deployment and just deploy XRay? I also assume that CloudWatchAgentServerPolicy won't be needed if we only deploy xray.

shapirov103 (commented):
Looks great, I added one comment.

lib/addons/xray/xray-ds.yaml
```diff
@@ -11,111 +11,72 @@
...
+kind: Namespace
 metadata:
-  name: xray-daemon
+  name: amazon-cloudwatch
   labels:
-    app: xray-daemon
+    name: amazon-cloudwatch
+---
+# create role binding for XRay SDK to read config map
+apiVersion: rbac.authorization.k8s.io/v1
```
shapirov103, line 23: Why do we need these roles and role binding if the SA is not even referenced in the ds deployment?

shapirov103 (commented):
@elamaran11 getting there, I have one comment.

lib/addons/xray/xray-ds.yaml
elamaran11: Makes sense. Removed it. Pleae check now.

shapirov103:
Let's wait for the e2e to pass

PR #735, opened 2023-06-20
Upgrading support for SSM for External Secrets

elamaran11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/addons/external-secrets/index.ts
+++ b/lib/addons/external-secrets/index.ts
@@ -40,6 +40,12 @@ const defaultIamPolicy: iam.PolicyStatement = new iam.PolicyStatement({
     "secretsmanager:DescribeSecret",
     "secretsmanager:ListSecretVersionIds",
     "secretsmanager:ListSecrets",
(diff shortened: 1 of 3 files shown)


lib/addons/external-secrets/index.ts
```diff
@@ -40,6 +40,12 @@ const defaultIamPolicy: iam.PolicyStatement = new iam.PolicyStatement({
     "secretsmanager:DescribeSecret",
     "secretsmanager:ListSecretVersionIds",
     "secretsmanager:ListSecrets",
+    "ssm:DescribeParameters",
+    "ssm:GetParameter",
+    "ssm:GetParameters",
+    "ssm:GetParametersByPath",
+    "ssm:GetParameterHistory",
+    "kms:Decrypt"
```
shapirov103, line 48: At some point we will need to review all such policies for wildcard resource and provide an approach to lock it down to specific resources only, like we did in the EBS csi driver.

shapirov103 (approved):
LGTM

PR #739, opened 2023-06-21
ARM64 Support for External DNS

elamaran11:
*Issue #, if available:*

*Description of changes:*

ARM64 Support for External DNS


lib/addons/external-dns/index.ts
```diff
@@ -22,9 +22,9 @@ const defaultProps = {
     name: 'external-dns',
     chart: 'external-dns',
     namespace: 'external-dns',
-    repository: 'https://charts.bitnami.com/bitnami',
+    repository: 'https://kubernetes-sigs.github.io/external-dns/',
     release: 'blueprints-addon-external-dns',
-    version: '6.20.3',
+    version: '1.12.0',
```
youngjeong46, line 27: fyi the latest version is at 1.13. Not sure if that one's ARM compatible.

youngjeong46 (commented):
Left one comment. Otherwise LGTM.

lib/addons/external-dns/index.ts
shapirov103, line 27: is there a warning on deploy when using 1.12? There should be. 
Has it been confirmed to work with arm?

shapirov103 (commented):
LGTM , however let's move to the latest version if possible. I will kick off e2e in the mean time

lib/addons/external-dns/index.ts
elamaran11: ya nice catch, fixed. It works now with `1.13.0`

elamaran11: @shapirov103 It did throw a warning which i ignored. I fixed it now and it works perfectly fine with ARM. Please run e2e and merge.

shapirov103 (approved):
LGTM

Issue #747, opened 2023-06-27
Multi version support to Kubeproxy Addon.


elamaran11:
### Describe the feature

Multi version support to Kubeproxy Addon like how ALB Addon works. The addon should use version via mapping based on K8s version.

### Use Case

Multi version support to Kubeproxy Addon like how ALB Addon works. The addon should use version via mapping based on K8s version.

### Proposed Solution

Multi version support to Kubeproxy Addon like how ALB Addon works. The addon should use version via mapping based on K8s version.

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.83.0

### EKS Blueprints Version

1.9.2

### Node.js Version

20.0.0

### Environment details (OS name and version, etc.)

Mac

shapirov103:
One thing to make sure here: we cannot always catch up with the latest EKS version, there is a lag between EKS release and CDK update to include that version in the `KubernetesVersion` enum. 
So if the version of the target EKS cluster is not in the map, then the latest should be applied.

PR #750, opened 2023-06-29
update kube-proxy to auto choose version

zjaco13:
*Issue #, if available:* #747

*Description of changes:*
Added version map for kube-proxy addon


lib/addons/kube-proxy/index.ts
```diff
@@ -1,15 +1,54 @@
+import { CfnAddon, KubernetesVersion } from "aws-cdk-lib/aws-eks";
+import { Construct } from "constructs";
+import { ClusterInfo } from "../../spi";
+import { logger } from "../../utils";
 import { CoreAddOn } from "../core-addon";
 
+const versionMap: Map<KubernetesVersion, string> = new Map([
+    [KubernetesVersion.of("1.27"), "v1.27.1-eksbuild.1"],
```
elamaran11, line 8: Why is this line different?

elamaran11 (changes requested):
@zjaco13 Have a minor comment. Also the doc is not updated.
@shapirov103 Is this approach good. CA uses helm approach and coreaddon props is readonly.

lib/addons/kube-proxy/index.ts
zjaco13: The CDK team hasn't added Kubernetes version 1.27 to their enum list yet

lib/addons/kube-proxy/index.ts
```diff
@@ -1,15 +1,54 @@
...
         super({
             addOnName: "kube-proxy",
-            version: version ?? "v1.25.6-eksbuild.1",
+            version: version ?? "auto",
             saName: "kube-proxy"
         });
+        this.version = version ?? "auto";
+    }
+
+    deploy(clusterInfo: ClusterInfo): Promise<Construct> {
```
shapirov103, line 31: Hey Zach, while this implementation works, it almost nullifies the inheritance  - there is very little reason to derive this from a core addon. 

I think we what we should do in this case is to enable this type of version selection for any core addon. 

That means that the core addon should be able to accept optionally something like a versionProviderFunction that takes cluster info and spits out the version. 

The kubeproxy will supply its function that will use the map. However, the default function will to just return whatever customers passed. 

In this way, other core addons, the number of which will grow, will be able to address this aspect without sacrificing the central implementation component of such addons.

shapirov103 (commented):
@zjaco13 It is a great start, I like the map. Please see my comment on the design. Happy to discuss if needed.

lib/addons/kube-proxy/index.ts
zjaco13: Yep, that approach makes a ton of sense.  We should discuss implementation.

lib/addons/core-addon/index.ts
```diff
@@ -126,4 +139,15 @@ export class CoreAddOn implements ClusterAddOn {
         return result;
     }
 
-}
\ No newline at end of file
+    provideVersion(clusterInfo: ClusterInfo, versionMap: Map<KubernetesVersion, string>) : string {
+        let version: string;
+        if(!versionMap.get(clusterInfo.version)) {
```
shapirov103, line 144: nitpick: no need to get from the map twice on line 144 and 147. 
`let version = versionMap.get();`

shapirov103 (commented):
All good, one small nit.

lib/addons/core-addon/index.ts
zjaco13: resolved

shapirov103 (approved):
LGTM
pending e2e tests

shapirov103:
@elamaran11 please review the latest

elamaran11 (approved):
LGTM

elamaran11:
> @elamaran11 please review the latest

We can merge after e2e.

PR #751, opened 2023-07-05
Bump ArgoCD default version to 5.37.0 (currently latest)

tsahiduek:
Argo CD helm version is outdated.
Helm chart version 5 introduced couple of changes, and I think we should upgrade it. It was tested on the pipeline-multi-env-gitops pattern

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/addons/argocd/index.ts
+++ b/lib/addons/argocd/index.ts
@@ -26,7 +26,7 @@ export interface ArgoCDAddOnProps extends HelmAddOnUserProps {
 
     /**
     * Helm chart version to use to install.
-    * @default 3.33.5
+    * @default 5.37.0
     */
     version?: string;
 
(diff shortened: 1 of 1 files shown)


elamaran11 (approved):
LGTM

shapirov103:
@tsahiduek the reason why I deferred upgrade to the latest version was because of the differences they introduced in handling private repositories. That is SSH key, github token, user/pwd authorization. Before we merge it, have you validated the current approach still works? I believe my initial research spotted some incompatibilities.

shapirov103 (changes requested):
Please see my comment on private repo support. Happy to merge if it is not going to break the patterns for private repos.

tsahiduek:
> 

Hi @shapirov103 ,
I've tested this with the GitOps Multi-env pattern that uses ArgoCD with SSH connection type --> https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/lib/pipeline-multi-env-gitops/index.ts#L250
Is this is what you meant?

shapirov103 (commented):
@tsahiduek I also see indirect confirmation of the quality/compatibility here: https://github.com/aws-quickstart/cdk-eks-blueprints/issues/671#issuecomment-1623680147

LGTM

shapirov103 (approved):
LGTM

Issue #752, opened 2023-07-05
Add secret name validation in ArgoCD for private repos to avoid incompatible characters resulting in runtime failures that are hard to troubleshoot


shapirov103:
I finally found some interesting logs:

```
E0704 11:50:41.576864       1 secretproviderclasspodstatus_controller.go:336] "failed to create Kubernetes secret" err="Secret \"argocd-repo-ssh_key\" is invalid: metadata.name: Invalid value: \"argocd-repo-ssh_key\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" spc="argocd/blueprints-secret" pod="argocd/blueprints-addon-argocd-server-5c966d46df-sd2t5" secret="argocd/argocd-repo-ssh_key" spcps="argocd/blueprints-addon-argocd-server-5c966d46df-sd2t5-argocd-blueprints-secret"
```

_Originally posted by @cyril94440 in https://github.com/aws-quickstart/cdk-eks-blueprints/issues/671#issuecomment-1620139249_

shapirov103:
Note for implementation: should use the internal constraints validation framework https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/internal/input-validations-framework-readme.md

shapirov103:
resolved by https://github.com/aws-quickstart/cdk-eks-blueprints/pull/757

PR #757, opened 2023-07-05
Argocd check

zjaco13:
*Issue #, if available:* #752

*Description of changes:*
Added new constraints to ensure argocd credentials secret name will be correct regex.  Also added composite constraint and general regex constraint.

### DIFF
--- a/lib/spi/types.ts
+++ b/lib/spi/types.ts
@@ -86,7 +86,7 @@ export interface ApplicationRepository extends GitRepositoryReference {
  * Adds Constraints to application repository
  */
 export class ApplicationRepositoryConstraints implements constraints.ConstraintsType<ApplicationRepository> {
-    credentialsSecretName = new constraints.DomainNameConstraint();
+    credentialsSecretName = new constraints.DomainNameStringConstraint();
 }
 
 /**
--- a/lib/spi/types.ts
+++ b/lib/spi/types.ts
@@ -6,7 +6,7 @@ import { Construct, IConstruct } from 'constructs';
 import { ResourceProvider } from '.';
 import { EksBlueprintProps } from '../stacks';
 import { logger } from "../utils/log-utils";
-
+import * as constraints from '../utils/constraints-utils';
 /**
  * Data type defining helm repositories for GitOps bootstrapping.
  */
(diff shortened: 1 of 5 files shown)


shapirov103 (approved):
Looks really good. I was wondering where to place the validation logic, but given the current framework constraints your placement is reasonable. 
We should think about extending the framework.

One approach is to have a decorator which we can attach to the props directly, e.g. 

```typescript
export interface ArgoCDAddOnProps extends HelmAddOnUserProps {
    bootstrapRepo?: spi.ApplicationRepository;
}

export interface ApplicationRepository extends GitRepositoryReference {

    @validate(InternetHostStringConstraint)
    credentialsSecretName?: string,

}

export class ArgoCDAddOn implements spi.ClusterAddOn, spi.ClusterPostDeploy {
    constructor(props?: ArgoCDAddOnProps) {
       utils.validateConstaints(props, ArgoCDAddOn.name); // note we pass the whole object. If there is a constraint registered against any nested types it will throw. 
    }
```

Another approach is to have a decorator that will validate right on set so we don't even need to invoke utils.validateConstraints (need POC with interfaces). 

That is for later though.

lib/spi/types.ts
```diff
@@ -82,6 +82,12 @@ export interface ApplicationRepository extends GitRepositoryReference {
     credentialsType?: "USERNAME" | "TOKEN" | "SSH"
 
 }
+/**
+ * Adds Constraints to application repository
+ */
+export class ApplicationRepositoryConstraints implements constraints.ConstraintsType<ApplicationRepository> {
+    credentialsSecretName = new constraints.InternetHostStringConstraint();
```
vincentgna, line 89: Shouldn't it be allowed to use secrets named as `/${product}/${stage}/cluster-bootstrap/${repo-credentials}` ?

shapirov103: It will require a mapping logic to transform that name to the k8s secret name. The current design is that the secret you create in AWS Secrets Manager is propagated as is (with the same name) to EKS. The k8s secret name constraint is:

> You can create a name with a maximum length of 253 characters using only alphanumeric characters and hyphens. Names cannot start with a hyphen and the alpha characters can only be lowercase.

Please feel free to create a feature request to implement this logic for the name translation.

PR #776, opened 2023-07-14
Feature/fluxcd multi path support

freschri:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/addons/fluxcd/index.ts
+++ b/lib/addons/fluxcd/index.ts
@@ -65,6 +65,11 @@ export interface FluxCDAddOnProps extends HelmAddOnUserProps {
   fluxTargetNamespace?: string;
 
   /** 
+  * Flux Kustomization paths within the repository;
+  * Flux Kustomizations will be created for bootstrapRepo.path and for the paths specified in this array */
+  additionalFluxKustomizationPaths?: string[];
+
+  /** 
   * Flux Kustomization Prune.
   * Default `true` */
   fluxPrune?: boolean;
@@ -119,10 +124,11 @@ export class FluxCDAddOn extends HelmAddOn {
 
     //Lets create a GitRepository resource as a source to Flux
     if (this.options.bootstrapRepo?.repoUrl) {
-      const gitRepositoryConstruct = createGitRepository(clusterInfo, this.options.bootstrapRepo, this.options);
+      const gitRepositoryConstruct = createGitRepository(clusterInfo, this.options);
       gitRepositoryConstruct.node.addDependency(chart);
-      const kustomizationConstruct = createKustomization(clusterInfo, this.options.bootstrapRepo, this.options);
-      kustomizationConstruct.node.addDependency(gitRepositoryConstruct);
+
+      const kustomizationConstructs = createKustomizations(clusterInfo, this.options);
+      kustomizationConstructs.map(kustomizationConstruct => kustomizationConstruct.node.addDependency(gitRepositoryConstruct));
     }
     return Promise.resolve(chart);
   }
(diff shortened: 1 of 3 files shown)


lib/addons/fluxcd/index.ts
```diff
@@ -131,19 +137,42 @@ export class FluxCDAddOn extends HelmAddOn {
...
   return construct;
 }
 
 /**
- * create Kustomization calls the FluxKustomization().generate to create Kustomization resource.
+ * create Kustomizations calls the FluxKustomization().generate multiple times to create Kustomization resources.
  */
-function createKustomization(clusterInfo: ClusterInfo, bootstrapRepo: spi.ApplicationRepository, fluxcdAddonProps: FluxCDAddOnProps): KubernetesManifest {
-  const manifest = new FluxKustomization(bootstrapRepo).generate(fluxcdAddonProps.namespace!, fluxcdAddonProps.fluxSyncInterval!, fluxcdAddonProps.fluxTargetNamespace!, fluxcdAddonProps.fluxPrune!, fluxcdAddonProps.fluxTimeout!, fluxcdAddonProps.bootstrapValues!);
-  let manifestName: string | undefined = fluxcdAddonProps.name + 'kustomization';
```
elamaran11, line 146: Why is this logic removed?

lib/addons/fluxcd/index.ts
```diff
@@ -131,19 +137,42 @@ export class FluxCDAddOn extends HelmAddOn {
...
-  const construct = clusterInfo.cluster.addManifest(manifestName!, manifest);
-  return construct;
-}
\ No newline at end of file
+function createKustomizations(clusterInfo: ClusterInfo, fluxcdAddonProps: FluxCDAddOnProps): KubernetesManifest[] {
+  let fluxKustomizationPaths = fluxcdAddonProps.bootstrapRepo?.path ? [fluxcdAddonProps.bootstrapRepo?.path] : ["."];
```
elamaran11, line 155: What is this ?

elamaran11 (changes requested):
Have some comments. Also you might have to add a new documentation section showing how to demonstrate multi paths.

lib/addons/fluxcd/index.ts
freschri: it is not removed, it is extended to multi kustomizations in the function below

lib/addons/fluxcd/index.ts
freschri: it by "this" you mean:
  let fluxKustomizationPaths = fluxcdAddonProps.bootstrapRepo?.path ? [fluxcdAddonProps.bootstrapRepo?.path] : ["."];

the code assigns an array of a single element to the variable. it assigns "path", if path is assigned by user, otherwise it assigns the root directory of the repo. the array will be extended later with the additional paths if assigned.

shapirov103 (commented):
Looks good to me. Let's add documentation to this PR describing this support. It would also help if you provide a use case when this can be used (both to the PR and to the docs). While we know the context, the customer and other contributors do not.

elamaran11:
@freschri Did you get a chance to close the doc request for this PR. We can add this to patch release.

freschri:
@elamaran11 @shapirov103 following some more thinking about how Flux is supposed to work, I think the problem I am trying to fix comes from the repo structure: looking at https://github.com/fluxcd/flux2-kustomize-helm-example/tree/main I can see that there should be a single path where if needed, other Kustomizations are stored, with possible dependencies referencing other paths within the repo, so that the cluster setup is completely defined.

So, I updated the observability repo with an additional cdk folder, where we store our single path and related Kustomizations, pointing to the common folders (so avoiding interfering with terraform):
https://github.com/freschri/aws-observability-accelerator/blob/java-pattern/artifacts/grafana-operator-manifests/eks/cdk/patterns/java/java.yaml
please let me know what you think

elamaran11:
@freschri I like the way you are thinking about different approaches to solve problems. Definitely this is highly encouraged. 
Following are the comments from my end on this approach.

Firstly, think we are confusing too many things together here across many repos to get to an outcome. Multi path support by itself is a good feature to have to our FluxCD Addon irrespective of it is needed for Observability Accelerator or not. So i think we should find ways to close the development of this feature as this is already started. We will find other customers using this feature.
Secondly, having `cdk` references in `aws-observability-accelerator` is not the right approach. We have designed in a way to hold only common jsons, manifests and yamls in this repo which will be usable across both terraform and cdk accelerator solutions so we should not keep cdk related references so we should avoid this approach.
Thirdly, we wanted to strive feature parity between terraform accelerator and cdk accelerator. If you take a look at terraform accelerator we are using Kustomizations with terraform code for Java and Nnginx monitoring and we would want to follow the same strictly and that is also the right approach.
Fourth, Have Kubernetes Manifest yaml part of CDK is nothing new and you can check CDK observability Accelerator open source pattern to see how i have handled external secrets and cluster secret store for grafana access. We should follow the same process to close the thread here. 
Also as discussed over slack, we recommend have a conversation to close the loop on these design decisions. We are not making good progress for all your amazing efforts by just having GitHub comments and Slack chats.

freschri:
> @freschri I like the way you are thinking about different approaches to solve problems. Definitely this is highly encouraged. Following are the comments from my end on this approach.
> 
> Firstly, think we are confusing too many things together here across many repos to get to an outcome. Multi path support by itself is a good feature to have to our FluxCD Addon irrespective of it is needed for Observability Accelerator or not. So i think we should find ways to close the development of this feature as this is already started. We will find other customers using this feature. Secondly, having `cdk` references in `aws-observability-accelerator` is not the right approach. We have designed in a way to hold only common jsons, manifests and yamls in this repo which will be usable across both terraform and cdk accelerator solutions so we should not keep cdk related references so we should avoid this approach. Thirdly, we wanted to strive feature parity between terraform accelerator and cdk accelerator. If you take a look at terraform accelerator we are using Kustomizations with terraform code for Java and Nnginx monitoring and we would want to follow the same strictly and that is also the right approach. Fourth, Have Kubernetes Manifest yaml part of CDK is nothing new and you can check CDK observability Accelerator open source pattern to see how i have handled external secrets and cluster secret store for grafana access. We should follow the same process to close the thread here. Also as discussed over slack, we recommend have a conversation to close the loop on these design decisions. We are not making good progress for all your amazing efforts by just having GitHub comments and Slack chats.

like discussed completing this PR since feature is convenient and used in the field

freschri:
> @freschri Did you get a chance to close the doc request for this PR. We can add this to patch release.

done

elamaran11 (approved):
LGTM.

elamaran11:
@shapirov103 LGTM. Please check from your end.  Also we should run `e2e`.

shapirov103 (approved):

PR #779, opened 2023-07-16
new withEnv function

### LINKED ISSUE #778: Stack builder - add support to clone using `cdk.Environment` rather than strings of account and region
### Describe the feature

AWS CDK has a basic type of cdk.Environment, which is an object holds the account and the region for a target environemtn for deployment.
Having a new clone function that accept a `cdk.Evironment` type can simplify code using the clone method

### Use Case

Cloning when using already existing CDK Environment variables 

### Proposed Solution

Creating another function `cloneEnv` to add this support

### Other Information

_No response_

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.83.1 (build 006b542)

### EKS Blueprints Version

1.10.0

### Node.js Version

v20.2.0

MacOS Ventura 13.4

tsahiduek:
Fix #778.
Add `cloneEnv` method.
See description of 778 --> 
AWS CDK has a basic type of cdk.Environment, which is an object holds the account and the region for a target environemtn for deployment.
Having a new clone function that accept a cdk.Evironment type can simplify code using the clone method

### DIFF
--- a/lib/stacks/eks-blueprint-stack.ts
+++ b/lib/stacks/eks-blueprint-stack.ts
@@ -191,6 +191,11 @@ export class BlueprintBuilder implements spi.AsyncStackBuilder {
             .account(account ?? this.env.account).region(region ?? this.env.region);
     }
 
+    public cloneEnv(env? : cdk.Environment): BlueprintBuilder {
+        return new BlueprintBuilder().withBlueprintProps(this.props)
+            .account(env?.account ?? this.env.account).region(env?.region ?? this.env.region);
+    }
+
     public build(scope: Construct, id: string, stackProps?: cdk.StackProps): EksBlueprint {
         return new EksBlueprint(scope, { ...this.props, ...{ id } },
             { ...{ env: this.env }, ...stackProps });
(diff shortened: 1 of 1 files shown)


shapirov103 (commented):
@tsahiduek I like your thinking and making code level improvements. 
However, cloneEnv is not very idiomatic in typescript (clone is clone). The approach that I planned for such case is `with`, for example `builder.withBlueprintProps` that we have. 

I propose a change to the method and renaming to `withEnv` similar to `withBlueprintProps`. Then you can achieve your scenario like:

```typescript
blueprint.clone().withEnv(env);
```

Also please make env mandatory in the `withEnv` method. Implementation will change to just set `this.env` to the pass param.

tsahiduek:
> @tsahiduek I like your thinking and making code level improvements. However, cloneEnv is not very idiomatic in typescript (clone is clone). The approach that I planned for such case is `with`, for example `builder.withBlueprintProps` that we have.
> 
> I propose a change to the method and renaming to `withEnv` similar to `withBlueprintProps`. Then you can achieve your scenario like:
> 
> ```ts
> blueprint.clone().withEnv(env);
> ```
> 
> Also please make env mandatory in the `withEnv` method. Implementation will change to just set `this.env` to the pass param.

Thank you @shapirov103 . Just pushed an updated commit as requested.

Issue #792, opened 2023-07-26
Version Field as Optional Field in GenericClusterProvider which defaults to latest EKS Version


smrutiranjantripathy:
### Describe the feature

We would need the version parameter in GenericClusterProvider to be optional. Just like without mentioning provider it launches with the latest available EKS Cluster Version. This would help to create cluster with the latest version which has access to a specific role or an existing IAM Role.

```
const clusterProvider = new blueprints.GenericClusterProvider({
    version: eks.KubernetesVersion.V1_27,  ----> Optional
    mastersRole: blueprints.getResource(context => {
        return iam.Role.fromRoleName(context.scope, "my-role", "Role Name")})
    });
```
    
### Use Case

We are offering a sample blueprint to a customer as a use case of leveraging the EKS Cluster with certain Workloads.
This will be a sample blueprint which will be used by Customer and we don't want to update that blueprint time and again by coding the version. If the version is optional this will remove the pain of updating the version every time with certain workloads.

### Proposed Solution

Don't have a way to implement this. Will have to do a deep dive on the code to figure out a way to do that.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.88.0

### EKS Blueprints Version

1.10.1

### Node.js Version

v14.17.3

### Environment details (OS name and version, etc.)

OSX 13.4

shapirov103:
The implementation should align with the following logic:

blueprintbuilder will have an opt-in method to always use latest, e.g. builder().version("auto"). If version is set to auto then cluster provider should not require version. If builder.version is set to an explicit value then cluster provider should also not force the version to be specified. 
If customer did not set the version at the blueprint level, and did not opt in to use latest "auto", then clusterprovider must require version to be set. 
@elamaran11 if you have a candidate for this please let me know (e.g. Zach).

elamaran11:
@shapirov103 Zach would be able to turn this out quick. Let me ask him.

shapirov103:
Completed with https://github.com/aws-quickstart/cdk-eks-blueprints/pull/793

PR #793, opened 2023-07-26
Optional version Cluster provider

zjaco13:
*Issue #, if available:* #792

*Description of changes:*
Made version optional for all cluster providers as long as .version() is run on the builder

### DIFF
--- a/lib/cluster-providers/generic-cluster-provider.ts
+++ b/lib/cluster-providers/generic-cluster-provider.ts
@@ -30,16 +30,16 @@ export function clusterBuilder() {
  * @returns ILayerVersion or undefined
  */
 export function selectKubectlLayer(scope: Construct, version: eks.KubernetesVersion): ILayerVersion | undefined {
-    switch(version) {
-        case eks.KubernetesVersion.V1_23:
+    switch(version.version) {
+        case "1.23":
             return new KubectlV23Layer(scope, "kubectllayer23");
-        case eks.KubernetesVersion.V1_24:
+        case "1.24":
             return new KubectlV24Layer(scope, "kubectllayer24");
-        case eks.KubernetesVersion.V1_25:
+        case "1.25":
             return new KubectlV25Layer(scope, "kubectllayer25");
-        case eks.KubernetesVersion.V1_26:
+        case "1.26":
             return new KubectlV26Layer(scope, "kubectllayer26");
-        case eks.KubernetesVersion.V1_27:
+        //case "1.27":
             //return new KubectlV27Layer(scope, "kubectllayer27");
     }
     
--- a/lib/cluster-providers/generic-cluster-provider.ts
+++ b/lib/cluster-providers/generic-cluster-provider.ts
@@ -248,7 +248,7 @@ export class GenericClusterProvider implements ClusterProvider {
         if(builder.versionCalled) {
            version = builder.props.version!;
         } else {
-            version = this.props.version!
+            version = this.props.version!;
         }
         const privateCluster = this.props.privateCluster ?? utils.valueFromContext(scope, constants.PRIVATE_CLUSTER, false);
         const endpointAccess = (privateCluster === true) ? eks.EndpointAccess.PRIVATE : eks.EndpointAccess.PUBLIC_AND_PRIVATE;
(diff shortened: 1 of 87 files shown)


shapirov103 (commented):
@zjaco13 let's refactor to make the logic simpler. We already have version method on the builder. We should just make version required to simplify and it is either an actual version or auto. 

Please see my comment on the casts that you applied, it is a bit confusing. Let's meet tomorrow (~2 pm EST for example) to discuss.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -236,18 +240,26 @@ export class GenericClusterProvider implements ClusterProvider {
     createCluster(scope: Construct, vpc: ec2.IVpc, secretsEncryptionKey: IKey | undefined): ClusterInfo {
         const id = scope.node.id;
 
+        const builder = scope as unknown as BlueprintBuilder;
```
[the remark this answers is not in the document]
zjaco13: You're right.  I'm not sure how to figure out whether or not the version function has been called, since the properties do not get passed onto the blueprint.

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -132,8 +132,16 @@ export class BlueprintBuilder implements spi.AsyncStackBuilder {
         return this;
     }
 
-    public version(version: KubernetesVersion): this {
-        this.props = { ...this.props, ...{ version } };
+    public version(version: string | KubernetesVersion): this {
```
shapirov103, line 135: Ideally, there will be one location in the codebase for the auto-selected version.

shapirov103 (commented):
Almost there, a  couple of comments, mostly minor. Thanks for the test cases, it helps validate. If something is unclear please ping me on slack.

lib/spi/cluster-contracts.ts
```diff
@@ -2,12 +2,13 @@ import { ClusterInfo } from '.';
...
 import {IKey} from "aws-cdk-lib/aws-kms";
+import { KubernetesVersion } from 'aws-cdk-lib/aws-eks';
 
 
 /**
  * ClusterProvider is the interface to which all Cluster Providers should conform.
  */
 export declare interface ClusterProvider {
-    createCluster(scope: Construct, vpc: IVpc, secretsEncryptionKey?: IKey): ClusterInfo;
+    createCluster(scope: Construct, vpc: IVpc, secretsEncryptionKey?: IKey, blueprintVersion?: KubernetesVersion): ClusterInfo;
```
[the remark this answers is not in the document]
zjaco13: resolved

lib/stacks/eks-blueprint-stack.ts
shapirov103: @zjaco13 can we make this type `version: KubernetesVersion | "auto"`? This will take care of compile time check for correctness of the passed values.

shapirov103 (commented):
@zjaco13 looks great, a bit more work than expected, but that is generally what happens with the framework level changes in the types. 
I added one comment, and I believe what is missing is the doc updates, since examples will start failing now in our docs. 
The patterns will need to be updated as well, but that will be a separate task when we uprev the patterns repo.

lib/stacks/eks-blueprint-stack.ts
zjaco13: done

lib/stacks/eks-blueprint-stack.ts
```diff
@@ -42,7 +42,7 @@ export class EksBlueprintProps {
     /**
      * Kubernetes version (must be initialized for addons to work properly)
      */
-    readonly version?: KubernetesVersion = KubernetesVersion.V1_25;
+    readonly version?: KubernetesVersion | "auto" = KubernetesVersion.V1_25;
```
shapirov103, line 45: Let's remove this assignment to 1.25 - 1/ the auto is setting to 1.27 and 2/ we don't want more than one place where defaults will apply.

shapirov103 (dismissed):
LGTM

shapirov103:
Tests failed (`make run-test`) due to the main updated with a test case that does not have version() included. Please rebase again, I assume you can just merge the main branch and resubmit.

shapirov103 (approved):

Issue #376, opened 2022-05-06
Non-idiomatic use of Map throughout


otterley:
### Describe the bug

Map types aren't idiomatic for Typescript. Where types like `Map<String,T>` are used, they should be replaced with `{ [key:string]: T }`. 

### Expected Behavior

N/A

### Current Behavior

N/A

### Reproduction Steps

N/A

### CDK CLI Version

any

### Node.js Version

any

### Environment details (OS name and version, etc.)

N/A

shapirov103:
Thank you, @otterley. I see the API usability issue. I spotted this a bit late and this feature is part of the stable API.
 
We are using the record style map approach in the new `GenericClusterProvider` [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/cluster-providers/generic-cluster-provider.ts#L42).

What I can do to address the usability and preserve backwards compatibility is simplify construction of the fargate profiles for the users by adding convenience methods.

otterley:
Have you considered the possibility of using [union types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to fix the problem without breaking existing code?

shapirov103:
@otterley I will take a look.

PR #798, opened 2023-08-01
Use TypeScript mapping for Fargate profile

otterley:
Use a TypeScript-native mapping type for Fargate profiles instead of the non-idiomatic `Map<>` type. The old type remains supported for backward compatibility, but should be deprecated.


otterley:
Partially addresses #376

shapirov103 (approved):
@otterley thank you for your contributions, appreciate keeping it backwards compatible.

PR #804, opened 2023-08-02
Observability Builder for Observability Accelerator

elamaran11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/builders/observability-builder.md
+++ b/docs/builders/observability-builder.md
@@ -23,9 +23,6 @@ export default class SingleNewEksConstruct {
         const region = process.env.COA_AWS_REGION! || process.env.CDK_DEFAULT_REGION!;
         
         const addOns: Array<blueprints.ClusterAddOn> = [
-            new blueprints.addons.KubeProxyAddOn(),
-            new blueprints.addons.AwsLoadBalancerControllerAddOn(),
-            new blueprints.addons.CertManagerAddOn(),
             new blueprints.addons.CloudWatchLogsAddon({
                 logGroupPrefix: `/aws/eks/${stackId}`,
                 logRetentionDays: 30
@@ -37,7 +34,7 @@ export default class SingleNewEksConstruct {
         ObservabilityBuilder.builder()
             .account(account)
             .region(region)
-            .addNewClusterObservabilityBuilderAddOns()
+            .addNativeObservabilityBuilderAddOns()
             .addOns(...addOns)
             .build(scope, stackId);
     }
@@ -106,7 +103,7 @@ export default class ExistingEksMixedobservabilityConstruct {
         ObservabilityBuilder.builder()
             .account(account)
             .region(region)
-            .addExistingClusterObservabilityBuilderAddOns()
+            .addMixedObservabilityBuilderAddOns()
             .clusterProvider(importClusterProvider)
             .resourceProvider(blueprints.GlobalResources.Vpc, new blueprints.VpcProvider(vpcId)) 
             .addOns(...addOns)
--- a/docs/builders/observability-builder.md
+++ b/docs/builders/observability-builder.md
@@ -1,0 +1,116 @@
+# Observability Builder
+
+The `ObservabilityBuilder` allows you to get started with a builder class to configure required addons as you prepare a blueprint for setting up Observability on an existing EKS cluster or a new EKS cluster.
+
+## Usage 
+
+The framework provides a couple of convenience methods to instantiate the `` by leveraging the SDK API calls.
+
+### Usage 1 - Observability For a New EKS Cluster
+
+The below usage helps you with a demonstration to use `ObservabilityBuilder` to setup required addons as you prepare a blueprint for setting up observability on a new EKS cluster.
+
+```typescript
+import { Construct } from 'constructs';
+import * as blueprints from '@aws-quickstart/eks-blueprints';
+import { ObservabilityBuilder } from '@aws-quickstart/eks-blueprints';
+
+export default class SingleNewEksConstruct {
+    constructor(scope: Construct, id: string) {
+        const stackId = `${id}-observability-accelerator`;
+
+        const account = process.env.COA_ACCOUNT_ID! || process.env.CDK_DEFAULT_ACCOUNT!;
+        const region = process.env.COA_AWS_REGION! || process.env.CDK_DEFAULT_REGION!;
+        
+        const addOns: Array<blueprints.ClusterAddOn> = [
+            new blueprints.addons.KubeProxyAddOn(),
+            new blueprints.addons.AwsLoadBalancerControllerAddOn(),
+            new blueprints.addons.CertManagerAddOn(),
+            new blueprints.addons.CloudWatchLogsAddon({
+                logGroupPrefix: `/aws/eks/${stackId}`,
+                logRetentionDays: 30
+            }),
+            new blueprints.addons.ContainerInsightsAddOn(),
+            new blueprints.addons.XrayAddOn()
+        ];
+
+        ObservabilityBuilder.builder()
+            .account(account)
+            .region(region)
+            .addNewClusterObservabilityBuilderAddOns()
+            .addOns(...addOns)
+            .build(scope, stackId);
+    }
+}
+
+```
+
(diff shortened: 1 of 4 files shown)


lib/builders/observability-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+            new blueprints.addons.VpcCniAddOn(),
+            new blueprints.addons.CoreDnsAddOn(),
+            new blueprints.addons.MetricsServerAddOn(),
+            new blueprints.addons.PrometheusNodeExporterAddOn(),
+            new blueprints.addons.KubeStateMetricsAddOn());
+    }
+
+    public addExistingClusterObservabilityBuilderAddOns(): ObservabilityBuilder {
+        return this.addOns(
+            new blueprints.addons.AwsLoadBalancerControllerAddOn(),
```
shapirov103, line 19: Why Cert manager and AWS Load balancer controller not included in the new cluster addons but are included in the existing cluster? what relevance do they have for existing cluster, for example what if the existing cluster already has this software?

docs/builders/observability-builder.md
```diff
@@ -0,0 +1,116 @@
+# Observability Builder
```
shapirov103, line 1: This has to be included in the overall navigation. You can add a top level module, like builders. But it has to have a doc on the regular builder (we probably already have this content) and a doc on the observability builder that you already have. We should also have a top level into about builders in general, which I can help with.

shapirov103 (commented):
A couple of comments

docs/builders/observability-builder.md
elamaran11: Added the required navigation to `mkdocs.yml` and it will be awesome if you can add an `index.md` for builders in general.

lib/builders/observability-builder.ts
elamaran11: The relevance is we have bunch of addons to new cluster addons but we have nothing in existing cluster addons. So we cant keep the existing cluster addons array empty. I know this will impact if the existing cluster already has ALB and Cert manager but not sure how to approach this. Any thoughts if we can leave addons array empty in existing cluster?

shapirov103: So the new cluster add-ons will just not work on the existing cluster, is that what you are saying?

lib/builders/observability-builder.ts
```diff
@@ -0,0 +1,77 @@
...
+    public addNativeObservabilityBuilderAddOns(): ObservabilityBuilder {
+        return this.addOns(
+            new blueprints.addons.AwsLoadBalancerControllerAddOn(),
+            new blueprints.addons.CertManagerAddOn(),
+            new blueprints.addons.CoreDnsAddOn(),
+            new blueprints.addons.KubeProxyAddOn(),
+            new blueprints.addons.KubeStateMetricsAddOn(),
+            new blueprints.addons.MetricsServerAddOn(),
+            new blueprints.addons.PrometheusNodeExporterAddOn(),
+            new blueprints.addons.VpcCniAddOn());
```
shapirov103, line 17: Do we need Vpc cni add-on here? My concern is that frequently customers will want to use custom networking when they run something that needs observability. Can we move it out of these methods?

lib/builders/observability-builder.ts
```diff
@@ -0,0 +1,77 @@
+import * as blueprints from '../../lib';
+import * as utils from "../utils";
+import { NestedStack, NestedStackProps } from 'aws-cdk-lib';
+import { Construct } from 'constructs';
+
+export class ObservabilityBuilder extends blueprints.BlueprintBuilder {
+
+    public addNativeObservabilityBuilderAddOns(): ObservabilityBuilder {
```
shapirov103, line 8: Why not add container insights right here? E.g. if the method optionally passes CI config, we can add it right away.

lib/builders/observability-builder.ts
```diff
@@ -0,0 +1,77 @@
...
+            new blueprints.addons.CertManagerAddOn(),
+            new blueprints.addons.CoreDnsAddOn(),
+            new blueprints.addons.KubeProxyAddOn(),
+            new blueprints.addons.KubeStateMetricsAddOn(),
+            new blueprints.addons.MetricsServerAddOn(),
+            new blueprints.addons.PrometheusNodeExporterAddOn(),
+            new blueprints.addons.VpcCniAddOn());
+    }
+
+    public addOpenSourceObservabilityBuilderAddOns(): ObservabilityBuilder {
```
shapirov103, line 33: Why not add AMP here, optionally accepting the config for it?

lib/builders/observability-builder.ts
```diff
@@ -0,0 +1,77 @@
...
+            new blueprints.addons.CertManagerAddOn(),
+            new blueprints.addons.CoreDnsAddOn(),
+            new blueprints.addons.KubeProxyAddOn(),
+            new blueprints.addons.KubeStateMetricsAddOn(),
+            new blueprints.addons.MetricsServerAddOn(),
+            new blueprints.addons.PrometheusNodeExporterAddOn(),
+            new blueprints.addons.VpcCniAddOn());
+    }
+
+    public addMixedObservabilityBuilderAddOns(): ObservabilityBuilder {
```
shapirov103, line 20: method name is too verbose, we should remove `builder` from it as this class already has the word builder in it. I also propose dropping `observability` from the method names for the same reason. 
 
If you add the actual observability addons then I propose `addMixedPatternAddons`.

If there are only prereqs and not the actual working solution, then change `add` to `enable`.

shapirov103 (commented):
@elamaran11 moving in the right direction. Added a few comments. 
Please update the public method for native/mixed/oss with a small js doc to describe what you mean. the method names require intimate knowledge of our thinking process and will be confusing to the customers without explanation.

lib/builders/observability-builder.ts
elamaran11: removed

lib/builders/observability-builder.ts
elamaran11: removed

lib/builders/observability-builder.ts
elamaran11: remove

lib/builders/observability-builder.ts
elamaran11: Fixed

elamaran11:
> @elamaran11 moving in the right direction. Added a few comments. Please update the public method for native/mixed/oss with a small js doc to describe what you mean. the method names require intimate knowledge of our thinking process and will be confusing to the customers without explanation.

@shapirov103 All comments are mixed, Please review and merge.

docs/builders/observability-builder.md
```diff
@@ -0,0 +1,121 @@
...
+                logRetentionDays: 30
+            }),
+            new blueprints.addons.VpcCniAddOn(),
+            new blueprints.addons.XrayAddOn()
+        ];
+
+        ObservabilityBuilder.builder()
+            .account(account)
+            .region(region)
+            .addNativeObservabilityBuilderAddOns()
```
shapirov103, line 45: method name wrong

docs/builders/observability-builder.md
```diff
@@ -0,0 +1,121 @@
...
+            }),
+            new blueprints.addons.VpcCniAddOn(),
+            cloudWatchAdotAddOn,
+            new blueprints.addons.XrayAdotAddOn(),
+        ];
+
+        ObservabilityBuilder.builder()
+            .account(account)
+            .region(region)
+            .addMixedObservabilityBuilderAddOns()
```
shapirov103, line 114: method name wrong

shapirov103 (commented):
Please see a couple of places that were missed once method names were refactored.

docs/builders/observability-builder.md
elamaran11: Nice catch.

docs/builders/observability-builder.md
elamaran11: Nice catch

elamaran11:
> Please see a couple of places that were missed once method names were refactored.

Nice catch, issues are fixed.

Issue #799, opened 2023-08-02
[Bug] Karpenter add-on: interruption handling partially enabled by default


otterley:
By default, the `interruptionHandling` attribute for a Karpenter add-on is set to `undefined` (i.e., `false`) but the global configuration has it enabled, leading to errors like:

```
2023-08-02T04:15:03.603Z        ERROR   controller.interruption getting messages from queue, discovering queue url, fetching queue url, AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.
```

Here's what the ConfigMap looks like:

```
$ kubectl -n karpenter get configmap karpenter-global-settings -o yaml
apiVersion: v1
data:
  aws.interruptionQueueName: <stack-name>
  [snip]
```

If `interruptionHandling` is enabled, this key should not be set in the ConfigMap.

Separately, I wonder why  `interruptionHandling` is disabled by default. It should probably be enabled by default, as that would be a better customer experience.

youngjeong46:
@otterley Thank you for catching on the bug - I will push a quick fix to resolve this issue. As for your other request - we will discuss and probably incorporate it as part of the fast follow (to incorporate 0.29)

youngjeong46:
@otterley - this fix is merged, closing the ticket for now. Let us know if the issue persists on your end.

PR #807, opened 2023-08-03
fix(addon/karpenter): Disable Interruption Queue on Helm when the addon flag is false. 

youngjeong46:
*Issue #, if available:* #799

*Description of changes:* Interruption Queue Name was resolved to the stack name even though the queue was disabled. This PR will fix to ensure if the queue is disabled, then the helm interruption queue name will resolve to a blank value (as intended for default helm value).

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -280,7 +280,7 @@ export class KarpenterAddOn extends HelmAddOn {
                 clusterEndpoint: endpoint,
                 clusterName: name,
                 defaultInstanceProfile: karpenterInstanceProfile.instanceProfileName,
-                interruptionQueueName: stackName
+                interruptionQueueName: interruption ? stackName : "",
             });
         } else {
             setPath(values, "clusterEndpoint", endpoint);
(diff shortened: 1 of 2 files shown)


lib/addons/karpenter/index.ts
```diff
@@ -280,7 +280,7 @@ export class KarpenterAddOn extends HelmAddOn {
                 clusterEndpoint: endpoint,
                 clusterName: name,
                 defaultInstanceProfile: karpenterInstanceProfile.instanceProfileName,
-                interruptionQueueName: stackName
+                interruptionQueueName: interruption ? stackName : "",
```
shapirov103, line 283: If interruptionQueueName is empty string, is the same as not providing the interruptionQueueName attribute at all? Was it tested?

shapirov103 (commented):
One question, looks good otherwise.

lib/addons/karpenter/index.ts
youngjeong46: yes - the default value is "" as shown [here](https://github.com/aws/karpenter/blob/5e02e4c6eb3432b5da12df100474a114db66607f/charts/karpenter/values.yaml#L187)

shapirov103 (approved):
LGTM

PR #814, opened 2023-08-08
Fluxcd fixes

zjaco13:
*Issue #, if available:*

*Description of changes:*
Added a new fluxcd repository object, and array of those objects to fluxcd props to allow for multiple repositories and multiple paths.

Changes are breaking changes

### DIFF
--- a/docs/addons/fluxcd.md
+++ b/docs/addons/fluxcd.md
@@ -43,16 +43,18 @@ import * as blueprints from '@aws-quickstart/eks-blueprints';
 const app = new cdk.App();
 
 const addOn = new blueprints.addons.FluxCDAddOn({
-  bootstrapRepo: {
-      repoUrl: 'https://github.com/aws-observability/aws-observability-accelerator',
-      name: "aws-observability-accelerator",
-      targetRevision: "main",
-      path: "./artifacts/grafana-operator-manifests/eks/infrastructure"
-  },
-  bootstrapValues: {
-      "region": "us-east-1"
-  },
-  additionalFluxKustomizationPaths: ["./artifacts/grafana-operator-manifests/eks/java"]
+    repositories: [{
+        bootstrapRepo: {
+            repoUrl: 'https://github.com/aws-observability/aws-observability-accelerator',
+            name: "aws-observability-accelerator",
+            targetRevision: "main",
+            path: "./artifacts/grafana-operator-manifests/eks/infrastructure"
+        },
+        bootstrapValues: {
+            "region": "us-east-1"
+        },
+        additionalFluxKustomizationPaths: ["./artifacts/grafana-operator-manifests/eks/java"]
+    }],
 })
 ...
 
@@ -75,26 +77,28 @@ const app = new cdk.App();
 const nginxDashUrl = "https://raw.githubusercontent.com/aws-observability/aws-observability-accelerator/main/artifacts/grafana-dashboards/eks/nginx/nginx.json"
 
 const addOn = new blueprints.addons.FluxCDAddOn({
-    bootstrapRepo: {
-        repoUrl: 'https://github.com/stefanprodan/podinfo',
-        name: "podinfo",
-        targetRevision: "master",
-        path: "./kustomize"
-    },
-    bootstrapValues: {
-        "region": "us-east-1"
-    },
-    workloadApplications: [
+    repositories: [
         {
-            name: "nginx-grafanadashboard",
-            namespace: "grafana-operator",
-            repository: {
+            bootstrapRepo: {
                 repoUrl: 'https://github.com/aws-observability/aws-observability-accelerator',
+                name: "aws-observability-accelerator",
                 targetRevision: "main",
-                path: "./artifacts/grafana-operator-manifests/eks/nginx"
+                path: "./artifacts/grafana-operator-manifests/eks/infrastructure"
             },
-            values: {
-                "GRAFANA_NGINX_DASH_URL" : nginxDashUrl,
+            bootstrapValues: {
+                "region": "us-east-1"
+            },
+            additionalFluxKustomizationPaths: ["./artifacts/grafana-operator-manifests/eks/java"]
+        },
+        {
+            bootstrapRepo: {
+                repoUrl: 'https://github.com/stefanprodan/podinfo',
+                name: "podinfo",
+                targetRevision: "master",
+                path: "./kustomize"
+            },
+            bootstrapValues: {
+                "region": "us-east-1"
             },
         }
     ],
@@ -139,18 +143,22 @@ const app = new cdk.App();
 const addOns: Array<blueprints.ClusterAddOn> = [
   new blueprints.addons.ExternalsSecretsAddOn(),
   new blueprints.addons.FluxCDAddOn({
-    bootstrapRepo: {
-        repoUrl: '<<YOUR_PRIVATE_GIT_REPOSITORY>>',
-        name: "<<YOUR_FLUX_APP_NAME>>",
-        targetRevision: "<<YOUR_TARGET_REVISION>>",
-        path: "<<YOUR_FLUX_SYNC_PATH>>",
-        fluxVerifyMode: "head",
+    repositories:[
+      {
+        bootstrapRepo: {
+            repoUrl: '<<YOUR_PRIVATE_GIT_REPOSITORY>>',
+            name: "<<YOUR_FLUX_APP_NAME>>",
+            targetRevision: "<<YOUR_TARGET_REVISION>>",
+            path: "<<YOUR_FLUX_SYNC_PATH>>",
+        },
+        bootstrapValues: {
+            "region": "us-east-1"
+        },
         // This is the name of the kubernetes secret to be created by `ExternalSecret` shown in step 3.
         fluxSecretRefName: "repository-creds" 
-    },
-    bootstrapValues: {
-        "region": "us-east-1"
-    },
+      }
+    ],
+
   }),
   new ExternalOperatorSecretAddon(),
 ];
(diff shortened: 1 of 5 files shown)


lib/addons/fluxcd/index.ts
```diff
@@ -10,22 +10,9 @@ import { FluxKustomization } from "./kustomization";
...
-   */
-  namespace?: string;
-
-  /**
-  * Helm chart version to use to install.
-  * @default 2.8.0
-  */
-  version?: string;
-
+export interface FluxCDGitRepo {
```
shapirov103, line 15: It should derive from `GitopsApplicationDeployment ` and add fields that are flux specific.

shapirov103, line 15: Why is bootstrapRepository optional here? Does this structure make sense without bootstrap repo? See my other comment that it should really extend `GitOpsApplicationDeployment`

lib/addons/fluxcd/index.ts
```diff
@@ -61,7 +39,7 @@ export interface FluxCDAddOnProps extends HelmAddOnUserProps {
 
   /** 
   * Flux Kustomization Target Namespace.
-  * Note: when set, this parameter will override all the objects that are part of the Kustomization, please see https://fluxcd.io/flux/components/kustomize/kustomization/#target-namespace */
+  * Default `default` */
   fluxTargetNamespace?: string;
```
shapirov103, line 43: is this namespace ever different for these apps?  my assumption is that this is the namespace where the GitRepository CR will be created, please confirm. If my assumption is correct, this namespace is going to be the same for all of them. 
If it is something else let's discuss.

shapirov103 (commented):
Overall, refactoring makes sense and there should not be first/second class citizen between bootstrap and workload repos  - that consistency improves the design. 
However, I have a couple of comments on the specific design choices for the structure, please take a look.

lib/addons/fluxcd/index.ts
zjaco13: The target namespace is where we want the kustomizations to be deployed.  This can be specified per kustomization, whereas the gitrespositories are deployed in the same namespace that flux is deployed

shapirov103: The name doesn't convey it, should be fixed, e.g. kustomizationNamespace or kustomizationTargetNamespace. 
If there is a particular set of attributes specific to kustomizations then it would make more sense to have an interface for `Kustomization` with path, namespace, etc.

lib/addons/fluxcd/index.ts
```diff
@@ -12,22 +12,12 @@ import { KubernetesManifest } from 'aws-cdk-lib/aws-eks/lib/k8s-manifest';
...
-  bootstrapRepo?: spi.ApplicationRepository;
-
-  /**
-   * Optional values for the bootstrap application. These may contain values such as domain named provisioned by other add-ons, certificate, and other parameters to pass 
-   * to the applications. 
-   */
-  bootstrapValues?: spi.Values,
+export interface FluxCDGitRepo extends Omit<spi.GitOpsApplicationDeployment, "repository">{
 
+    /**
```
elamaran11, line 17: Indentation issues and the comment is incomplete.

lib/addons/fluxcd/index.ts
```diff
@@ -37,25 +27,31 @@ export interface FluxCDGitRepo {
...
-  /** 
-  * Flux Kustomization paths within the repository;
-  * Flux Kustomizations will be created for bootstrapRepo.path and for the paths specified in this array */
-  additionalFluxKustomizationPaths?: string[];
-
-  /** 
-  * Flux Kustomization Prune.
-  * Default `true` */
-  fluxPrune?: boolean;
+    /**
```
elamaran11, line 30: Indentation issues and the comment is incomplete.

elamaran11 (commented):
Few minor comments.

freschri:
Please @zjaco13 confirm the changes have been tested with https://github.com/aws-observability/cdk-aws-observability-accelerator/tree/main/lib/single-new-eks-opensource-observability-construct
thanks

lib/addons/fluxcd/index.ts
freschri: > is this namespace ever different for these apps? my assumption is that this is the namespace where the GitRepository CR will be created, please confirm. If my assumption is correct, this namespace is going to be the same for all of them. If it is something else let's discuss.

I think we might want NOT to make the assumption of a single namespace. E.g. in this case https://github.com/aws-observability/aws-observability-accelerator/blob/main/artifacts/grafana-operator-manifests/eks/infrastructure/amg_grafana-amp-datasource.yaml we use grafana-operator as a namespace, but we might want to add another path, with other manifests related to another operator and so likely another namespace.
I agree on the name. The parameter is finally going to be applied to https://fluxcd.io/flux/components/kustomize/kustomization/#target-namespace and like mentioned elsewhere, it should be optional, to avoid forcing override.

lib/addons/fluxcd/index.ts
```diff
@@ -9,6 +9,51 @@ import { FluxGitRepository } from "./gitrepository";
...
+  * Internal for Flux sync.
+  * Default `5m0s` */
+  fluxSyncInterval?: string;
+
+    /**
+     * List of kustomizations to create */
+    kustomizations?: FluxKustomizationProps[];
+}
+
+export interface FluxKustomizationProps {
```
freschri, line 35: would it make sense to move this interface and the repo interface to the related files?

lib/addons/fluxcd/index.ts
```diff
@@ -9,6 +9,51 @@ import { FluxGitRepository } from "./gitrepository";
 import { FluxKustomization } from "./kustomization";
 import { KubernetesManifest } from 'aws-cdk-lib/aws-eks/lib/k8s-manifest';
 
+/**
+ * Options for a FluxCD repository
+ */
+export interface FluxCDGitRepo extends Omit<spi.GitOpsApplicationDeployment, "repository">{
```
freschri, line 15: we probably need to review namings and make consistent choices, e.g. flux vs fluxcd. also do we want the flux prefix on kustomizations? then use everywhere... e.g. I see kustomizations and fluxKustomizationPath and fluxPrune in Kustomization props...

lib/addons/fluxcd/gitrepository.ts
```diff
@@ -6,23 +6,23 @@ import { setPath } from "../../utils";
  */
 export class FluxGitRepository {
 
-    constructor(private readonly bootstrapRepo: spi.GitOpsApplicationDeployment) {}
+    constructor(private readonly bootstrapRepo: spi.ApplicationRepository) {}
```
freschri, line 9: maybe bootstrap in the name can be removed now?

lib/addons/fluxcd/index.ts
```diff
@@ -9,6 +9,51 @@ import { FluxGitRepository } from "./gitrepository";
...
+    /** 
+    * Flux Kustomization Target Namespace.
+    * Note: when set, this parameter will override all the objects that are part of the Kustomization, please see https://fluxcd.io/flux/components/kustomize/kustomization/#target-namespace */
+    kustomizationTargetNamespace?: string;
+
+    /** 
+    * Flux Kustomization Prune.
+    * Default `true` */
+    fluxPrune?: boolean;
+
```
freschri, line 50: there should be another interval property here for Kustomization. it is different than GitRepository.

lib/addons/fluxcd/index.ts
shapirov103: Flux in the name of the interface is enough to remove the prefix from each field.  The interfaces should be publicly exported, and if they are part of the addon contract, they should stay with the add-on.

lib/addons/fluxcd/index.ts
```diff
@@ -126,75 +135,59 @@ export class FluxCDAddOn extends HelmAddOn {
...
-  fluxKustomizationPaths.map((fluxKustomizationPath, index) => {
-        constructs.push(createKustomization(clusterInfo, bootstrapRepo, fluxcdAddonProps, fluxKustomizationPath, "" + index, fluxKustomization));
+  const kustomizations = fluxcdGitRepo.kustomizations ?? [{fluxKustomizationPath: "."}]
+  const fluxKustomization = new FluxKustomization(fluxcdGitRepo.repository!);
+  kustomizations?.map((kustomization, index) => {
+    kustomization = {...defaultKustomiationProps, ...kustomization}
+    const manifest =fluxKustomization.generate(
+      fluxcdGitRepo.name + "-" + index,
+      fluxcdGitRepo.namespace ?? namespace,
+      fluxcdGitRepo.fluxSyncInterval!,
```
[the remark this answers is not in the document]
elamaran11: I agree to this comment but generally speaking the intervals are coded the same across `GitRepository` and `Kustomization`

docs/addons/fluxcd.md
```diff
@@ -75,27 +81,31 @@ const app = new cdk.App();
...
+        },
+        {
+            name: "podinfo",
+            namespace: undefined,
+            repository: {
+                repoUrl: 'https://github.com/stefanprodan/podinfo',
+                targetRevision: "master",
+            },
+            values: {
+                "region": "us-east-2"
```
elamaran11, line 106: Region can be consistent.

elamaran11 (commented):
Minor comment.

zjaco13:
> Please @zjaco13 confirm the changes have been tested with https://github.com/aws-observability/cdk-aws-observability-accelerator/tree/main/lib/single-new-eks-opensource-observability-construct thanks

@freschri can confirm it works

lib/addons/fluxcd/index.ts
shapirov103: I clearly remember replying to this, but my comment does not show up. We keep contracts together as customers will need all of them exported. so rather than creating dependencies between files, i suggest that we keep all contracts related to the addon either in the index or in a separate file called contracts.ts, types.ts and such.

shapirov103 (approved):
LGTM,  we will need a PR to the main branch

elamaran11:
@zjaco13 Please PR to main

Issue #791, opened 2023-07-26
external-dns: Not filtering to provided HostedZoneID's


DanielAtanasovski:
### Describe the bug

Since [this change](https://github.com/aws-quickstart/cdk-eks-blueprints/pull/739) from the `Bitnami` to the `kubernetes-sigs` helm chart, provided HostedZones are not being filtered to and passed as args to the `external-dns` container like previously.

### Expected Behavior

Previous generated container args:

```YAML
      containers:
        - name: external-dns
          image: docker.io/bitnami/external-dns:0.13.4-debian-11-r14
          imagePullPolicy: "IfNotPresent"
          args:
            # Generic arguments
            - --metrics-address=:7979
            - --log-level=info
            - --log-format=text
            - --zone-id-filter=XXXXXXXXXX
            - --policy=upsert-only
            - --provider=aws
            - --registry=txt
            - --interval=1m
            - --source=service
            - --source=ingress
            # AWS arguments
            - --aws-api-retries=3
            - --aws-zone-type=
            - --aws-batch-change-size=1000
```

### Current Behavior

Current container args:

```YAML
          image: registry.k8s.io/external-dns/external-dns:v0.13.5
          imagePullPolicy: IfNotPresent
          args:
            - --log-level=info
            - --log-format=text
            - --interval=1m
            - --source=service
            - --source=ingress
            - --policy=upsert-only
            - --registry=txt
            - --provider=aws
          ports:
```

Resulting in errors where it fails to list records in an unrelated zone in the same account:
```
level=fatal msg="failed to list resource records sets for zone /hostedzone/SOME_OTHER_ZONE: AccessDenied: User: arn:aws:sts::ACCOUNTID:assumed-role/CLUSTER_ROLE/ is not authorized to perform: route53:ListResourceRecordSets on resource: arn:aws:route53:::hostedzone/SOME_OTHER_ZONE because no identity-based policy allows the route53:ListResourceRecordSets action\n\tstatus code: 403,"
```

### Reproduction Steps

Deploy a cluster with the external-dns addon and pass in the hostedzone from the resource provider.

```TYPESCRIPT

const hostedZoneName = some-zone

const addOn = new blueprints.addons.ExternalDnsAddOn({
    hostedZoneResources: [hostedZoneName]; // can be multiple
});

const blueprint = blueprints.EksBlueprint.builder()
  .addOns(addOn)
[...]

PR #818, opened 2023-08-15
Add zone id filter using external args.

2start:
*Issue #791*

*Description of changes:*
Unlike the `bitnami/external-dns chart`, `kubernetes-sigs/external-dns` did not support the previously used helm value to set zoneIdFilters. I.e. external dns tried to access all hosted zones in the AWS account. This led to a crash because the policies in the `HelmAddOn` just allow access to the specified hosted zones.

*Options to fix this:*

1: Use the `domainFilters` in the `kubernetes-sigs/external-dns` and pass in the correct domains. 
Problems with this approach are that hosted zone ids have to be passed anyways to configure the correct policy. 

2: Just set the appropriate command line flag `--zone-id-filter` for each hosted zone id. The flags can be set by the helm value `extraArgs`.

*How did I test it:*
I built the node module. Then I copied the external dns add-on into an existing project's node_modules to overwrite the previous buggy add-on. Then, I checked if the command line arguments for the external dns container were set properly and if the correct domains are checked by external dns. Both was true. I.e. it works.

Please let me know how you would've tested this. Set up a new test cluster in an AWS account with multiple hosted zones and just specify one hosted zone? Or is there an easier way that I'm missing?

Additional thoughts: I think the perfect user experience would be for the user to just pass in domain names and then to infer the hosted zones. If that's of interest, then I can look into it.

### DIFF
--- a/lib/addons/external-dns/index.ts
+++ b/lib/addons/external-dns/index.ts
@@ -78,15 +78,18 @@ export class ExternalDnsAddOn extends HelmAddOn {
 
         sa.node.addDependency(namespaceManifest);
 
+        // Create a --zone-id-filter arg for each hosted zone
+        const zoneIdFilterArgs = hostedZones.map((hostedZone) => `--zone-id-filter=${hostedZone!.hostedZoneId}`);
+
         let values: Values = {
             provider: "aws",
-            zoneIdFilters: hostedZones.map((hostedZone) => hostedZone!.hostedZoneId),
+            extraArgs: zoneIdFilterArgs,
             aws: {
               region,
             },
             serviceAccount: {
               create: false,
-             name: sa.serviceAccountName,
+              name: sa.serviceAccountName,
             },
         };
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@aws-quickstart/eks-blueprints",
-    "version": "1.10.1",
+    "version": "1.10.2",
     "license": "Apache-2.0",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",


shapirov103:
@2start thank you for your contribution! The fix seems reasonable provided that it works of course. I looked in the helm chart and it was not clear how exactly the extraArgs option is treated by the add-on, I see extraArgs are part of the values [here](https://github.com/kubernetes-sigs/external-dns/blob/master/charts/external-dns/values.yaml#L182) but very obscure what the usage is.

If the issue is addressed we can merge. @parkand1 please take a look, we can test out with the nginx ingress pattern for example: https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/lib/nginx-ingress-construct/index.ts

2start:
Here `extraArgs` is passed as an container argument to external-dns: https://github.com/kubernetes-sigs/external-dns/blob/7ddc9daba70794e7a9fa77e664b5d60e1e8322f6/charts/external-dns/templates/deployment.yaml#L102

Here the flags e.g. `--zone-id-filter` are parsed and passed to the external dns AWS provider: https://github.com/kubernetes-sigs/external-dns/blob/7ddc9daba70794e7a9fa77e664b5d60e1e8322f6/main.go

2start:
Just in case you didn't get a notification @shapirov103 feel free to merge.

shapirov103:
@2start thank you, I saw your notification, all good. We are looking into a functional validation test, as our e2e test does not cover route53 integration at the moment, so it will have to be manual.

parkand1:
Testing in progress will update this thread

parkand1:
I was able to get this up and running with your proposed changes, thank you for your contribution! Good to merge!

parkand1 (approved):
LGTM!

2start:
Awesome, happy to hear that!

shapirov103 (approved):
LGTM

PR #827, opened 2023-08-21
Fixing GH Actions

elamaran11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -2,6 +2,9 @@ name: 'Release Website Publish'
 on:
   release:
     types: [published]
+  push:
+    branches:
+      - ghActions
 permissions:
   contents: write
 jobs:
@@ -19,4 +22,4 @@ jobs:
       - run: pip install mkdocs-material
       - run: npm install
       - run: npx typedoc --out docs/api lib/index.ts
-      - run: mkdocs gh-deploy
+      - run: mkdocs gh-deploy --force
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -4,7 +4,7 @@ on:
     types: [published]
   push:
     branches:
-      - ghActions
+      - fix/ghActions
 permissions:
   contents: write
 jobs:
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -22,4 +22,4 @@ jobs:
       - run: pip install mkdocs-material
       - run: npm install
       - run: npx typedoc --out docs/api lib/index.ts
-      - run: mkdocs gh-deploy --force
+      - run: mkdocs gh-deploy
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -2,9 +2,6 @@ name: 'Release Website Publish'
 on:
   release:
     types: [published]
-  push:
-    branches:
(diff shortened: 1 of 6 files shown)


.github/workflows/docbuild.yml
```diff
@@ -19,4 +22,4 @@ jobs:
       - run: pip install mkdocs-material
       - run: npm install
       - run: npx typedoc --out docs/api lib/index.ts
-      - run: mkdocs gh-deploy
+      - run: mkdocs gh-deploy --force
```
shapirov103, line 25: what is up with --force? I never had to it when deploying from my box.

shapirov103 (commented):
Please see my comment. I am a bit unclear on how to test this since we already released, so I really don't want to run another release. I can manually invoke the action, but unsure if it has any dependency on the event data.

.github/workflows/docbuild.yml
elamaran11: Its not required, i added it to try out to see if any issues. Removed it now.

elamaran11: It works well only if you have `--force` with GH Actions. See it is tested with this PR and works fine.

elamaran11:
> Please see my comment. I am a bit unclear on how to test this since we already released, so I really don't want to run another release. I can manually invoke the action, but unsure if it has any dependency on the event data.

You can test it by adding your branch on Push like i did it with the PR. If you push a commit to this PR, it will test it.

.github/workflows/docbuild.yml
```diff
@@ -2,6 +2,9 @@ name: 'Release Website Publish'
 on:
   release:
     types: [published]
+  push:
+    branches:
```
shapirov103, line 6: do we still need this?

lib/builders/observability-builder.ts
```diff
@@ -1,24 +1,26 @@
-import * as blueprints from '../../lib';
+import * as blueprintbuilder from '../../lib/stacks';
```
shapirov103, line 1: let's replace with `import { BlueprintBuilder } from ...` and just use it as BlueprintBuilder

lib/builders/observability-builder.ts
```diff
@@ -1,24 +1,26 @@
...
+import * as addons from '../../lib/addons';
 import * as utils from "../utils";
+import * as spi from '../../lib/spi'
```
shapirov103, line 4: `make lint` - I see GH actions missing semicolons and such

shapirov103 (commented):
A couple of comments

lib/builders/observability-builder.ts
elamaran11: fixed

lib/builders/observability-builder.ts
elamaran11: fixed

elamaran11:
> A couple of comments

@shapirov103 All comments fixed, pls check

shapirov103 (approved):
LGTM

shapirov103 (approved):

shapirov103:
@elamaran11 can you confirm that local test provisioning and destroying the cluster was successful?

elamaran11:
> @elamaran11 can you confirm that local test provisioning and destroying the cluster was successful?

@shapirov103 Yes both actions are successful in my local test.

PR #834, opened 2023-08-30
Gpu Builder Instance Type prop fix

Howlla:
*Issue #, if available:*

*Description of changes:* Concatenate instanceClass and instanceSize for instanceType prop

### DIFF
--- a/lib/builders/gpu-builder.ts
+++ b/lib/builders/gpu-builder.ts
@@ -158,7 +158,7 @@ function addGpuNodeGroup(options: GpuOptions): clusterproviders.ManagedNodeGroup
     return {
         id: "mng-linux-gpu",
         amiType: NodegroupAmiType.AL2_X86_64_GPU,
-        instanceTypes: [new ec2.InstanceType('g5.xlarge')],
+        instanceTypes: [new ec2.InstanceType(`${options.instanceClass}.${options.instanceSize}`)],
         desiredSize: options.desiredNodeSize, 
         minSize: options.minNodeSize, 
         maxSize: options.maxNodeSize,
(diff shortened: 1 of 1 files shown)


lib/builders/gpu-builder.ts
```diff
@@ -158,7 +158,7 @@ function addGpuNodeGroup(options: GpuOptions): clusterproviders.ManagedNodeGroup
     return {
         id: "mng-linux-gpu",
         amiType: NodegroupAmiType.AL2_X86_64_GPU,
```
elamaran11, line 160: What about `NodegroupAmiType` ? Also please remove the blocksize field as discussed yesterday

elamaran11 (changes requested):
@Howlla Thanks for the fix, couple of minor changes

lib/builders/gpu-builder.ts
Howlla: I think NodegroupAMIType is better off being fixed. removing it altogether as a GPU Option is the way to go

elamaran11: Agree, also remove the blockdevice one

lib/builders/gpu-builder.ts
```diff
@@ -44,10 +40,6 @@ export interface GpuOptions {
      * Optional, Maximum number of nodes to use for the cluster. 
      */
     maxNodeSize?: number,
-    /** 
-     * Optional, Block device size.
-     */
-    blockDeviceSize?: number,
```
elamaran11, line 50: This should be used and passed as paramter to nodegroup

elamaran11 (changes requested):
BlockSize is needed for nodegroup

elamaran11 (commented):
@Howlla Could you please make sure you `tar` the blueprints and test it on patterns. Since there is no e2e for builder classes, we need to test it before we can merge.

Howlla:
Sure

Howlla:
@elamaran11 Tested in the gpu monitoring pattern here https://github.com/aws-samples/cdk-eks-blueprints-patterns/pull/141

elamaran11 (approved):
LGTM

PR #846, opened 2023-09-21
New Decorator for Architecture Type for all Addons and Graviton Builder Classes && Gen AI Builder and Team

elamaran11:
*Issue #, if available:*

*Description of changes:*
Generative AI and Graviton Builder Classes


lib/builders/graviton-builder.ts
```diff
@@ -0,0 +1,172 @@
...
+ * This builder class helps you prepare a blueprint for setting up 
+ * Graviton nodes with EKS cluster. The `GravitonBuilder` creates the following:
+ * 1. An EKS Cluster` with passed k8s version.
+ * 2. A managed graviton nodegroup for ARM64 software.
+ */
+
+export class GravitonBuilder extends BlueprintBuilder {
+
+
+    public addIstioBaseAddOn(props?: addons.IstioBaseAddOnProps) : GravitonBuilder {
```
shapirov103, line 30: Hey Ela, what is the point of adding these addons in the builder? Since this extends the BlueprintBuilder customer can just use `builder.addOns(new IstioBaseAddOn())` and similar. Are you trying to isolate the surface area to only the addons validated on Graviton? 

If that is the case, maybe we can come up with a better approach, like using a decorator to add metadata to each addon. 

For example:

```
@arch(ArchType.X86, ArchType.ARM)
export class IstioBaseAddOn ... {
}
```

This decorator will create a global map of all addons that support ARM. In the graviton builder we can query that map and check if any of the registered addons are not supporting ARM and fail fast. 

We can discuss if needed.

lib/builders/graviton-builder.ts
```diff
@@ -0,0 +1,172 @@
...
+    public static builder(options: GravitonOptions): GravitonBuilder {
+        const builder = new GravitonBuilder();
+
+        builder
+            .clusterProvider(
+                new MngClusterProvider({
+                    version: options.kubernetesVersion,
+                    instanceTypes: [new ec2.InstanceType(`${options.instanceClass}.${options.instanceSize}`)],
+                    amiType: eks.NodegroupAmiType.AL2_ARM_64,
+                    desiredSize: 3,
```
shapirov103, line 135: If we are to move it beyond just a demo, these should be configurable. We need to add something like `withNodeGroupOptions(options: Partial<ManagedNodeGroupProps>)` to the builder to allow any kind of an override. The approach is similar to the one I recommended to Bhavye on his builder.

lib/builders/genai-builder.ts
```diff
@@ -0,0 +1,54 @@
+import { BlueprintBuilder } from '../stacks';
+import * as utils from "../utils";
+import * as addons from '../addons';
+import * as spi from '../spi';
+import { NestedStack, NestedStackProps } from 'aws-cdk-lib';
+import { Construct } from 'constructs';
+
+export class GenAIBuilder extends BlueprintBuilder {
```
shapirov103, line 8: I am unclear on what specifically makes this builder a GenAI builder other than usage tracking? Is there anything that we should add specific to genAI like enable a particular instance types or deploy an example workload  - something?

docs/builders/genai-builder.md
```diff
@@ -0,0 +1,29 @@
...
+### Generative AI on EKS Cluster
+
+The below usage helps you with a demonstration to use `GenAIBuilder` to setup required addons as you prepare a blueprint for setting up Generative AI on a new EKS cluster.
+
+```typescript
+import { Construct } from 'constructs';
+import * as blueprints from '@aws-quickstart/eks-blueprints';
+import { ObservabilityBuilder } from '@aws-quickstart/eks-blueprints';
+
+export default class GenAIShowcase {
```
shapirov103, line 16: What showcase will customer get after deploying this other a vanilla EKS cluster with some addons?

shapirov103 (commented):
@elamaran11 great effort and there is some work we should perform before exposing it to the customers. Please see my comments.

lib/builders/graviton-builder.ts
```diff
@@ -0,0 +1,176 @@
...
+        );
+    }  
+
+    public addXrayAdotAddOn(props?: addons.XrayAdotAddOnProps) : GravitonBuilder {
+        return this.addOns(
+            new addons.XrayAdotAddOn(props)
+        );
+    }
+
+    public addOns(...addOns: spi.ClusterAddOn[]): this {
```
shapirov103, line 125: If we have this method we don't need explicit methods for the addons, like addIstio, etc. If they add an addon that does not support ARM they will get an error telling them. We have docs for reference.

lib/utils/architecture-utils.ts
```diff
@@ -0,0 +1,51 @@
+import "reflect-metadata";
```
shapirov103, line 1: is this needed?

lib/utils/architecture-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Decorator function that accepts a list of architectures and
+ * adds this metatdata to globalmap.
+ * @param arch
+ * @returns 
+ */
+export function arch(...archType: ArchType[]) {
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
shapirov103, line 44: Let remove this line and make the `target: Object` as `target: any` instead. I know it is copied, let's apply to the addon-utils.ts as well.

lib/utils/architecture-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+      return true;
+    }
+    else {
+      return false;
+    }
+  }
+  return archs.includes(arch);
+}
+
+export function validateSupportedArchitecture(addOnName: string, arch: ArchType) : void {
```
shapirov103, line 31: let's think if there is a case when let's say there is a partner addon that does not have any `@arch` annotations. And they do support arm for example. I am think we should provide a workaround to force such cases if needed and document. For example, the builder will have a setting like `strictValidation` which is true by default. If not strict, we will issue warning. This boolean flag can be passed to this function as a parameter from the builder.

shapirov103 (commented):
Looks great. Some comments.
Please also run `make lint`  - there are many warnings

elamaran11:
> Looks great. Some comments. Please also run `make lint` - there are many warnings

Fixed all issues pointed in this iteration.

elamaran11:
@shapirov103 All issues fixed.

lib/builders/graviton-builder.ts
shapirov103: Why `desiredSize, min, max` are left hardcoded and not leveraging the potentially passed value from the options?

shapirov103: My recommendation is to use default MngClusterProviderProps and overwrite them with the supplied. That will be easier to implement and provides the best flexibility. In the current implementation it is more of a demo rather than a production ready builder.

shapirov103 (commented):
@elamaran11  looking great. A couple of comments left, please take a look.

elamaran11:
@shapirov103 GenAI builder is removed, Please review this.

elamaran11:
@shapirov103 Added Gen AI Class and Team. Please check the approach

lib/teams/geneai-team/genai-team.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * @public
+   * @param {GenAITeamProps} props the Gen AI team definition {@link GenAITeamProps}
+   */
+  constructor(props: GenAITeamProps) {
+    super(props);
+    this.genAITeam = props;
+  }
+
+  setup(clusterInfo: ClusterInfo): void {
```
shapirov103, line 44: Does it work? I see this code will clash with the setupServiceAccount method from the application team. I don't think you need to set it up like this at all. All you need to do is pass the serviceaccountname and serviceAccountPolicies to with the props to the Application team and its default setup method should take care of the IRSA stuff.

shapirov103 (commented):
A couple of high level comments for now on the GenAI design.

elamaran11:
@shapirov103 PR is updated with `teams` supporting function as a property to `teamProps` and have also supplied an example which passes function as property for deploying a workload manifest to `teamProps`. Please review this.

lib/builders/bedrock-builder.ts
```diff
@@ -0,0 +1,53 @@
...
+import { NestedStack, NestedStackProps } from 'aws-cdk-lib';
+import { Construct } from 'constructs';
+
+export class BedrockBuilder extends BlueprintBuilder {
+
+    /**
+     * This method helps you prepare a blueprint for setting up EKS cluster with 
+     * usage tracking addon
+     */
+    public static builder(): BedrockBuilder {
```
shapirov103, line 14: aren't we supposed to add a bedrock team here ootb? e.g. take those props with the function and add the team. Or another alternative is to add a method like addBedrockTeam() to make it explicit. 
The way it stands now, there is no differentiator for this builder for the customers.

shapirov103 (commented):
Please see my comment

elamaran11:
@shapirov103 Makesense and i missed this, i fixed it now and all good. Please check and run e2e.

lib/utils/architecture-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+export function supportsALL(constructor: Function) {
+  const addonName = constructor.name;
+  addAddonArch(addonName, ArchType.X86);
+  addAddonArch(addonName, ArchType.ARM);
+}
+
+function addAddonArch(addonName: string, architecture: ArchType) {
+  if (addonArchitectureMap.has(addonName)) {
+    const value = addonArchitectureMap.get(addonName);
+    if (value !== undefined) {
```
shapirov103, line 68: Based on my observation:
this code when executed multiple times will keep adding supported archs to the global array. So if in my app I instantiate an addon 100 times, this will have at least 100 entries. 

Let's add a check to add value only if it is not already there.

shapirov103 (commented):
One small comment. 

because of the quite significant surface area of chages, please test with patterns. Does it work for make test-all?

shapirov103 (approved):
LGTM - great work!

Issue #850, opened 2023-09-29
VpcCniAddOn - enable supports of Kubernetes Network Policies with `enableNetworkPolicy` after Amazon VPC CNI version 1.14.0


bnaydenov:
### Describe the feature

AWS announce on Aug 29, 2023 the native support for enforcing [Kubernetes network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) with [Amazon VPC Container Networking Interface (CNI) Plugin](https://github.com/aws/amazon-vpc-cni-k8s). You can now use Amazon VPC CNI to implement both pod networking and network policies to secure the traffic in your Kubernetes clusters. Native support for network policies has been one of the most requested features on aws containers-roadmap (https://github.com/aws/containers-roadmap/issues/1478).

Network policies will be supported on new Amazon EKS (IPv4 and IPv6) clusters using v1.25 and later. Amazon EKS automatically upgrades all existing clusters to the latest Amazon EKS platform version for their corresponding Kubernetes minor version. 

To enable network policies in your cluster, ensure that your cluster is running Amazon VPC CNI version 1.14.0 or a later version. When creating a new cluster, you may specify Amazon VPC CNI version 1.14.0 or later Update Amazon VPC CNI for clusters not created with Amazon VPC CNI version 1.14.0. Once the platform version update is completed for existing clusters, make sure to update the VPC CNI to the supported version.

The network policy feature is supported for clusters running Kubernetes minor versions 1.25 and above but disabled by default at launch. You can enable Network policy for your cluster using Amazon VPC CNI configuration parameters Enter the JSON key `"enableNetworkPolicy"`: and value `"true"` in Configuration values.

### Use Case

By default, there are no restrictions in Kubernetes for IP addresses, ports, or connections between any Pods in your cluster or between your Pods and resources in any other network. You can use Kubernetes network policy to restrict network traffic to and from your Pods. For more information, see [Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) in the Kubernetes documentation.

If you have version 1.13 or earlier of the Amazon VPC CNI plugin for Kubernetes on your cluster, you need to implement a third party solution to apply Kubernetes network policies to your cluster. Version 1.14 or later of the plugin can implement network policies, so you don't need to use a third party solution. In this topic, you learn how to configure your cluster to use Kubernetes network policy on your cluster without using a third party add-on.

[...]

PR #851, opened 2023-09-29
VpcCniAddOn - enable supports of Kubernetes Network Policies with enableNetworkPolicy after Amazon VPC CNI version 1.14.0 #850

bnaydenov:
*Issue #850 :*

*Description of changes:*

Introduce enableNetworkPolicy parameter and when is set to 'true' enable Network policy with AWS VPC-CNI plugin.

For example:
```typescript
    const addOns: Array<blueprints.ClusterAddOn> = [
      new blueprints.addons.CoreDnsAddOn(),
      new blueprints.addons.KubeProxyAddOn(),
      new blueprints.addons.EbsCsiDriverAddOn(),
      new blueprints.addons.VpcCniAddOn({ enableNetworkPolicy: 'true', version: 'v1.15.0-eksbuild.2' })
    ];
```

### DIFF
--- a/lib/addons/vpc-cni/index.ts
+++ b/lib/addons/vpc-cni/index.ts
@@ -225,14 +225,14 @@ export interface VpcCniAddOnProps {
   * `MAX_ENI` Environment Variable. Format integer.
   * Specifies the maximum number of ENIs that will be attached to the node. 
   */
-  maxEni?: number;  
+  maxEni?: number;
 
   /**
   * `MINIMUM_IP_TARGET` Environment Variable. Format integer.
   * Specifies the number of total IP addresses that the ipamd 
   * daemon should attempt to allocate for pod assignment on the node.
   */
-  minimumIpTarget?: number; 
+  minimumIpTarget?: number;
 
   /**
    * `POD_SECURITY_GROUP_ENFORCING_MODE` Environment Variable. Type: String. 
@@ -274,6 +274,13 @@ export interface VpcCniAddOnProps {
    * 
    */
   serviceAccountPolicies?: iam.IManagedPolicy[];
+
+  /**
+   * Enable kubernetes network policy in the VPC CNI introduced in vpc-cni 1.14
+   * More informaton on official AWS documentation: https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html
+   * 
+   */
+  enableNetworkPolicy?: string;
 
   /**
    * Version of the add-on to use. Must match the version of the cluster where it
@@ -410,14 +417,15 @@ function populateVpcCniConfigurationValues(props?: VpcCniAddOnProps): Values {
       POD_SECURITY_GROUP_ENFORCING_MODE: props?.podSecurityGroupEnforcingMode,
       WARM_ENI_TARGET: props?.warmEniTarget,
       WARM_IP_TARGET: props?.warmIpTarget,
-      WARM_PREFIX_TARGET: props?.warmPrefixTarget
-    }
+      WARM_PREFIX_TARGET: props?.warmPrefixTarget,
+    },
+    enableNetworkPolicy: props?.enableNetworkPolicy
   };
 
   // clean up all undefined
   const values = result.env;
   Object.keys(values).forEach(key => values[key] === undefined ? delete values[key] : {});
-  Object.keys(values).forEach(key => values[key] = typeof values[key] !== 'string' ?  JSON.stringify(values[key]):values[key]);
- 
+  Object.keys(values).forEach(key => values[key] = typeof values[key] !== 'string' ? JSON.stringify(values[key]) : values[key]);
+
   return result;
 }
--- a/lib/addons/vpc-cni/index.ts
+++ b/lib/addons/vpc-cni/index.ts
@@ -299,7 +299,7 @@ export interface CustomNetworkingConfig {
 
 const defaultProps: CoreAddOnProps = {
   addOnName: 'vpc-cni',
-  version: 'v1.13.4-eksbuild.1',
+  version: 'v1.14.1-eksbuild.1',
   saName: 'aws-node',
   namespace: 'kube-system',
   controlPlaneAddOn: false,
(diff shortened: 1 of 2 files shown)


bnaydenov:
Running cdk diff with this changes produces this diff:

Screenshot 2023-09-29 at 10 31 43 AM

And when cdk deploy is done VPC-CNI is run with enabled Network policy. How to check:

```bash
kubectl get pods -n kube-system | grep 'aws-node\|amazon'
```

An example output is as follows.
```bash
aws-node-gmqp7     2/2     Running   1 (1h ago)   1h
aws-node-prnsh     2/2     Running   1 (1h ago)   1h
```

If network policy is enabled, there are 2 containers in the aws-node pods. In previous versions and if network policy is disabled, there is only a single container in the aws-node pods.

More info here: [https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html](https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html)

elamaran11 (commented):
LGTM. @shapirov103 Please review and run `e2e`

bnaydenov:
@shapirov103 please check this MR when you have time.

lib/addons/vpc-cni/index.ts
```diff
@@ -292,7 +299,7 @@ export interface CustomNetworkingConfig {
 
 const defaultProps: CoreAddOnProps = {
   addOnName: 'vpc-cni',
-  version: 'v1.13.4-eksbuild.1',
+  version: 'v1.14.1-eksbuild.1',
```
shapirov103, line 302: Based on https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html the version should be v1.15.0-eksbuild.2 , however incremental update is needed. 

However, it looks like the network policy will not be available with this version, is that correct? If so, can we add a check to fail if version is less than 1.15 so that we don't get questions from the customers on why my network policy dos not work?

The error should state that customers should apply incremental update and use network policy starting from 1.15. 

I will need to include it in the release notes, so if you can update the PR dsc with this instructions as well that would help.

shapirov103 (commented):
@bnaydenov awesome work. Please my comment, it is mostly release process related and docs.

bnaydenov:
@shapirov103 Network policy is available on Amazon VPC CNI version 1.14.0 or a later version. 

I am setting here as default current latest version for major version `1.14.x`. Currently there is version `v1.15.0-eksbuild.2` , but if you set it straight to `v1.15.0-eksbuild.2` update fails with following error:

<code>Resource handler returned message: "Updating VPC-CNI can only go up or down 1 minor version at a time 
</code> 

And you have to make 2 step update, first to `v1.14.1-eksbuild.1`, and then from `v1.14.1-eksbuild.1` to `v1.15.0-eksbuild.2`.

**Step 1**
```typescript
const addOns: Array<blueprints.ClusterAddOn> = [
      new blueprints.addons.VpcCniAddOn({ enableNetworkPolicy: 'true', version: 'v1.14.1-eksbuild.1' })
 ];
```
```bash
cdk deploy
```

**Step 2**
```typescript
const addOns: Array<blueprints.ClusterAddOn> = [
      new blueprints.addons.VpcCniAddOn({ enableNetworkPolicy: 'true', version: 'v1.15.0-eksbuild.2' })
];
```
```bash
cdk deploy
```


@shapirov103 Tell me what you think about this? Should we set default to `1.14.x` or straight to `1.15.x`. We can include in release notes a statement above 2 steps upgrade process is necessary .

shapirov103:
> @shapirov103 Tell me what you think about this? Should we set default to `1.14.x` or straight to `1.15.x`. We can include in release notes a statement above 2 steps upgrade process is necessary .

@bnaydenov if 14.1 works with network policy, then I agree with your decision to use 1.14 at the moment to avoid inadvertent deployment failures in customers (and AWS SDE internal) environments due to the version skip. Let's release with 1.14 and mention 1.15 in the docs so that they can upgrade in a two step process. 

I will upgrade the blueprints to use 1.15 in the next minor release.

bnaydenov:
ok let's go with `v1.14.1-eksbuild.1` then. I will leave this PR as it is.

lib/addons/vpc-cni/index.ts
```diff
@@ -275,6 +275,13 @@ export interface VpcCniAddOnProps {
    */
   serviceAccountPolicies?: iam.IManagedPolicy[];
 
+  /**
+   * Enable kubernetes network policy in the VPC CNI introduced in vpc-cni 1.14
+   * More informaton on official AWS documentation: https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy.html
+   * 
+   */
+  enableNetworkPolicy?: string;
```
shapirov103, line 283: Minor comment: why is it a string? Let's make it a boolean, then it is clear and we don't have to handle cases like "FALSE" vs "false".

shapirov103 (commented):
One minor comment

bnaydenov:
it must be a `string` not a `boolean`, because there is a JSON scheme about the vpc-cni addon configuration. And this scheme says it must be `string`. When you set it to boolean it gives error when you do `cdk synth` 

Check here JSON scheme .

```json
"enableNetworkPolicy": {
          "format": "boolean",
          "type": "string"
        },
```

shapirov103:
> it must be a `string` not a `boolean`, because there is a JSON scheme about the vpc-cni addon configuration. And this schema says it must be `string`. When you set it to boolean it gives error when you do `cdk synth`
> 

@bnaydenov in the addon it should be a boolean as this is the best experience for the customer. Please check the rest of the  properties in the `VpcCniAddOnProps` which are declared as `boolean`.

bnaydenov:
@shapirov103 check my latest [commit](https://github.com/aws-quickstart/cdk-eks-blueprints/pull/851/commits/f9a5b6c72860aca5916dbbd59ceb89400552c52f). 

now `enableNetworkPolicy ` is `boolean` and `JSON.stringify()` is used

bnaydenov:
@shapirov103 any more comments or concerns?

shapirov103 (approved):
LGTM

Issue #863, opened 2023-10-23
AMP Collector Config needs to be Upgrade for AMP Addon


elamaran11:
### Describe the feature

Amp Connector [Config](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/amp/collector-config-amp.ytpl) needs following update. 

```yaml
  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.33.1
  resources:
    limits:
      cpu: "1"
      memory: "2Gi"
    requests:
      cpu: "1"
      memory: "2Gi"
```

### Use Case

Upgrade AMP Collector Config.

### Proposed Solution

Amp Connector [Config](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/amp/collector-config-amp.ytpl) needs following update. 

```yaml
  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.33.1
  resources:
    limits:
      cpu: "1"
      memory: "2Gi"
    requests:
      cpu: "1"
      memory: "2Gi"
```

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.99.0

### EKS Blueprints Version

1.12.0

### Node.js Version

20.0.0

### Environment details (OS name and version, etc.)

Mac

elamaran11:
@ratnopamc Can work on this with Observability Accelerator PR.

elamaran11:
This is completed. Closing the ticket.

PR #868, opened 2023-10-31
update AMP collector configuration

ratnopamc:
*Issue #, if available:*
#863 
*Description of changes:*
Updated AMP collector configuration to use latest otel collector image and also set request and limits.

### DIFF
--- a/lib/addons/amp/collector-config-amp.ytpl
+++ b/lib/addons/amp/collector-config-amp.ytpl
@@ -9,7 +9,14 @@ metadata:
   namespace: "{{namespace}}"
 spec:
   mode: "{{deploymentMode}}"
-  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.21.0
+  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.33.1
+  resources:
+    limits:
+      cpu: "1"
+      memory: "2Gi"
+    requests:
+      cpu: "1"
+      memory: "2Gi"  
   serviceAccount: adot-collector
   config: |
     receivers:


elamaran11 (commented):
LGTM. @shapirov103 Please check and run e2e.

elamaran11:
@shapirov103 Could we run e2e here?

lib/addons/amp/collector-config-amp.ytpl
```diff
@@ -9,7 +9,14 @@ metadata:
   namespace: "{{namespace}}"
 spec:
   mode: "{{deploymentMode}}"
-  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.21.0
+  image: public.ecr.aws/aws-observability/aws-otel-collector:v0.33.1
+  resources:
+    limits:
+      cpu: "1"
+      memory: "2Gi"
```
shapirov103, line 16: Should not this be configurable? Do you foresee cases that this particular constraint won't work in production for larger volumes?

shapirov103 (commented):
Please see my question.

lib/addons/amp/collector-config-amp.ytpl
elamaran11: This could be made configurable but this configuration is for collector so we dont run in to constraint in production also this is the config received from service team.

shapirov103: I am not questioning the need for the settings, I actually like them. We can apply for now, but these are hard coded and there will be no way for customers to override them if needed. Let's wait for the customer signals ;)

shapirov103 (approved):

Howlla:
Ran e2e for aws  observability accelerator Graviton and OSS Patterns with latest otel collector. works as intended

PR #890, opened 2023-12-14
feat: adding support for karpenter blockdevicemapping

paulmowat:
*Description of changes:*

Adding in support to pass blockDeviceMappings through to karpenter as per https://karpenter.sh/v0.30/concepts/node-templates/#specblockdevicemappings

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -14,6 +14,34 @@ import { Rule } from 'aws-cdk-lib/aws-events';
 import { SqsQueue } from 'aws-cdk-lib/aws-events-targets';
 import { Cluster } from 'aws-cdk-lib/aws-eks';
 
+enum VolumeType {
+    STANDARD = 'standard',
+    IO1 = 'io1',
+    IO2 = 'io2',
+    GP2 = 'gp2',
+    SC1 = 'sc1',
+    ST1 = 'st1',
+    GP3 = 'gp3',
+}
+
+interface BlockDeviceMapping {
+    deviceName?: string;
+    virtualName?: string;
+    ebs?: Ebs;
+    noDevice?: string;
+}
+  
+interface Ebs {
+    deleteOnTermination?: boolean;
+    iops?: number;
+    snapshotId?: string;
+    volumeSize?: number;
+    volumeType?: VolumeType;
+    kmsKeyId?: string;
+    throughput?: number;
+    outpostArn?: string;
+    encrypted?: boolean;
+}
 
 /**
  * Configuration options for the add-on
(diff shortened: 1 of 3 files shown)


test/karpenter.test.ts
```diff
@@ -114,42 +114,81 @@ describe('Unit tests for Karpenter addon', () => {
...
+    
+        const stack = await blueprint
+          .version("auto")
+          .account("123567891")
+          .region("us-west-1")
+          .addOns(
+            new blueprints.KarpenterAddOn({
+              version: 'v0.29.2',
+              values: {
+                blockDeviceMappings: [{
```
pennywisdom, line 169: I think this incorrectly setting the helm chart values instead of the addon blockDeviceMappings property in the test.

paulmowat: I've fixed this now.

lib/addons/karpenter/index.ts
```diff
@@ -13,7 +13,26 @@ import * as sqs from 'aws-cdk-lib/aws-sqs';
...
+import { EbsDeviceVolumeType } from 'aws-cdk-lib/aws-ec2';
 
+interface BlockDeviceMapping {
+    deviceName?: string;
+    virtualName?: string;
+    ebs?: Ebs;
+    noDevice?: string;
+}
+  
+interface Ebs {
```
shapirov103, line 25: Ebs is a bit too bold for this interface  as it appears to represent ElasticBlockStorage as a service. How about either "EbsVolumeSettings` | `EbsVolumeMappings` or just `EbsMappings`?

lib/addons/karpenter/index.ts
```diff
@@ -13,7 +13,26 @@ import * as sqs from 'aws-cdk-lib/aws-sqs';
 import { Rule } from 'aws-cdk-lib/aws-events';
 import { SqsQueue } from 'aws-cdk-lib/aws-events-targets';
 import { Cluster } from 'aws-cdk-lib/aws-eks';
+import { EbsDeviceVolumeType } from 'aws-cdk-lib/aws-ec2';
 
+interface BlockDeviceMapping {
```
shapirov103, line 18: These should be exported along with the `KarpenterAddOnProps` (not part of the PR atm) as customers will have issue consuming these otherwise.  Let's add `export` keyword to all public data types.

shapirov103 (commented):
@paulmowat thank you for this contribution. Please see a couple relatively minor comments.

lib/addons/karpenter/index.ts
paulmowat: thanks for the feedback. I've renamed to EBSVolumeMapping

lib/addons/karpenter/index.ts
paulmowat: thanks for the feedback. I've exported all the interfaces (including KarpenterAddOnProps) and now use these as part of the unit test.

shapirov103 (approved):
LGTM

shapirov103:
Validated with unit tests, e2e will be used with the upgraded versions of helm/cdk in the 1.13 release prep.

PR #928, opened 2024-02-13
Upgrade Istio Suite of Addons with CNI and Ingress Gateway Addons.

elamaran11:
*Issue #, if available:*

*Description of changes:*


lib/addons/istio-addons/istio-cni.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ClusterInfo } from "../../spi";
+import { HelmAddOn, HelmAddOnProps } from "../helm-addon";
+import { dependable, supportsALL } from '../../utils';
+
+const defaultProps: HelmAddOnProps = {
+    name: 'istio-cni',
+    release: 'cni',
+    namespace: 'istio-system',
+    chart: 'cni',
+    version: "1.20.1",
```
shapirov103, line 11: let's have a central point for the version. e.g. `export const VERSION` and use that in each istio addon when setting version in the defaultProps, so that we can set it once for all, but customers can still override it per addon.

lib/addons/istio-addons/istio-cni.ts
```diff
@@ -0,0 +1,27 @@
...
+    values: {}, 
+};
+
+@supportsALL
+export class IstioCniAddon extends HelmAddOn {
+
+    constructor() {
+        super({...defaultProps});
+    }
+    @dependable('IstioBaseAddOn','IstioControlPlaneAddOn')
```
shapirov103, line 22: minor: line break between methods

lib/addons/istio-addons/istio-cni.ts
```diff
@@ -0,0 +1,27 @@
...
+    chart: 'cni',
+    version: "1.20.1",
+    repository: 'https://istio-release.storage.googleapis.com/charts',
+    values: {}, 
+};
+
+@supportsALL
+export class IstioCniAddon extends HelmAddOn {
+
+    constructor() {
```
shapirov103, line 19: why is it empty? How will customers accomplish the following use cases:

1. Pass values to this helm chart from the blueprint. 
2. Discover what values are possible to pass. Generally we put most popular choices to the props object.
3. Override the version of helm chart to be anything other than the hard-coded 1.20.1. e.g. I am sure once 1.20.2 is out they will want to use that. 

Applies to all new addons added in the PR.

docs/addons/istio-cni.md
```diff
@@ -0,0 +1,61 @@
...
+
+***IMPORTANT***:
+
+1. This add-on depends on [Istio Base](istio-base.md) and [istio Control Plane](istio-control-plane.md) Add-ons for cluster-wide resources and CRDs.
+
+***Istio Base add-on and Istio Control Plane addon-on must be present in add-on array*** and ***must be in add-on array before the Istio Cni add-on*** for it to work, as shown in below example. Otherwise will run into error `Assertion failed: Missing a dependency for IstioBaseAddOn`.
+
+## Usage
+
+Add the following as an add-on to your main.ts file to add Istio Control Plane to your cluster
```
shapirov103, line 13: let's figure out what default settings are applied based on the helm chart. I assume that this addon is not replacing VPC-CNI - if so we need to call it out. 
I see that istio cni addon has the setting values.cni.chained. I assume it must be true for it to work on AWS. Let's call it out and make it configurable.

lib/addons/istio-addons/istio-ingress-gateway.ts
```diff
@@ -0,0 +1,27 @@
...
+    release: 'ingressgateway',
+    namespace: 'istio-system',
+    chart: 'gateway',
+    version: "1.20.1",
+    repository: 'https://istio-release.storage.googleapis.com/charts',
+    values: {},
+};
+
+@supportsALL
+export class IstioIngressGatewayAddon extends HelmAddOn {
```
shapirov103, line 17: this should be configurable to allow at least to specify host and port based on https://istio.io/latest/docs/tasks/traffic-management/ingress/ingress-control/

shapirov103 (commented):
Added a few comments, I think it is a great start but we will need to solve the use cases I mentioned in the comments.

lib/addons/istio-addons/istio-cni.ts
elamaran11: Fixed

lib/addons/istio-addons/istio-cni.ts
elamaran11: Fixed

docs/addons/istio-cni.md
elamaran11: First part is done, unclear on the second part, lets discuss on this design!

lib/addons/istio-addons/istio-ingress-gateway.ts
elamaran11: Likewise as previous comment!

elamaran11: Taken Care!

docs/addons/istio-cni.md
elamaran11: Done.

elamaran11:
> Added a few comments, I think it is a great start but we will need to solve the use cases I mentioned in the comments.

@shapirov103 All Comments fixed, resolved, please double check and let me know.

shapirov103 (approved):
LGTM, great work!

Issue #930, opened 2024-02-14
ADOT: intermittent failure namespace not found


shapirov103:
### Describe the bug

ADOT addon creation fails intermittently.

### Expected Behavior

ADOT addon installs with no issues

### Current Behavior

Creation of stack fails with the error message:

Custom::AWSCDK-EKS-KubernetesResource | blueprint-construct-dev/adot-collector-sa/manifest-adot-collector-saServiceAccountResource/Resource/Default (blueprintconstructdevadotcollectorsamanifestadotcollectorsaServiceAccountResourceA34512D7) Received response status [FAILED] from custom resource. Message returned: Error: **'Error from server (NotFound): error when creating "/tmp/manifest.yaml": namespaces "adot" not found\n'**

### Reproduction Steps

Reproduction is difficult since it is a race condition. 
https://github.com/aws-quickstart/cdk-eks-blueprints/pull/777#issuecomment-1943942642

### CDK CLI Version

2.115

### Node.js Version

20.,x

### Environment details (OS name and version, etc.)

Mac

PR #931, opened 2024-02-14
Bugfix: ADOT fails intermittently 

shapirov103:
*Issue #, if available:* #930

*Description of changes:*
Refactored core addon and adot addon to handle proper namespace dependencies. 

### DIFF
--- a/lib/addons/adot/index.ts
+++ b/lib/addons/adot/index.ts
@@ -1,5 +1,5 @@
 import { KubernetesManifest } from "aws-cdk-lib/aws-eks";
-import { Construct } from 'constructs';
+import { Construct, IConstruct } from 'constructs';
 import { ClusterInfo } from "../../spi";
 import { createNamespace, dependable, loadYaml, readYamlDocument, supportsALL } from "../../utils";
 import { CertManagerAddOn } from "../cert-manager";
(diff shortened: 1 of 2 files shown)


lib/addons/adot/index.ts
```diff
@@ -47,18 +47,27 @@ export class AdotCollectorAddOn extends CoreAddOn {
...
+        return addOnPromise;
+    }
 
+    /**
+     * Overriding base class method to create namespace and register permissions.
+     * @param clusterInfo 
+     * @param name 
+     * @returns 
+     */
+    createNamespace(clusterInfo: ClusterInfo, namespaceName: string): IConstruct | undefined {
```
elamaran11, line 67: The name `createNamespace` is misleading because we also deploying OTEL permission Manfest here. Can we move that out?

elamaran11 (commented):
LGTM. Minor feedback on the fix.

lib/addons/adot/index.ts
shapirov103: Not for ADOT. I am using a pattern called template method. A noop (abstract) method named `createNamespace` is defined in the core addon. Core addon uses this method to create namespace. ADOT overrides this method to create a namespace and handle other manifest that must be created with the namespace. Since the base template method is defined in the core addon, it has to use a generic name.

elamaran11: Makes total sense i saw that, I would rather rename the `createNamespace` to something meaningful like `createAddonResources`. But approach is good!

shapirov103: I see your logic, but i would rather keep the name of the method as `createNamespace` to address 99% of cases and override it for adot which has a design peculiarity that adds extra compexity, than make the name like createAddonResources which should probably return an array of constructs.

elamaran11 (approved):
@shapirov103 Approach LGTM

Issue #859, opened 2023-10-18
[Karpenter-AddOn] Remove AmazonEKS_CNI_Policy if VPC-CNI AddOn enabled with serviceAccountPolicies defined


vumdao:
### Describe the feature

In order for the Amazon VPC CNI to manage IP addresses, it requires AWS Identity and Access Management ([AWS IAM](https://docs.aws.amazon.com/iam/?icmpid=docs_homepage_security)) permissions. Amazon EKS recommends creating a separate AWS IAM role with the permissions defined in [AmazonEKS_CNI_Policy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEKS_CNI_Policy.html) managed policy then associate that role with the VPC CNI using [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (i.e., AWS IAM roles for service account).

When specified, the node role for the cluster provider can be configured without the CNI policy (Ref: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/vpc-cni/index.ts#L268). This help the node role more least privilege

### Use Case

My addons
```
    new VpcCniAddOn({
      enableNetworkPolicy: true,
      serviceAccountPolicies: [ ManagedPolicy.fromAwsManagedPolicyName('AmazonEKS_CNI_Policy')]
    }),
    new KarpenterAddOn()
```

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.101.1

### EKS Blueprints Version

1.12.0

### Node.js Version

v20.8.1

### Environment details (OS name and version, etc.)

macOS Sonoma 14.0

shapirov103:
@youngjeong46 please take a look. Since we are hardcoding the policies for the node role in the Karpenter addon, let's 1/ create a mechanism to override the behavior with custom policies and 2/ consider adding a flag to exclude CNI policy. 

We should not make the Karpenter addon dependent on the VPC CNI but we can make it configurable to satisfy the customer use case.

shapirov103:
Seems reasonable to go with approach 1/ and see what customer experience if the use case is to just exclude a specific policy. 

For pipeline build policies for example, I used DEFAULT BUILD policies here https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/pipelines/code-pipeline.ts#L200 There can be several sets of policies defined that customers can pick and choose. 

[...]

Issue #924, opened 2024-02-12
[Karpenter] Update for 0.34 Beta CRDs


jsamuel1:
### Describe the feature

Karpenter 0.33 is an interim version, with Karpenter 0.34 requiring new Beta CRD's, to replace the Alpha CRD's.
Generated resources for Provisioner/etc will need to generate Beta versions of the resources.

Upgrades will need to import the new CRD's.

### Use Case

Want to use features from the latest Karpenter

### Proposed Solution

Regenerate all resources according to the Beta schema.
Force-import the CRD's if upgrading from an Alpha version -- or use the karpenter-crd helm chart.

https://karpenter.sh/docs/upgrading/upgrade-guide/

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.115.0

### EKS Blueprints Version

1.13

### Node.js Version

20

### Environment details (OS name and version, etc.)

any :)

jsamuel1:
Note:  Karpenter compatibility matrix says that to use Karpenter on EKS 1.29, the upgrade to 0.34 is required.
https://karpenter.sh/v0.34/upgrading/compatibility/#compatibility-matrix

Issue #893, opened 2023-12-17
[Karpenter] Cannot configure node role policy in KarpenterAddOn


Shellmode:
### Describe the feature

Since the policies of node role and service account are hardcode, there is no way to configure the policy of nodes created by Karpenter in cdk-eks-blueprints.

I can only modify the policy in another way which is not so IaC.

### Use Case

Define policy of nodes created by Karpenter as needed.

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.105.0

### Node.js Version

v20.9.0

### Environment details (OS name and version, etc.)

MacOS 14.1.2 (23B92)

youngjeong46:
@Shellmode I'm trying to understand this better.

1. There are two roles in question - Node Role that gets attached to the Karpenter nodes, and the role for the controller that gets attached to the service account, and which the karpenter controller uses to provision instances. Are you trying to replace both with your custom roles? What would be the reasoning?
2. Would your ideal feature here be to 1/ provide additional policy on top of what is being generated, or 2/ provide a new role all together (which you as a user would manage to make sure it has the minimal required policy to operate Karpenter functionally)?

Shellmode:
> @Shellmode I'm trying to understand this better.
> 
> 
> 
> 1. There are two roles in question - Node Role that gets attached to the Karpenter nodes, and the role for the controller that gets attached to the service account, and which the karpenter controller uses to provision instances. Are you trying to replace both with your custom roles? What would be the reasoning?
> 
> 2. Would your ideal feature here be to 1/ provide additional policy on top of what is being generated, or 2/ provide a new role all together (which you as a user would manage to make sure it has the minimal required policy to operate Karpenter functionally)?

First question, the role of the nodes which are scaled in or scaled out by karpenter.

It's a common situation that worker nodes(karpenter scale them in or out) need various of permissions to finish tasks.

Second question, I think additional policy will be better, because developers always focus on the permissions and don't want to define another role resource.

jsamuel1:
+1.  Hitting this now, when trying to configure CloudwatchInsightsAddon with Karpenter.  No easy way to configure the extra permissions needed.
Ideally, I just want to either pass in the extra policies, or expose the role out from the addon, so that I can add to it after creation.

yubingjiaocn:
[...]

PR #940, opened 2024-03-01
[Karpenter Add-on][BREAKING CHANGE] Update for v0.34.0

youngjeong46:
*Issue #, if available:* #924

*Description of changes:*
This PR changes the Karpenter add-on to support up to v0.34.0 (v0.35.0 just released as of the day of the PR, so it will be following as another PR). This is considered a breaking change:
- Refactored properties for the add-on to better reflect the parameters needed for the NodePool and the EC2NodeClass. 
- Ability to deploy the CRDs based on the versions: v0.32.x and up will enforce v1beta1 CRDs (NodePool and EC2NodeClass), while below versions will enforce v1alpha5 CRDs (Provisioner and AWSNodeTemplate)
- Both parameters are optional to be used for either the alpha or the beta CRDS. However, there are error checks implemented in case the wrong one is used (i.e. using `consolidation` with v1beta1 CRDs or `disruption` with v1alpha5 CRDs)
- Addon will no longer support any versions below v0.21.x, as provided in the compatibility matrix from the official website [here](https://karpenter.sh/docs/upgrading/compatibility/). Version Map is used to issue a warning if incompatible version is used for the Kubernetes version, but will proceed with deployment.
- Additional unit tests
- Doc revisions


elamaran11:
@youngjeong46 Nice PR, lof of work. Does this also fix #859 and #893 Issues?

youngjeong46:
> @youngjeong46 Nice PR, lof of work. Does this also fix #859 and #893 Issues?

No, not yet. I'm trying to figure out the exact ask (at least for 893). Given this PR, it would most likely be a fast follow (along with 0.35 update that came out last night).

docs/addons/karpenter.md
```diff
@@ -16,66 +16,63 @@ Karpenter works by:
...
   .addOns(karpenterAddOn)
   .build(app, 'my-stack-name');
 ```
 
-The add-on automatically sets the following Helm Chart [values](https://github.com/aws/karpenter/tree/main/charts/karpenter#values), and it is **highly recommended** not to pass these values in (as it will result in errors):
-- settings.aws.defaultInstanceProfile
-- settings.aws.clusterEndpoint
-- settings.aws.clusterName
-- settings.aws.interruptionQueueName (if interruption handling is enabled)
+The add-on automatically sets the following Helm Chart [values](https://github.com/aws/karpenter-provider-aws/tree/main/charts/karpenter#values), and it is **highly recommended** not to pass these values in (as it will result in errors):
```
shapirov103, line 73: why would passing the values result in errors? if the values are correct, it should succeed.

docs/addons/karpenter.md
```diff
@@ -94,57 +91,88 @@ blueprints-addon-karpenter-54fd978b89-hclmp   2/2     Running   0          99m
...
+4. Deploys Karpenter helm chart in the `karpenter` namespace, configuring the cluster name, endpoint, Instance Profile, and others necessary for functional addon.
+5. If the user provides NodePoolSpec (and EC2NodeClassSpec), the addon will provisions a default Karpenter NodePool and EC2NodeClass CRDs. NodePoolSpec requires [requirements](https://karpenter.sh/docs/concepts/nodepools/#spectemplatespecrequirements) while EC2NodeClassSpec requires subnets and security groups. Based on what version of Karpenter you provide, you will need either `subnetSelector` and `securityGroupSelector` (for versions v0.31.x or down), or `subnetSelectorTerms` and `securityGroupSelectorTerms` (for versions v0.32.x and up).
```
shapirov103, line 95: these attributes should be lower-case similar to the rest of the attrbutes based on the naming conventions: attributes/functions lower case, types and class name upper case.

docs/addons/karpenter.md
```diff
@@ -94,57 +91,88 @@ blueprints-addon-karpenter-54fd978b89-hclmp   2/2     Running   0          99m
...
-5. Interruption Handling, which is a native way to handle interruption due to involuntary interruption events, is supported on versions 0.19.0 and later. For interruption handling in the earlier versions, Karpenter supports using AWS Node Interruption Handler (which you will need to add as an add-on and ***must be in add-on array after the Karpenter add-on*** for it to work.
-6. Karpenter allows overrides of the default "Name" tag but does not allow overrides to restricted domain (such as "karpenter.sh", "karpenter.k8s.aws", and "kubernetes.io/cluster").
+1. EKS Blueprints will only support minimum Karpenter version that matches the supporter EKS Kubernetes version. Please see the compatibility matrix [here](https://karpenter.sh/docs/upgrading/compatibility/). If you provide incompatible version (i.e. providing version 0.27.x for EKS version 1.27), you will see warnings in the logs but will proceed deployment. You will run into compatibility issues.
+2. EKS Blueprints npm v1.13.x introduces the following breaking changes:
```
shapirov103, line 100: this will 1.14 and above.

docs/addons/karpenter.md
```diff
@@ -213,35 +241,6 @@ Karpenter, starting from the OCI registry versions, will untar the files under `
...
-
-3. Starting with v0.17.0, Karpenter's Helm chart package is stored in OCI (Open Container Initiative) registry. With this change, [charts.karpenter.sh](https://charts.karpenter.sh/) is no longer updated to preserve older versions. You have to adjust for the following:
-
-* The full URL needs to be present (including 'oci://').
-* You need to append a `v` to the version number (i.e. v0.17.0, not 0.17.0)
+1. Certain upgrades require reapplying the CRDs since Helm does not maintain the lifecycle of CRDs. Please see the [official documentations](https://karpenter.sh/docs/upgrading/upgrade-guide/) for details.
 
-4. Starting with v0.22.0, Karpenter will no longer work on Kubernetes version prior to 1.21. Either upgrade your Kubernetes to 1.21 or later version and apply Karpenter, or use prior Karpenter versions.
+2. Starting v0.32.0, Karpenter introduces the new beta APIs (v1beta1). While you can use both v1beta1 and v1alpha5 CRDs for v0.32, starting in v0.33, v1alpha5 CRDs are obsolete. If you want to upgrade to the latest, ensure that you have upgraded to v0.32.x first before proceeding.
```
shapirov103, line 246: trying to figure out exact upgrade steps for the current blueprint customers. Does this paragraph implies that current customer should:
1. upgrade to 0.32.x 
2. upgrade to the latest

Are the any other steps? My assumption is that CRDs will be upgrade (new set of CRDs based on the versoin) and the above approach should work.

docs/addons/karpenter.md
```diff
@@ -213,35 +241,6 @@ Karpenter, starting from the OCI registry versions, will untar the files under `
...
-
-The property is changed to align with the naming convention of the provisioner, and to allow multiple operators (In vs NotIn). The values correspond similarly between the two, with type change being the only difference.
-
-2. Certain upgrades require reapplying the CRDs since Helm does not maintain the lifecycle of CRDs. Please see the [official documentations](hhttps://karpenter.sh/docs/upgrading/upgrade-guide/) for details.
-
-3. Starting with v0.17.0, Karpenter's Helm chart package is stored in OCI (Open Container Initiative) registry. With this change, [charts.karpenter.sh](https://charts.karpenter.sh/) is no longer updated to preserve older versions. You have to adjust for the following:
-
-* The full URL needs to be present (including 'oci://').
-* You need to append a `v` to the version number (i.e. v0.17.0, not 0.17.0)
+1. Certain upgrades require reapplying the CRDs since Helm does not maintain the lifecycle of CRDs. Please see the [official documentations](https://karpenter.sh/docs/upgrading/upgrade-guide/) for details.
```
shapirov103, line 244: is this something that blueprints can handle seemlessly. Please see my comment on the next paragraph as well.

lib/addons/karpenter/index.ts
```diff
@@ -39,124 +65,203 @@ export interface EbsVolumeMapping {
...
-    securityGroupTags?: Values,
-
-    /**
-     * AMI Family: If provided, Karpenter will automatically query the appropriate EKS optimized AMI via AWS Systems Manager
+     * This is the top level nodepool specification. Nodepools launch nodes in response to pods that are unschedulable. 
+     * A single nodepool is capable of managing a diverse set of nodes. 
+     * Node properties are determined from a combination of nodepool and pod scheduling constraints.
      */
-    amiFamily?: "AL2" | "Bottlerocket" | "Ubuntu" | "Windows2019" | "Windows2022"
+    NodePoolSpec?: {
```
shapirov103, line 72: attributes lower case - `nodePoolSpec` just like the rest of the attributes. 

You can extract a type like 
`export type NodePoolSpec ...`, then use `nodePoolSpec: NodePoolSpec`.

The above approach with exported type allows configuring objects of type NodePoolSpec in customer projects that import blueprints. 

The way you declared it now - it creates an anonymous type that customers cannot use directly (not exported) unless they create an object of type `KarpenterAddOnProps`. 

Some customers brought it up, that if they want to reuse karpenter across several blueprints and just modify for example nodePoolSpec, it becomes a problem as they need to clone the whole outer structure.

lib/addons/karpenter/index.ts
```diff
@@ -39,124 +65,203 @@ export interface EbsVolumeMapping {
...
+     * It contains configuration necessary to launch instances in AWS. 
      */
-    ttlSecondsAfterEmpty?: number,
-
-    /**
-     * Priority given to the provisioner when the scheduler considers which provisioner
-     * to select. Higher weights indicate higher priority when comparing provisioners.
-     */
-    weight?: number,
+    EC2NodeClassSpec?: {
```
shapirov103, line 175: `ec2NodeClassSpec`

lib/addons/karpenter/index.ts
```diff
@@ -332,46 +474,121 @@ export class KarpenterAddOn extends HelmAddOn {
...
-        if ((Object.keys(subnetTags).length > 0) && (Object.keys(sgTags).length > 0)){
-            const provisioner = cluster.addManifest('default-provisioner', {
-                apiVersion: 'karpenter.sh/v1alpha5',
-                kind: 'Provisioner',
-                metadata: { name: 'default' },
-                spec: {
-                    providerRef: {
-                        name: "default"
+        // Deploy Provisioner (Alpha) or NodePool (Beta) CRD based on the Karpenter Version
+        if (this.options.NodePoolSpec){
```
shapirov103, line 478: code style: I observe cases when the template is quite different but injected values are the same. Such code blocks with if/then take up space and make code script style that is tough to navigate. 
Consider using kubectlProvider.addManifest which allows passing replacement values to the templates. example [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/cloudwatch-adot-addon/index.ts#L100) and similar in the codebase.

In that style you can load two templates and pass the same set of values to get the right manifest. 

Can be deferred.

lib/builders/windows-builder.ts
```diff
@@ -121,27 +121,34 @@ const defaultOptions: WindowsOptions = {
 
 const defaultKarpenterProps: addons.KarpenterAddOnProps = {
```
shapirov103, line 122: let's run `make lint` to remove this warning that defaultKarpenterProps is assigned a value but never used. 
Also is it an indication of a defect? why default props are never used?

shapirov103 (commented):
@youngjeong46 tremendous work, a lot of changes and new features. 
Please see my review comments. I will then kick off a test on 1.28. The move the default EKS version to 1.29 in a separate PR and retest before release on 1.29. 
Great work!

docs/addons/karpenter.md
jsamuel1: If we handle the translation of the CRDs to the new versions, and we have linting to ensure only valid config, what is the advantage/benefit of the middle step of upgrading to 0.32.x?
For customers self-managing Karpenter - this allows a transition, whereby both alpha and beta  CRDs are active, and therefore they may upgrade, then upgrade config to beta, then upgrade again.

docs/addons/karpenter.md
youngjeong46: lower cased ec2NodeClass and nodePoolSpec - the others (subnet, etc) are attributes inside those and should remain as so. addressed.

lib/addons/karpenter/index.ts
youngjeong46: addressed

lib/builders/windows-builder.ts
youngjeong46: This was leftovers from previous commit from another PR. Can remove as I don't see the necessity.

youngjeong46: addressed

docs/addons/karpenter.md
youngjeong46: revised. addressed.

docs/addons/karpenter.md
youngjeong46: I can change the wording to recommend not to pass them in

youngjeong46: done

lib/addons/karpenter/index.ts
youngjeong46: exported outside to SPI

docs/addons/karpenter.md
shapirov103: it seems easier to just create a separate Addon for Karpenter post 0.32 and just focus on beta and keep the existing addon as is and deprecate it. The upgrade path: remove the existing addon and add Karpenter Beta addon. If that path does not work, then we can discuss.

shapirov103: Given the amount of rework, it does not appear to be the optimal path for the current release. Refactoring can be deferred. Ideally, this type of decisions are best to discuss ahead of the implementation. If we release in the current state, it is not a one-way door, we can move forward.

docs/addons/karpenter.md
youngjeong46: This should be taken care of. I can remove this wording.

lib/addons/karpenter/index.ts
youngjeong46: will defer.

docs/addons/karpenter.md
youngjeong46: I think we can defer this.

lib/spi/types.ts
```diff
@@ -24,15 +26,251 @@ export type Values = {
     [key: string]: any;
 };
 
+/**
+ * Utility type for Karpenter requirement values
+ */
+export type reqValues = {
```
shapirov103, line 32: minor: Type names start with uppercase - ReqValues. What does Req mean here? required?

lib/spi/types.ts
```diff
@@ -24,15 +26,251 @@ export type Values = {
...
+    key: string,
```
shapirov103, line 33: is there are reason why Karpenter specific type is plance in spi as opposed to a regular export from index.ts?

shapirov103 (commented):
See my last comment, I will kick off e2e once addressed since code changes invalidate the e2e approval.

lib/spi/types.ts
youngjeong46: moved the types only relevant to Karpenter back to `index.ts`

shapirov103:
@youngjeong46 getting failure on jest test:


FAIL test/karpenter.test.ts
--
214 | ● Test suite failed to run
215 |  
216 | test/karpenter.test.ts:7:10 - error TS2724: '"../lib/spi"' has no exported member named 'reqValues'. Did you mean 'Values'?
217 |  
218 | 7 import { reqValues } from "../lib/spi";

shapirov103 (approved):
LGTM

Issue #944, opened 2024-03-08
aws-cdk-lib hardcoded to a specific version


DeveloperC286:
### Describe the bug

The version of the dependency `aws-cdk-lib` in `cdk-eks-blueprints` is hardcoded to a specific version. Therefore, if there is a mismatch with the version of `aws-cdk-lib` of your project you encounter a TypeScript complication error.

### Expected Behavior

That I am able to upgrade the version of `aws-cdk-lib` in my `package.json` and I do not get a complication error.

### Current Behavior

If I change `aws-cdk-lib` in my `package.json` to a version not matching what `cdk-eks-blueprints` wants I get a complication error.

### Reproduction Steps

Use `"@aws-quickstart/eks-blueprints": "^1.13.1` with `"aws-cdk-lib": "2.97.0"` defined in your `package.json`.

### Possible Solution

* https://github.com/aws-quickstart/cdk-eks-blueprints/pull/519

### CDK CLI Version

N/A

### Node.js Version

N/A

### Environment details (OS name and version, etc.)

N/A

## Related
* https://github.com/aws-quickstart/cdk-eks-blueprints/issues/423
* https://github.com/aws-quickstart/cdk-eks-blueprints/issues/468
* https://github.com/aws-quickstart/cdk-eks-blueprints/issues/517

DeveloperC286:
Thinking more about this, you actually do not even need to touch your `package.json` manually to get this complication errors. Your usage/installation commands will install `@aws-quickstart/eks-blueprints` with a caret(^) so it is not pinned. So if I do an `npm update` I could start getting complication errors.

DeveloperC286:
I think an important thing to note about CDK users is that they may not come from a development background. Even with my development background, prior experience with CDK and basic Node/NPM knowledge this was still an hour of debugging for me to resolve it.

shapirov103:
@DeveloperC286 thank you for your feedback. 

You can upgrade the version of CDK to any version without issues if you use the override functionality as shown [here](https://github.com/aws-samples/cdk-eks-blueprints-patterns/blob/main/package.json#L50). We should document it better for sure. 
 
However, your feedback is spot-on. Many users don't have dev background or background in node/npm and it created more issues than we expected. 

The pinned version was a conscious choice, as we observed cases when upgrading CDK version broke blueprints functionality. It is probably time to reevaluate this as it was more common in the past.

DeveloperC286:
[...]

PR #946, opened 2024-03-11
Feature/peer dependency

shapirov103:
*Issue #, if available:* #944

*Description of changes:* CDK dependency is set as `peerDependencies` to allow for smoother installation and upgrade process. Documentation (Getting started) updated for common troubleshooting cases when installing and upgrading. 

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -332,6 +332,10 @@ export class KarpenterAddOn extends HelmAddOn {
 
         karpenterChart.node.addDependency(ns);
 
+        if(clusterInfo.nodeGroups) {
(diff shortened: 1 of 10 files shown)


elamaran11 (dismissed):
LGTM. Thanks for the peer dependency documentation.

lib/addons/karpenter/index.ts
```diff
@@ -469,6 +469,10 @@ export class KarpenterAddOn extends HelmAddOn {
 
         karpenterChart.node.addDependency(ns);
 
+        if(clusterInfo.nodeGroups) {
```
shapirov103, line 472: @youngjeong46 this is what I added to fix. I was able to run it once e2e in the target account, I will run one more time. 
This code should only refence the node group (or fargate profile) used to deploy karpenter. I added deps on all node groups.

Issue #945, opened 2024-03-08
(Karpenter add-on): Update to support v0.35


youngjeong46:
### Describe the feature

Karpenter is now at v0.35. This is a feature request for Blueprints Karpenter add-on to support the latest version.

### Use Case

This is to be aligned with the latest version of Karpenter.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.131.0

### Node.js Version

v18.18.2

### Environment details (OS name and version, etc.)

AL2

PR #947, opened 2024-03-13
feat (Karpenter Add-on): v0.35 update, Node Role custom policies, and other fixes

### LINKED ISSUE #859: [Karpenter-AddOn] Remove AmazonEKS_CNI_Policy if VPC-CNI AddOn enabled with serviceAccountPolicies defined

youngjeong46:
*Issue #, if available:* Fixes #859 #893 #945

*Description of changes:*
 
This PR makes the following changes in the Karpenter Add-on:
- Update to v0.35
- Ability to use a directory of IAM policy JSONs to add custom policies to the Karpenter Node Role, in addition to the required policies
- Remove EKS CNI policy if the VPC CNI addon is in the cluster
- Fixes minor error on `instanceStorePolicy` under EC2NodeClass not setting correctly when not provided.

### DIFF
--- a/lib/addons/karpenter/index.ts
+++ b/lib/addons/karpenter/index.ts
@@ -14,6 +14,13 @@ import { Rule } from 'aws-cdk-lib/aws-events';
 import { SqsQueue } from 'aws-cdk-lib/aws-events-targets';
 import { Cluster, KubernetesVersion } from 'aws-cdk-lib/aws-eks';
 
+export const DEFAULT_KARPENTER_NODE_ROLE_POLICIES = [
+    iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEKSWorkerNodePolicy"),
+    iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEC2ContainerRegistryReadOnly"),
+    iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonSSMManagedInstanceCore"),
+    iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEKS_CNI_Policy")
+];
+
 const versionMap: Map<KubernetesVersion, string> = new Map([
     [KubernetesVersion.V1_29, '0.34.0'],
     [KubernetesVersion.V1_28, '0.31.0'],
@@ -254,11 +261,11 @@ export interface KarpenterAddOnProps extends HelmAddOnUserProps {
     ec2NodeClassSpec?: Ec2NodeClassSpec,
 
     /**
-     * Optional for Karpenter Node Role additional policies
-     * As users may require other policies in order for pods on Karpenter nodes
-     * to execute its tasks
+     * Optional for Karpenter Node Role additional policies, as users may require other policies in order for pods on Karpenter nodes to execute its tasks
+     * This is a reference to the directory where the IAM Policy JSON files are located
+     * NOTE: This addon will not validate the JSON files, and invalid files may cause deployment errors
      */
-    additionalPolicies?: string,
+    nodeRoleAdditionalPoliciesDir?: string,
 
     /**
      * Flag for enabling Karpenter's native interruption handling 
@@ -275,7 +282,7 @@ const RELEASE = 'blueprints-addon-karpenter';
 const defaultProps: HelmAddOnProps = {
     name: KARPENTER,
     namespace: KARPENTER,
-    version: 'v0.35.0',
+    version: '0.35.1',
     chart: KARPENTER,
     release: KARPENTER,
     repository: 'oci://public.ecr.aws/karpenter/karpenter',
@@ -309,6 +316,7 @@ export class KarpenterAddOn extends HelmAddOn {
         const version = this.options.version!;
 
         const interruption = this.options.interruptionHandling || false;
+        const nodeRoleAdditionalPoliciesDir = this.options.nodeRoleAdditionalPoliciesDir;
 
         // NodePool variables
         const labels = this.options.nodePoolSpec?.labels || {};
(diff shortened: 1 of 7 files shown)


jsamuel1:
@youngjeong46 -- Rather than requiring unverified JSON for the extra node role policies, are we able to add an option to pass in an array of iam ManagedPolicy names (string name to feed to fromAwsManagedPolicyName) as another pathway?

lib/addons/karpenter/index.ts
```diff
@@ -686,40 +701,57 @@ export class KarpenterAddOn extends HelmAddOn {
...
+        }
+
         // Set up Instance Profile
         const instanceProfileName = md5.Md5.hashStr(stackName+region);
-        const karpenterInstanceProfile = new iam.CfnInstanceProfile(cluster, 'karpenter-instance-profile', {
+        const karpenterInstanceProfile = new iam.CfnInstanceProfile(clusterInfo.cluster, 'karpenter-instance-profile', {
             roles: [karpenterNodeRole.roleName],
             instanceProfileName: `KarpenterNodeInstanceProfile-${instanceProfileName}`,
             path: '/'
         });
```
jsamuel1, line 744: We need to add here:
 `karpenterInstanceProfile.node.addDependency(karpenterNodeRole);`
or else cdk will fail to delete the resource later.

lib/addons/karpenter/index.ts
```diff
@@ -686,40 +701,57 @@ export class KarpenterAddOn extends HelmAddOn {
...
-            assumedBy: new iam.ServicePrincipal(`ec2.${cluster.stack.urlSuffix}`),
-            managedPolicies: [
-                iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEKSWorkerNodePolicy"),
-                iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEKS_CNI_Policy"),
-                iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonEC2ContainerRegistryReadOnly"),
-                iam.ManagedPolicy.fromAwsManagedPolicyName("AmazonSSMManagedInstanceCore"),
-            ],
+        const karpenterNodeRole = new iam.Role(clusterInfo.cluster, 'karpenter-node-role', {
+            assumedBy: new iam.ServicePrincipal(`ec2.${clusterInfo.cluster.stack.urlSuffix}`),
+            managedPolicies: DEFAULT_KARPENTER_NODE_ROLE_POLICIES,
```
jsamuel1, line 715: Can we allow adding to the list of managed policies here, via the config of the addon?
eg.  When using the CloudwatchInsights addon, you need to add:
```
    iam.ManagedPolicy.fromAwsManagedPolicyName("CloudWatchAgentServerPolicy"),
    iam.ManagedPolicy.fromAwsManagedPolicyName("AWSXrayWriteOnlyAccess")
```
Much easier to add 2 lines like that than exporting the entire policy and putting in a folder.

jsamuel1 (changes requested):

lib/addons/karpenter/index.ts
jsamuel1: Actually, a better fix would be to use iam.InstanceProfile, instead of the CfnInstanceProfile.
The name of the profile & cdk item need to both change as part of that sort of change, so you might need to add something to the md5 hash above to ensure unique.
eg.
```        const instanceProfileName = md5.Md5.hashStr(stackName+region  + clusterInfo.cluster.clusterName);
        const karpenterInstanceProfile = new iam.CfnInstanceProfile(clusterInfo.cluster, 'karpenter-iam-instance-profile', {
```

shapirov103:
@youngjeong46 please take a look when you are back. Since it is addressing a few issues, potentially missing features and update to 0.35. the complexity of the karpenter addon is affecting maintainability of this addon.
It needs to be refactored and componentized. Separate classes for `NodePool` and `NodeClass` as well as moving out version specific logic to AlphaImplementation and BetaImplementation respectively. Let's discuss.

thpham:
@youngjeong46 thank you very much for your effort, if you are doing a refactor would you be able to address the following bug as well: https://github.com/aws-quickstart/cdk-eks-blueprints/issues/1039

Thank you.

Issue #962, opened 2024-03-15
Karpenter Addon: On upgrade to 0.33+, new CRDs are not installed by blueprints


jsamuel1:
### Describe the bug

The upgrade from Alpha to Beta CRDs requires new CRDs to be installed into the cluster ahead of the helm addon.
These CRDs should be installed if they don't yet exist as a pre-install step ahead of the helm chart, when installing versions >0.32.
Currently - cluster upgrade fails with Helm failing:
``` UPDATE_FAILED        | Custom::AWSCDK-EKS-HelmChart          | <stackname>/chart-karpenter/Resource/Default (<biglongstackname>chartkarpenter<hash>) Received response status [FAILED] from custom resource. Message returned: Error: b'Error: UPGRADE FAILED: client rate limiter Wait returned an error: context deadline exceeded\n' ```

### Expected Behavior

I expected the blueprints to handle installing the new CRDs.
For new installs, the easiest would be to use the CRD helm chart of the same version.  Note - this won't work if upgrading clusters that have manually added/worked around the CRD issue, due to helm dogmatically avoiding the issue of retrofitting helm over existing resources.
Otherwise, just install the 3x manifests, as per: https://karpenter.sh/docs/upgrading/upgrade-guide/#crd-upgrades

### Current Behavior

Upgrade fails and rolls back.  (And if a cluster upgrade at the same time, rollback fails.)

### Reproduction Steps

Install an EKS blueprint using December version + Karpenter 0.31.x
Upgrade cluster + Karpenter to 0.34.x

### Possible Solution

https://karpenter.sh/docs/upgrading/upgrade-guide/#crd-upgrades

Add the CRD's to the addon code.

### CDK CLI Version

2.131.0

### EKS Blueprints Version

1.14.0

### Node.js Version

20.11.0

### Environment details (OS name and version, etc.)

gitlab ci runners

jsamuel1:
Further - if we go down the helm-crd's path, then https://karpenter.sh/docs/troubleshooting/#helm-error-when-installing-the-karpenter-crd-chart provides upgrade troubleshooting.

jsamuel1:
example 1 becomes: `kubectl label crd ec2nodeclasses.karpenter.k8s.aws nodepools.karpenter.sh nodeclaims.karpenter.sh app.kubernetes.io/managed-by=Helm --overwrite`

jsamuel1:
example 2 becomes:
[...]

PR #963, opened 2024-03-17
Bugfix -- add in the Beta1 CRD's on upgrade to v0.32 or higher.

jsamuel1:
*Issue #, if available:*
Fixes Issue #962
*Description of changes:*
Install the CRD's for Karpenter if installing addon 0.32 or higher.  
This prevents upgrade failures if original install of the Helm chart was lower than 0.32.


jsamuel1:
Question for reviewers:  
* Are there cases where a user would not want the CRDs to be installed/managed by the addOn?  (If yes, how would be give a clean way to opt-out?)
* Internet download of CRDs as published against the version, vs taking a point in time copy to put in this repo.

shapirov103:
@jsamuel1 thank you for the PR. We are validating the upgrade path and planning thus to be part of the follow-up patch release (potentially 1.14.2).

youngjeong46:
@jsamuel1 Thanks for raising the PR. Here are my initial thoughts on your questions (@shapirov103 please share your thoughts as well)

> * Are there cases where a user would not want the CRDs to be installed/managed by the addOn?  (If yes, how would be give a clean way to opt-out?)

We have customers that have their own addon extensions for CRDs and CRD resources, and manage them through GitOps. We can provide them an opt out by using a flag to enable/disable CRD installations, and document that if they disable, they need to manage the CRD lifecycle.

> * Internet download of CRDs as published against the version, vs taking a point in time copy to put in this repo.

We tend to avoid doing this, as we want to leave the CRD management to Karpenter.

jsamuel1:
> @jsamuel1 Thanks for raising the PR. Here are my initial thoughts on your questions (@shapirov103 please share your thoughts as well)
> 
> > * Are there cases where a user would not want the CRDs to be installed/managed by the addOn?  (If yes, how would be give a clean way to opt-out?)
> 
> We have customers that have their own addon extensions for CRDs and CRD resources, and manage them through GitOps. We can provide them an opt out by using a flag to enable/disable CRD installations, and document that if they disable, they need to manage the CRD lifecycle.

created a new option `manageCRDs`, which defaults false.  
Users should therefore opt-in if they are upgrading and require new CRDs.

> > * Internet download of CRDs as published against the version, vs taking a point in time copy to put in this repo.
> 
> We tend to avoid doing this, as we want to leave the CRD management to Karpenter.

shapirov103:
@youngjeong46 please check the latest update with the opt-in support at the addon level. 
@jsamuel1 please resolve the conflict in the test.

jsamuel1:
> @youngjeong46 please check the latest update with the opt-in support at the addon level. @jsamuel1 please resolve the conflict in the test.

Updated to resolve conflicts and to ensure test is correct after adding the installCRDs option.

shapirov103 (approved):
LGTM

elamaran11 (approved):
LGTM

Issue #976, opened 2024-04-04
(addons/karpenter): (KubernetesVersion.of(1.28) isn't working)


ROunofF:
### Describe the bug

In the windows karpenter pattern we were doing

        kubernetesVersion: eks.KubernetesVersion.of('1.29'),
            
But that was failing at :

        assert(versionMap.has(kubeVersion), 'Please upgrade your EKS Kubernetes version to start using Karpenter.');

But we had to change to 

        kubernetesVersion: eks.KubernetesVersion.V1_29

For our code to pass.

### Expected Behavior

No matter how you create the kubernetesVersion it should have to same behavior

### Current Behavior

Assert failed

    > 748 |         assert(versionMap.has(kubeVersion), 'Please upgrade your EKS Kubernetes version to start using Karpenter.');

### Reproduction Steps

I have a test case coming up as PR shortly

### Possible Solution

Not an expert, but the versionMap.set() of individuals value seems to provide a more consistent experience. Not sure if this is the best solution.

### CDK CLI Version

2.135.0 (build d46c474)

### EKS Blueprints Version

1.14.1

### Node.js Version

v21.7.1

### Environment details (OS name and version, etc.)

MacOS Sonoma 14.3.1

PR #977, opened 2024-04-04
[addons/karpenter] Covers the use of KubernetesVersion.of('1.28') syntax to get a KubernetesVersion

ROunofF:
*Issue #, if available:* #976

*Description of changes:* This is a test that reproduce the problem I faced in #976, I'll look at possible solution next. It's currently failing (hence the bug). Let me know if you think that is wrong.

### DIFF
--- a/test/karpenter.test.ts
+++ b/test/karpenter.test.ts
@@ -428,4 +428,28 @@ describe('Unit tests for Karpenter addon', () => {
         expect(manifest.spec.blockDeviceMappings.length).toEqual(1);
         expect(manifest.spec.blockDeviceMappings[0]).toMatchObject(blockDeviceMapping);
       });
+
+      test("Stack creation succeeded using KubernetesVersion.Of", () => {
+        const app = new cdk.App();
+
+        const blueprint = blueprints.EksBlueprint.builder();
+
+        blueprint.account("123567891").region('us-west-1')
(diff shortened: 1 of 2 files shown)


test/karpenter.test.ts
```diff
@@ -428,4 +428,28 @@ describe('Unit tests for Karpenter addon', () => {
...
         expect(manifest.spec.blockDeviceMappings[0]).toMatchObject(blockDeviceMapping);
       });
+
+      test("Stack creation succeeded using KubernetesVersion.Of", () => {
+        const app = new cdk.App();
+
+        const blueprint = blueprints.EksBlueprint.builder();
+
+        blueprint.account("123567891").region('us-west-1')
+            .version(KubernetesVersion.of('1.28'))
```
elamaran11, line 438: Why are we doing `.of`? Can we not use `eks.KubernetesVersion.V1_28` ?

elamaran11 (commented):
Thanks for the PR. Minor comment!

test/karpenter.test.ts
ROunofF: It should be able possible to use both style... This PR allows both ways to set the version.

elamaran11 (commented):
@ROunofF My question though is why is this even required? Can we not use `KubernetesVersion.V1_2X`
@shapirov103 Thoughts?

Issue #974, opened 2024-04-03
ACKAddOn ARM64 Support


dwright20:
### Describe the feature

Per the [Supported Add-ons](https://aws-quickstart.github.io/cdk-eks-blueprints/addons/#supported-add-ons), ACKAddOn doesn't support arm64, but it looks like ACK has rolled out arm64 support: https://github.com/aws-controllers-k8s/community/issues/138#issuecomment-1917776240. Can we get the [ACKAddOn supports](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/ack/index.ts#L60) tag updated?

### Use Case

I'd like to be able to deploy ACK in an EKS cluster with ARM compute

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

2.133

### Node.js Version

20.11.1

### Environment details (OS name and version, etc.)

Mac

elamaran11:
@dwright20 MemoryDB still does not support ARM64. Making this ARM64 Addon will make few deployments on non supported charts to fail. Please let us know if you have a need to support ARM64 with ACK.

dwright20:
I do -- I have a CDK app that deploys multiple baseline clusters that will have applications deployed onto them that will need to create AWS resources from the cluster. Ideally the baseline cluster has the bulk, if not all, of the addons/external charts needed. S3, DynamoDB, SQS, API GW v2 are the immediate ones that come to mind.

Probably a bit hacky, but could we do something like this:
```
export interface ServiceAckAddOnProps extends Omit<AckAddOnProps, "serviceName">{}

@supportsALL
export class S3AckAddOn extends AckAddOn{
    constructor(props?: ServiceAckAddOnProps) {
        const serviceOptions = { ...props, serviceName: AckServiceName.S3 }
        super(serviceOptions);
    }
}

@supportsX86
export class MemoryDBAckAddOn extends AckAddOn{
    constructor(props?: ServiceAckAddOnProps) {
        const serviceOptions = { ...props, serviceName: AckServiceName.MEMORYDB }
        super(serviceOptions);
    }
}
```

It's not the end all be all if getting this added into the blueprint isn't feasible (for whatever reason), but happy to implement the feature if it's something as "light" as ^

elamaran11:
[...]

PR #978, opened 2024-04-05
[Feature] Service-specific ACKAddOns

dwright20:
*Issue #, if available:* [974](https://github.com/aws-quickstart/cdk-eks-blueprints/issues/974)

*Description of changes:*
- Added service-specific ACKAddOn classes that are individually annotated with `@supportsALL` or `@supportsX86` per [ACK Registry](https://gallery.ecr.aws/aws-controllers-k8s)
- Added unit tests for each created addon that tests attempting to create the addon using an x86 builder (`EksBlueprint.builder()`) & an ARM builder (`GravitonBuilder()`)

### DIFF
--- a/lib/addons/ack/service.ts
+++ b/lib/addons/ack/service.ts
@@ -175,6 +175,7 @@ export class KafkaAckAddOn extends AckAddOn {
         super(ackProps);
     }
 }
+@supportsALL
 export class KinesisAckAddOn extends AckAddOn {
     constructor(props?: ServiceAckAddOnProps) {
         const ackProps: AckAddOnProps = {
@@ -184,6 +185,7 @@ export class KinesisAckAddOn extends AckAddOn {
         super(ackProps);
     }
 }
+@supportsALL
 export class KMSAckAddOn extends AckAddOn {
     constructor(props?: ServiceAckAddOnProps) {
         const ackProps: AckAddOnProps = {
--- a/lib/addons/ack/service.ts
+++ b/lib/addons/ack/service.ts
@@ -1,0 +1,346 @@
+import { AckAddOnProps, AckAddOn } from ".";
+import { supportsALL, supportsX86 } from "../../utils";
+import { AckServiceName } from "./serviceMappings";
+
+export interface ServiceAckAddOnProps
+    extends Omit<AckAddOnProps, "serviceName"> { }
+
+@supportsALL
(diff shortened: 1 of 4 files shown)


lib/addons/ack/service.ts
```diff
@@ -0,0 +1,348 @@
+import { AckAddOnProps, AckAddOn } from ".";
+import { supportsALL, supportsX86 } from "../../utils";
+import { AckServiceName } from "./serviceMappings";
+
+export interface ServiceAckAddOnProps
+    extends Omit<AckAddOnProps, "serviceName"> { }
+
+@supportsALL
+export class ACMAckAddOn extends AckAddOn {
```
elamaran11, line 9: I dont think we should use this design approach of one class extension per ACK chart, this will make the design difficult the expand.

elamaran11 (changes requested):
@dwright20 First of all thankyou so much for you PR. Especially liked the test scripts a lot. Fundmentally, I dont think we should use this design approach of one class per ACK chart, this will make the design difficult the expand for future addons. I think we should stick to existing design with following as design choices:
1. Add `supportsALL` to original ACK addon and create another enum of list of services which do not support ARM, In the `ACKAddon` if the AWS Service is in the ENUM list of non-ARM supported like memoryDB and if Nodes are ARM, we should error out. 
2. Add `supportsALL` to original ACK addon along with adding a documentation to say ARM is not supported today for MemoryDB (This is more of temp solution).
@shapirov103 Any other design choices?

shapirov103 (commented):
@dwright20 seems like a bit of duplication going on, but on the other hand it creates a placeholder for each ACK. 

Ideally, we would like to keep the approach generic so that little maintenance is needed.

I don't want to dismiss the design choice, however, let's use the principle of "as simple as possible but not simpler".

Let's say we change the existing approach slightly and add the target CPU architecture to the `serviceMappings.ts`, then implement this support in the `AckAddOn`. We will also add supportsAll to the AckAddOn by default and add code that will handle exclusions.  This can be done by replicating some of the code that we have in the supportsX86 and such decorators. 

If the above is implemented, what else is missing that requires explicit class per Ack controller implementation? Do you envision additional use cases?

shapirov103:
@dwright20 please see my comment and let me know what you think. I mentioned an approach to simplify the design approach to make it easier for us to maintain. This has been open for a while.

PR #985, opened 2024-04-11
[Karpenter] Fix ignored namespace prop

muckelba:
I've stumbled across a bug while installing karpenter: I can set the namespace in the KarpenterAddOnProps, but the createNamespace() and createServiceAccount() functions would still use the default variable as their namespace.


shapirov103:
@youngjeong46 please have a look, I see you have hardcoded the namespace, was there a particular reason for it?

shapirov103 (commented):
@muckelba thank you for your contribution. I wanted to check if there was any rationale to hardcode the ns before merging this. 
I will kick off testing.

PR #988, opened 2024-04-22
Feature to introduce Dev Containers

elamaran11:
*Issue #, if available:*

*Description of changes:*

Feature to introduce Dev Containers


shapirov103 (commented):
I finally got to test it out. 
From the good side, my setup worked at the end. Provisioned c5.xlarge  and setup devcontainers, with VSCode local IDE and remotessh access to the repo. My feedback (tactical and more serious items mixed):

1. I got an error on setup "setup failed, try to rerun under root". Unsure what it was, I ran reset on the workspace and it started working.   
2. Does your machine register on the PVRE security policy violation list?
3. By default c5.xlarge was provisioned. It seems a bit slow. What machine are you using? Also have you tried with graviton 3, i.e. c7g.2xlarge?
4. We need to add npm install to dev container, otherwise ws is not quite ready
5. What credentials are used in the workspace? I am getting errors like region is missing on `cdk list` as well as complaints about bootstrap and inability of the devcontainer role to assume bootstrap credentials. This should be customized to something reasonable and we need to document it. 
6. Can we do something about the terminal? e.g set to zsh with reasonable addons, i know you are a fan of powerlevel10k. Ideally would love to have command completion working in the terminal. 

Overall, the approach seems solid and we could use it for development or even workshop provided appsec approves.

elamaran11:
> I finally got to test it out. From the good side, my setup worked at the end. Provisioned c5.xlarge and setup devcontainers, with VSCode local IDE and remotessh access to the repo. My feedback (tactical and more serious items mixed):
> 
> 1. I got an error on setup "setup failed, try to rerun under root". Unsure what it was, I ran reset on the workspace and it started working.
> 2. Does your machine register on the PVRE security policy violation list?
> 3. By default c5.xlarge was provisioned. It seems a bit slow. What machine are you using? Also have you tried with graviton 3, i.e. c7g.2xlarge?
> 4. We need to add npm install to dev container, otherwise ws is not quite ready
> 5. What credentials are used in the workspace? I am getting errors like region is missing on `cdk list` as well as complaints about bootstrap and inability of the devcontainer role to assume bootstrap credentials. This should be customized to something reasonable and we need to document it.
> 6. Can we do something about the terminal? e.g set to zsh with reasonable addons, i know you are a fan of powerlevel10k. Ideally would love to have command completion working in the terminal.
> 
> Overall, the approach seems solid and we could use it for development or even workshop provided appsec approves.

@shapirov103 Thankyou for review the PR and getting the setup working. Please see my response below for your points:

1. This happens to me sometimes but not always. Probably something to do with using DevPod from work mac and Isengard accounts idk. I have not seen anyone complaining who dont use Isengard and PVRE.
2. No i have been using for 3 weeks now and i dont see any PVRE violation so far so we are good to use.
3. I try to use `c5.2xlarge` and works like charm. Also DevPod reuses machines if that option i set. It should work on Graviton machines too.
4. Check this, i did research on `npm install` but that was claimed to be anti pattern and container will fail to launch, check [this](https://stackoverflow.com/questions/55976358/should-a-vscode-dev-container-run-npm-install-as-part-of-its-setup) thread.
5. Git and AWS credentials are injected from your terminal. Watch the AWS provider settings in DevPod
6. The terminal already has ZSH installed but it defaults to bash like any linux machine. I will see if we can override this behavior. `ohmyzsh` is also installed by default all the user has to do is update to add plugins which is not automatic.

Overall, i will work on item number 6 and will keep you posted.

Issue #1010, opened 2024-06-10
ExternalDNS addon support for Gateway / VirtualService types


tdalbo92:
### Describe the feature

Allow a mechanism to add custom sources for ExternalDNS to monitor.  Currently Ingress and Service are the only supported ones.

### Use Case

If using Gateway API spec services, such as Istio or Envoy Gateway, routes are exposed through different CRDs than Service or Ingress.  I don't see a way to configure or override the container args[1] to use different sources.

[1] https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/istio.md#manifest-for-clusters-without-rbac-enabled

### Proposed Solution

Adding a values parameter to the addon that passes value overrides into the helm chart would suffice.

### Acknowledgements

- [X] I may be able to implement this feature request

### CDK version used

2.133

### EKS Blueprints Version

1.14.1

### Node.js Version

v18.20.1

### Environment details (OS name and version, etc.)

MacOS Sonoma 14.5 ARM64

PR #1013, opened 2024-06-11
[feat] Add service override parameter to ExternalDNS props

tdalbo92:
*Issue #, if available:*
Issue #1010 
*Description of changes:*

The current default values will allow ExternalDNS to listen for Service and Ingress only.  If you're using something else, such as Istio or Envoy Gateway, it will not look for host names in Gateway, VirtualService, or HTTPRoutes.  This allows that override via helm values.


tdalbo92:
Feel free to counter-propose or toss it entirely.  This is just an idea I had to minimally allow this values override, without exposing the entire values object to the caller.

hshepherd (approved):
This would be really helpful for our implementation

elamaran11:
LGTM

PR #1018, opened 2024-06-13
Generic Construct for EKS blueprint

bugathagit:
Description of changes: Currently, EKS blueprint can be used by only cdk.stack users or the users who relies on stack derivates. All other users cannot use this package. So, this Generic construct will help the users of any kind to use eks blueprint code. 

Example: 
In our company, we use different stack named PipelineStack which is not able to use cdk.stack to deploy the resources created by the blueprint. Which resulted in writing our own blueprint/blueprint builder code. The current change will help to use the eks blueprint code directly.

### DIFF
--- a/lib/spi/types.ts
+++ b/lib/spi/types.ts
@@ -123,7 +123,7 @@ export class ResourceContext {
 
     private readonly resources: Map<string, IConstruct> = new Map();
 
-    constructor(public readonly scope: any, public readonly blueprintProps: EksBlueprintProps) { }
+    constructor(public readonly scope: Construct, public readonly blueprintProps: EksBlueprintProps) { }
 
     /**
      * Adds a new resource provider and specifies the name under which the provided resource will be registered,
--- a/lib/spi/types.ts
+++ b/lib/spi/types.ts
@@ -123,7 +123,7 @@ export class ResourceContext {
 
     private readonly resources: Map<string, IConstruct> = new Map();
 
-    constructor(public readonly scope: cdk.Stack, public readonly blueprintProps: EksBlueprintProps) { }
+    constructor(public readonly scope: any, public readonly blueprintProps: EksBlueprintProps) { }
 
     /**
      * Adds a new resource provider and specifies the name under which the provided resource will be registered,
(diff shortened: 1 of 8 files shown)


lib/spi/types.ts
```diff
@@ -123,7 +123,7 @@ export class ResourceContext {
 
     private readonly resources: Map<string, IConstruct> = new Map();
 
-    constructor(public readonly scope: cdk.Stack, public readonly blueprintProps: EksBlueprintProps) { }
+    constructor(public readonly scope: any, public readonly blueprintProps: EksBlueprintProps) { }
```
shapirov103, line 126: Why scope is any? Let's make it `Construct`.

shapirov103 (commented):
Great work, please see my comment, otherwise looks good.

lib/spi/types.ts
bugathagit: Yes, Contruct was my first choice. But I ran into issues while doing that. So didn't want to refactor/change too much of current code. So, I used any type to be generic.

Issues because of Construct:

lib/resource-providers/hosted-zone.ts:92:41 - error TS2339: Property 'formatArn' does not exist on type 'Construct'.

92         const delegationRoleArn = stack.formatArn({
                                           ~~~~~~~~~

lib/resource-providers/iam.ts:60:43 - error TS2339: Property 'account' does not exist on type 'Construct'.

60             `arn:aws:iam::${context.scope.account}:oidc-provider/${this.url.substring(httpsPrefix.length)}`

shapirov103: @bugathagit I am glad compiler helps us catch these issues. So for the places when stack is needed, let's use Stack.of(context.scope), it is a minor change that will go into the hosted zone 
`const stack = context.scope; ` becomes `const stack = Stack.of(context.scope);`.

bugathagit: Thank you!! Updated the same in commit `1fe12f9a014c0c90be7cf91e0b2bb000f71f5c16`

shapirov103 (approved):
LGTM

Issue #625, opened 2023-03-17
Why is the VPC CNI a requirement for the Karpenter Addon?


danielloader:
### Describe the documentation issue

I can't see why this is a requirement, and isn't when deploying via eksctl or terraform. 

Can anyone explain why it's a requirement?

### Links

https://aws-quickstart.github.io/cdk-eks-blueprints/addons/karpenter/

youngjeong46:
@danielloader This change was pushed to release 1.6.1. Please verify that it is no longer required.

PR #1024, opened 2024-06-20
[KEDA] Update helm values to work from v2.14.0

vumdao:
*Issue:*
Since v2.14.0, [#625](https://github.com/kedacore/charts/pull/625/files#diff-3b4e4c10436059483f20fa5b4b2e507a0bcead63f71af2c02ca41dbff3c8b3ad) updates some Helm chart values that are not backward compatible with the old settings. The serviceAccount has been separated per service, such as operator, metricServer, and webhooks.This causes the upgrade to v2.14.0+ to recreate the new operator serviceAccount without the IAM role annotation

*Description of changes:*
- Update version to current latest v2.14.2
- Move creating namespace out of `irsaRole` condition check to avoid CDK delete `keda` namespace if we change irsaRole[] to empty
- Remove workaround for SQS Scalar https://github.com/kedacore/keda/issues/837 as KEDA-is-secure-by-default
- Update  `serviceAccount` to `serviceAccount.operator`
 
### DIFF
--- a/lib/addons/keda/index.ts
+++ b/lib/addons/keda/index.ts
@@ -12,7 +12,7 @@ import { HelmAddOn, HelmAddOnProps, HelmAddOnUserProps } from "../helm-addon";
 export interface KedaAddOnProps extends HelmAddOnUserProps {
     /**
      * Version of the helm chart to deploy
-     */    
+     */
     version?: string;
     /**
      * Name of the KEDA operator
@@ -25,12 +25,12 @@ export interface KedaAddOnProps extends HelmAddOnUserProps {
     /**
      * securityContext: fsGroup
      * Check the workaround for SQS Scalar with IRSA https://github.com/kedacore/keda/issues/837#issuecomment-789037326
-     */   
+     */
     podSecurityContextFsGroup?: number;
     /**
      * securityContext:runAsGroup
      * Check the workaround for SQS Scalar with IRSA https://github.com/kedacore/keda/issues/837#issuecomment-789037326
-     */   
+     */
     securityContextRunAsGroup?: number;
     /**
      * securityContext:runAsUser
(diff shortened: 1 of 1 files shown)


lib/addons/keda/index.ts
```diff
@@ -74,19 +74,20 @@ export class KedaAddOn extends HelmAddOn {
...
     const cluster = clusterInfo.cluster;
     let values: Values = populateValues(this.options);
     values = merge(values, this.props.values ?? {});
-    
+
+    const namespace = createNamespace(this.options.namespace! , cluster);
+
     if (this.options.irsaRoles!.length > 0) {
       //Create Service Account with IRSA
       const opts = { name: this.options.kedaOperatorName, namespace: this.options.namespace };
```
elamaran11, line 86: could we use `namespace` directly.

lib/addons/keda/index.ts
```diff
@@ -108,14 +109,10 @@ export class KedaAddOn extends HelmAddOn {
 function populateValues(helmOptions: KedaAddOnProps): Values {
   const values = helmOptions.values ?? {};
 
-  // Check the workaround for SQS Scalar https://github.com/kedacore/keda/issues/837
   setPath(values, "operator.name",  helmOptions.kedaOperatorName);
-  setPath(values, "podSecurityContext.fsGroup",  helmOptions.podSecurityContextFsGroup);
```
elamaran11, line 113: Why are these SecContext fields removed? Are these no longer supported?

elamaran11 (changes requested):
@vumdao Thanks for the PR. Have couple of comments.

shapirov103 (commented):
Thank you for this contribution!

`podSecurityContextFsGroup` as well as security context fields: they are not supported by default, let's add deprecation to these fields and a note that these have no effect for version 2.14 and above. 
There is also impact to `docs/addons/keda.md`, but you can defer it. 

Other than that looks great.

lib/addons/keda/index.ts
vumdao: Not sure I understand correctly
To use namespace directly, ensure that `createNamespace` is set to true. This setting is optional.

lib/addons/keda/index.ts
vumdao: As @shapirov103's comment, it's deprecated, let me add deprecation to these fields and a note it

elamaran11: Yes that helps.

lib/addons/keda/index.ts
elamaran11: Ignore the comment.

elamaran11 (commented):
@vumdao This is great, Please also update the documentation page - https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/addons/keda.md

elamaran11 (commented):
LGTM

shapirov103:
Failure in e2e unrelated to the changes here. 
FAIL test/cloudwatch-insights.test.ts

I will create a PR to fix them and rerun.

PR #1031, opened 2024-06-29
[Karpenter] Fix CRDs URL not found

vumdao:
*Issue*
If `installCRDs` enabled, it got following error
```
Error: Server responded to https://raw.githubusercontent.com/aws/karpenter/0.37.0/pkg/apis/crds/karpenter.sh_nodepools.yaml with status code 404:
404: Not Found
```

*Description of changes:*
Change `${version}` to `v${version}`
eg. https://raw.githubusercontent.com/aws/karpenter/v0.37.0/pkg/apis/crds/karpenter.sh_nodepools.yaml


shapirov103 (commented):
@vumdao the version is expected to be set in the format `v${semver}`. Please see the current setting [here](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/lib/addons/karpenter/index.ts#L303).

Here is what we set in the test: https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/blueprint-construct/index.ts#L152, also all our docs have the same convention.

If your fix is applied, seems like it will break the current logic, as the CRD url will have double v: `vv0.34.0`.

The [docs](https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/docs/addons/karpenter.md) may not articulate that aspect enough, that may something we should fix. 

Please let me know your thoughts, and decide how to move forward with this.

lib/addons/karpenter/index.ts
```diff
@@ -519,10 +519,14 @@ export class KarpenterAddOn extends HelmAddOn {
         }
 
         if (semver.gte(version, "0.32.0") && installCRDs){
+            // Since v0.35.0, Karpenter OCI tags and Helm chart version are now valid semantic versions,
+            // meaning that the v prefix from the git tag has been removed and they now follow the x.y.z pattern
+            // But the CRDs URL still uses the git tag format, so we need to check for the version and adjust the URL accordingly
+            let _version = (semver.gte(version, "0.35.0")) ? `v${version}` : version;
```
shapirov103, line 525: @vumdao alsmost there. 
I looked through the branches of Karpenter and it does not look like the version convention with `v` is something specifically enforced before 0.35.0. 

Let's add an explicit check: `if(!version.startsWith('v') then add `v` into the version in URL, to avoid potential mismatches.

shapirov103 (commented):
@vumdao please see my remaining comment

shapirov103 (approved):
LGTM, thank you

PR #1032, opened 2024-06-30
IPv6

bugathagit:
*Description of changes:*
This change will support ipv6 cluster with EKS cluster.

Currently, tested aws-cni and karpenter addon with IPv6 cluster. Other Addons are not yet tested with ipv6 blueprint code. Most of the addons could work if you pass IPv6 related props properly.

### DIFF
--- a/examples/examples.ts
+++ b/examples/examples.ts
@@ -42,6 +42,7 @@ builder()
     .clusterProvider(new bp.MngClusterProvider({
         ...publicCluster
     }))
+    .addOns(new bp.addons.VpcCniAddOn())
     .enableControlPlaneLogTypes(bp.ControlPlaneLogType.API, bp.ControlPlaneLogType.AUDIT)
     .build(app, "mng-blueprint");
 
--- a/lib/addons/cloud-watch-insights/index.ts
+++ b/lib/addons/cloud-watch-insights/index.ts
@@ -10,11 +10,11 @@ import {KubernetesVersion} from "aws-cdk-lib/aws-eks";
 // aws eks describe-addon-versions --kubernetes-version <kubernetes-version> --addon-name amazon-cloudwatch-observability \
 //     --query 'addons[].addonVersions[].{Version: addonVersion, Defaultversion: compatibilities[0].defaultVersion}' --output table
 const versionMap: Map<KubernetesVersion, string> = new Map([
-  [KubernetesVersion.V1_30, "v1.7.0-eksbuild.1"],
-  [KubernetesVersion.V1_29, "v1.7.0-eksbuild.1"],
-  [KubernetesVersion.V1_28, "v1.7.0-eksbuild.1"],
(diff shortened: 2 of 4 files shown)


lib/resource-providers/vpc.ts
```diff
@@ -2,26 +2,29 @@ import { Tags } from 'aws-cdk-lib';
...
+   primaryCidr?: string,
    secondaryCidr?: string,
    secondarySubnetCidrs?: string[]
 }
 
 /**
  * VPC resource provider 
  */
-export class VpcProvider implements ResourceProvider<ec2.IVpc> {
+export class VpcProvider extends Ipv6VpcProvider {
```
shapirov103, line 20: Do we really achieve much with inheritance here? Seems like delegation is cleaner option here. 

Ideally, we should have `CreateIpv6VpcProvider` and `CreateIpV4VpcProvider` and just delegate to one vs the other in the provide method if VPC needs to be created.

examples/blueprint-ipv6-construct/index.ts
```diff
@@ -0,0 +1,233 @@
...
+        userLog.settings.minLevel = 2; // debug
+
+        const teams: Array<blueprints.Team> = [
+            new team.TeamTroi,
+            new team.TeamRiker(scope, teamManifestDirList[1]),
+            new team.TeamBurnham(scope, teamManifestDirList[0]),
+            new team.TeamPlatform(process.env.CDK_DEFAULT_ACCOUNT!)
+        ];
+
+        const nodeRole = new blueprints.CreateIPv6NodeRoleProvider("blueprint-node-role", new iam.ServicePrincipal("ec2.amazonaws.com"),
```
shapirov103, line 41: Why do we need to pass principal and managed policies to this provider? Should not that be needed in 90% of cases? If so , why not put that logic in the provider itself and if customers want to pass extra, we can let them.

lib/utils/ipv6-utils.ts
```diff
@@ -0,0 +1,35 @@
+
+/*
+** This policy is required for all the roles which are manging/creating the nodes in the cluster.
+** So, we need it for both cluster node role and karpenter node role.
+** Refer to: https://docs.aws.amazon.com/eks/latest/userguide/cni-iam-role.html#cni-iam-role-create-role
+ */
+import {PolicyDocument} from "aws-cdk-lib/aws-iam";
+
+export function getEKSNodeIpv6PolicyDocument(){
```
shapirov103, line 9: minor: let's declare the return value in the function signature rather than letting js derive it.

shapirov103 (commented):
@bugathagit great work! 

I added a few comments, mostly code style / design related. 

Other than comments, let's add a documentation on IPv6 support to the markdowns so that customers can 1/ discover it 2/ learn how to use it. I think we should also mention it in the getting started and in the README with links to the IPV6 support. 

Let's resolve conflicts so that we can run unit tests and later functional test as well.

PR #1044, opened 2024-07-15
Additional Settings to VpcCniAddOn

PeterKoegel:
This PR adds support to provide values for certain settings of the VpcCniAddOn that were introduced in amazon-vpc-cni-k8s [v1.16.0](https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.16.0) in PR [additional settings to the helm chart](https://github.com/aws/amazon-vpc-cni-k8s/pull/2715)

- enable-windows-prefix-delegation
- warm-prefix-target
- warm-ip-target
- minimum-ip-target
- branch-eni-cooldown

I verified that this code change does not introduce any functional changes for the already existing vpc-cni settings by comparing the generated cloud formation output for different sets of parameters generated with and without this change and assure they are equal. For the newly introduced vpc-cni settings I tested that they are deployed to an eks ckuster as desired.

### DIFF
--- a/lib/addons/vpc-cni/index.ts
+++ b/lib/addons/vpc-cni/index.ts
@@ -432,19 +432,16 @@ export class VpcCniAddOn extends CoreAddOn {
   }
 }
 
-/**
- * Iterates over all values including nested child objects, removes undefined entries and stringifies the remaining if they are not already strings
- */
-function ConvertPropertiesToString(helmValues: Values): void {
+  /**
+   * Iterates over all Values including nested child objects and removes undefined entries
+   */
+function RemoveUndefined(helmValues: Values): void {
   Object.keys(helmValues).forEach(key => {
     if (helmValues[key] === undefined) {
       delete helmValues[key];
     }
     else if (typeof helmValues[key] === 'object'){
-      ConvertPropertiesToString(helmValues[key]);
-    }
-    else if (typeof helmValues[key] !== 'string'){
-      helmValues[key] = JSON.stringify(helmValues[key]);
+      RemoveUndefined(helmValues[key]);
     }
   });
 }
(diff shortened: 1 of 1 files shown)


lib/addons/vpc-cni/index.ts
```diff
@@ -450,7 +481,12 @@ function populateVpcCniConfigurationValues(props?: VpcCniAddOnProps): Values {
       WARM_PREFIX_TARGET: props?.warmPrefixTarget,
     },
     enableNetworkPolicy: JSON.stringify(props?.enableNetworkPolicy),
-    enableWindowsIpam: JSON.stringify(props?.enableWindowsIpam)
+    enableWindowsIpam: JSON.stringify(props?.enableWindowsIpam),
```
shapirov103, line 484: minor: is there a reason to stringify this here? the code on line 495 should take care of that, if it does not, please let me know. Ideally, I want to have a consistent approach to extend this structure.

shapirov103 (commented):
@PeterKoegel thank you for the PR, please see my minor comment.

lib/addons/vpc-cni/index.ts
PeterKoegel: The code in line 459 in deed does NOT take care for properties on the toplevel config object.

To avoid mixing up different things in one PR I created #1048 and added a more detailed description there.

PeterKoegel:
The schema that is returned by `aws eks describe-addon-configuration --addon-name vpc-cni --addon-version v1.18.1-eksbuild.1` shows that there are some properties defines as `format: integer, type: string` while some of the properties that are introduced by this PR are defined as `type: integer` in the VPC-CNI addon's config. So it is not possible anymore to convert all integers into strings.
Instead JSON.stringify() is called at assignment of each parameter with `format: boolean, type: string` or `format: integer, type: string`.

shapirov103 (approved):
LGTM

PR #1048, opened 2024-07-26
handle stringification the same way for all vpc-cni configuration pro…

PeterKoegel:
A finding of #1044 was that the way the configuration properties for the vpc-cni plugin are converted to strings is not handled in the same way for all properties.

Configuration for the vpc-cni core addon is defined as following structure:

```
  const result: Values = {
    init: {
      env: {
        DISABLE_TCP_EARLY_DEMUX: props?.disableTcpEarlyDemux,
        ENABLE_V6_EGRESS: props?.enableV6Egress,
      }
    },
    env: {
      AWS_EC2_ENDPOINT: props?.awsEc2Endpoint,
      ADDITIONAL_ENI_TAGS: props?.additionalEniTags,
      .. some more properties
    },
    enableNetworkPolicy: props?.enableNetworkPolicy,
    enableWindowsIpam: props?.enableWindowsIpam
  };
```

Before that configuration is passed to the core addon undefined properties are removed and non-string properties are converted into strings by calling JSON.stringify().

For the "env" and "init.env" subobjects this is done in line 456 and 461 in a loop that iterates over all properties of this child objects, for the properties of the main object however JSON.stringify() is called when assigning the values in line 469 and 470.

This pull request changes that in a way that there is a recursive function "ConvertPropertiesToString" that handles the functionality described above by iterating over the whole configuration object including the embeded subobjects.

I verified that this code change does not introduce any functional changes by comparing the generated cloud formation output for different sets of parameters generated with and without this change and assure they are equal.

### DIFF
--- a/lib/addons/vpc-cni/index.ts
+++ b/lib/addons/vpc-cni/index.ts
@@ -401,19 +401,19 @@ export class VpcCniAddOn extends CoreAddOn {
   }
 }
 
-  /**
-   * Iterates over all values including nested child objects, removes undefined entries and stringifies the remaining if they are not already strings
-   */
+/**
+ * Iterates over all values including nested child objects, removes undefined entries and stringifies the remaining if they are not already strings
+ */
 function ConvertPropertiesToString(helmValues: Values): void {
   Object.keys(helmValues).forEach(key => {
     if (helmValues[key] === undefined) {
-      delete helmValues[key]
+      delete helmValues[key];
     }
     else if (typeof helmValues[key] === 'object'){
       ConvertPropertiesToString(helmValues[key]);
     }
     else if (typeof helmValues[key] !== 'string'){
-      helmValues[key] = JSON.stringify(helmValues[key])
+      helmValues[key] = JSON.stringify(helmValues[key]);
     }
   });
 }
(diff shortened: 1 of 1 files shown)


PeterKoegel:
@shapirov103 : this PR is not fully tested by it would be interesting for me to know if this is the right approach to solve this issue.

lib/addons/vpc-cni/index.ts
```diff
@@ -401,6 +401,23 @@ export class VpcCniAddOn extends CoreAddOn {
...
+function ConvertPropertiesToString(helmValues: Values): void {
+  Object.keys(helmValues).forEach(key => {
+    if (helmValues[key] === undefined) {
+      delete helmValues[key]
+    }
+    else if (typeof helmValues[key] === 'object'){
+      ConvertPropertiesToString(helmValues[key]);
+    }
+    else if (typeof helmValues[key] !== 'string'){
+      helmValues[key] = JSON.stringify(helmValues[key])
```
shapirov103, line 416: github action issue missing semicolon here and line 410

shapirov103 (commented):
let me validate the e2e test and will merge after the minor code quality issues are addressed

shapirov103:
e2e test failure unrelated to the changes. Will be addressed in a separate PR.

shapirov103 (approved):
LGTM, thank you!

PR #1075, opened 2024-09-01
flag for deploy to isolated subnet

bezdomniy:
*Issue #, if available:*

*Description of changes:*
Add support for deploying Cluster resource in an isolated subnet. CDK EKS already supports deployment in an isolated subnet and this is describes in this example: https://github.com/aws-samples/aws-cdk-examples/tree/main/java/eks/private-cluster. This PR just adds an "isolatedCluster" flag, which adds Cluster props to based on the example.

Tested and deployed successfully.

Note that given the lack of internet access, addon's which require access to public repositories (like AwsLoadBalancerControllerAddOn) cannot be used.

### DIFF
--- a/cdk.json
+++ b/cdk.json
@@ -5,6 +5,7 @@
         "eks.default.min-size": 1,
         "eks.default.max-size": 2,
         "eks.default.desired-size": 1,
-        "eks.default.private-cluster": "false"
+        "eks.default.private-cluster": "false",
+        "eks.default.isolated-cluster": "false"
     }
 }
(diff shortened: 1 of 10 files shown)


shapirov103 (commented):
@bezdomniy great work, definitely makes it more convenient to create isolated clusters. Implementation looks straightforward to me. 

I will need to test it out, and atm our examples cluster is already overloaded with different use cases, so I will run it manually. 

Please add your test case (functional) to https://github.com/aws-quickstart/cdk-eks-blueprints/blob/main/examples/examples.ts. 
We are building an approach to run multiple e2e validation flows, so once it is in the codified form, it will be easy for me to migrate it to the fully automated approach.

jeremymcgee73:
I tested this PR, and it did solve my problem. I was able to deploy EKS in an isolated subnet.

shapirov103:
Validated by @jeremymcgee73 (thank you!). Merging.

shapirov103 (approved):
LGTM, thank you for your contribution!

PR #1084, opened 2024-10-09
Fixed tests, build issues, upgraded CDK and libs

shapirov103:
*Issue #, if available:*

*Description of changes:* In prep for the release fixing test and libs to merge the pull request that are currently in the queue.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
       },
       "ghcr.io/devcontainers/features/aws-cli:1": {},
       "ghcr.io/devcontainers-contrib/features/aws-cdk:2": {
-        "version": "2.147.3"
+        "version": "2.161.1"
       }
     },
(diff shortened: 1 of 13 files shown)


elamaran11 (commented):
@shapirov103 Changes look great. Only concern though is on the reoccurrence of this issue with all future e2e's. Could we do another e2e run before merge?

shapirov103:
> @shapirov103 Changes look great. Only concern though is on the reoccurrence of this issue with all future e2e's. Could we do another e2e run before merge?

There is not much I can do other than temporarily remove ipv6 test to prevent timeout. The plan was to move to ABP.

elamaran11:
> > @shapirov103 Changes look great. Only concern though is on the reoccurrence of this issue with all future e2e's. Could we do another e2e run before merge?
> 
> There is not much I can do other than temporarily remove ipv6 test to prevent timeout. The plan was to move to ABP.

Sounds good.

PR #514, opened 2022-10-03
fix(external-secrets): wait for eso deployed

pflorek:
fixes https://github.com/aws-quickstart/cdk-eks-blueprints/issues/509

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1102, opened 2024-11-06
fix(aws-lb-controller): wait till deployed

wombelix:
*Issue #, if available:*

N/A

*Description of changes:*

`AlbController` webhook isn't ready directly after the AddOn deployment. 
This cause a race condition when another AddOn creates Kubernetes Manifest resources that rely on it, like `TargetGroupBinding`.
Sometimes it is successful, sometimes not. To avoid such situations, `addHelmChart` of `AlbController` should wait similar as it was implemented for `External Secrets`: https://github.com/aws-quickstart/cdk-eks-blueprints/pull/514

Example of such an error when the proposed fix is not applied:
```
Received response status [FAILED] from custom resource. Message returned: Error: b'Error from server (InternalError): error when creating "/tmp/manifest.yaml": Internal error occurred:
failed calling webhook "mtargetgroupbinding.elbv2.k8s.aws": failed to call webhook: Post "https://aws-load-balancer-webhook-service.kube-system.svc:443/mutate-elbv2-k8s-aws-v1beta1-tar
getgroupbinding?timeout=10s": no endpoints available for service "aws-load-balancer-webhook-service"\n'
```

### DIFF
--- a/lib/addons/aws-loadbalancer-controller/index.ts
+++ b/lib/addons/aws-loadbalancer-controller/index.ts
@@ -6,6 +6,7 @@ import { registries } from "../../utils/registry-utils";
 import { HelmAddOn, HelmAddOnUserProps } from "../helm-addon";
 import { AwsLoadbalancerControllerIamPolicy } from "./iam-policy";
 import { supportsALL } from "../../utils";
+import { Duration } from "aws-cdk-lib";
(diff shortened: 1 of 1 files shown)


shapirov103 (approved):
LGTM

PR #1114, opened 2025-02-11
S3 CSI Addon - Support multiple buckets

usmanatron:
*Issue #, if available:*

None

*Description of changes:*

Tweak the new S3 CSI Addon so it supports multiple S3 buckets.  Looking at the CSI Driver itself, it does support multiple bucket; it's just the Service Account creation here which is currently constrained to a single bucket name.

One thing I'm not sure on is how this changes the Props object - it means consumers will have to change their usage.  I could instead make s3BucketNames a `string | string[]` to handle this, but I'm not sure if its worth the effort.  What are your thoughts?

### DIFF
--- a/lib/addons/s3-csi-driver/index.ts
+++ b/lib/addons/s3-csi-driver/index.ts
@@ -18,7 +18,7 @@ export interface S3CSIDriverAddOnProps extends HelmAddOnUserProps {
     /**
      * The names of the S3 buckets to be used by the driver
      */
-    s3BucketNames: string[];
+    bucketNames: string[];
     /**
      * Create Namespace with the provided one (will not if namespace is kube-system)
      */
@@ -36,7 +36,7 @@ const defaultProps: HelmAddOnUserProps & S3CSIDriverAddOnProps = {
   version: "v1.11.0",
   repository: "https://awslabs.github.io/mountpoint-s3-csi-driver",
   createNamespace: false,
-  s3BucketNames: []
+  bucketNames: []
 };
 
 @supportsALL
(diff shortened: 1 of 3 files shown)


lib/addons/s3-csi-driver/index.ts
```diff
@@ -16,9 +16,9 @@ const S3_DRIVER_POLICY = 's3-csi-driver-policy';
  */
 export interface S3CSIDriverAddOnProps extends HelmAddOnUserProps {
     /**
-     * ARN of the S3 bucket to be used by the driver
+     * The names of the S3 buckets to be used by the driver
      */
-    s3BucketName: string;
+    s3BucketNames: string[];
```
shapirov103, line 21: Minor: since it is an s3-csi-driver, s3 is redundant in the names. If we are changing the names, let's make it just `bucketNames`.

shapirov103 (commented):
@usmanatron thank you for the contribution!

The change makes sense. The only concern is that it is a breaking change, so I will have to put it into 1.17 release and put a note on it. It is nothing major, but will require customers to apply a code fix if they are already using this addon.

shapirov103:
@jesperalmstrom FYI ^^

shapirov103 (commented):
Looks good, let me run the tests

usmanatron:
@shapirov103 - I'd like to further extend this add-on by adding support for customer-managed KMS keys.  Which of the following would you prefer?

1. Add the changes to this PR and get it re-reviewed
2. Create a new PR - I'd like to base the other changes on the work here though, which might make things messy?

The changes are ready, lmk which is easier for you :-)

shapirov103:
@usmanatron this PR is ready for merge, so let me merge it and we will have incremental value in the repo already. Let's go with option 2.

PR #1117, opened 2025-02-19
S3 CSI Driver - add support for Customer Managed KMS Keys

usmanatron:
*Issue #, if available:* None

*Description of changes:*
The IAM policy for the S3 CSI Driver is missing permissions for customer managed KMS keys (when used).  This PR adds support for specifying KMS Keys by ARN.  The new props property is optional as I suspect this isn't going be needed most of the time!

Note: the kms permissions detailed in the Policy are enough to get it all working for me; I struggled to find any documentation to confirm that this includes all permissions we may need.  


shapirov103 (commented):
Changes look good, thank you for that. Running e2e and will be in  the next release, provided everything passes.

PR #1118, opened 2025-02-21
Adding dependency on instance role

misirlou-tg:
*Issue #, if available:*
N/A

*Description of changes:*
The instance profile created has a dependency on the instance role, but this is not captured by the level 1 construct `CfnInstanceProfile`. Explicitly adding the dependency so that the addon can be reliably destroyed in CloudFormation.


shapirov103 (approved):
Thank you for your contribution! 
LGTM

PR #1121, opened 2025-03-04
Updating cap on version in helm addons

Eli1123:
*Issue #, if available:*

Increasing cap for version which allows passing extra custom parameters in the version field that will be used for installation (valid for OCI repos).


shapirov103:
Fixes #1121

shapirov103 (approved):
LGTM

PR #1122, opened 2025-03-04
Private cluster defect (always on)

shapirov103:
Unable to create a cluster in the public subnets only

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/examples/examples.ts
+++ b/examples/examples.ts
@@ -38,13 +38,25 @@ builder()
     .clusterProvider(new bp.FargateClusterProvider({...publicCluster}))
     .build(app, "fargate-blueprint");
 
-builder()
+const mng = builder()
     .clusterProvider(new bp.MngClusterProvider({
-        ...publicCluster
+        ...publicCluster, 
+
+        launchTemplate: {
+            requireImdsv2: true, 
+            tags: {
+                "cost-center": "2122",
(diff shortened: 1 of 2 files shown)


examples/examples.ts
```diff
@@ -38,14 +38,26 @@ builder()
...
-builder()
+const mng = builder()
     .clusterProvider(new bp.MngClusterProvider({
-        ...publicCluster
+        ...publicCluster, 
+
+        launchTemplate: {
+            requireImdsv2: true, 
+            tags: {
+                "cost-center": "2122",
```
elamaran11, line 48: WHy is cost center required?

elamaran11 (commented):
LG, One comment.

examples/examples.ts
shapirov103: not requited. it is an example of using LT tags

elamaran11: ah ok. I think this should go to documentation too.

shapirov103: we have docs on LT tags, but not example. I used to based on the defect brought up by the customer. I will eventually add docs for all examples in this repo. In general, we tend to use patterns repo for such.

elamaran11 (approved):
LGTM

Issue #969, opened 2024-03-27
(addons/karpenter): Deployment fails due to missing CRD 


flo-mair:
### Describe the bug

The blueprint deployment fails due to missing CRD for NodePool:

`Message returned: Error: b'error: resource mapping not found for name: "default-nodepool" namespace: "" from "/tmp/manifest.yaml": no matches for kind "NodePool" in version "karpenter.sh/v1beta1"\nensure CRDs are installed first\n'
`

### Expected Behavior

Succesfull deployment

### Current Behavior

Message returned: Error: b'error: resource mapping not found for name: "default-nodepool" namespace: "" from "/tmp/manifest.yaml": no matches for kind "NodePool" in version "karpenter.sh/v1beta1"\nensure CRDs are installed first\n'

### Reproduction Steps

Deps:

 "@aws-quickstart/eks-blueprints": "^1.14.0",
 "aws-cdk-lib": "2.133.0",

```typescript 
new blueprints.addons.KarpenterAddOn({
        version: 'v0.34.3',
        nodePoolSpec: {
            requirements: [
                {key: 'karpenter.k8s.aws/instance-category', operator: 'In', values: ['t']},
                {key: 'kubernetes.io/arch', operator: 'In', values: [ 'arm64']},
                {key: 'karpenter.sh/capacity-type', operator: 'In', values: ['spot']},
            ],
            disruption: {
                consolidationPolicy: 'WhenEmpty',
                consolidateAfter: '30s',
                expireAfter: '20m'
            }
        },
        ec2NodeClassSpec: {
            amiFamily: 'AL2',
            subnetSelectorTerms: [{tags: {'Name': `${stackName}/${stackName}-vpc/PrivateSubnet*`}}],
            securityGroupSelectorTerms: [{tags: {'aws:eks:cluster-name': clusterName}}],
        },
        interruptionHandling: true,
    })


```

### CDK CLI Version

2.134.0

### EKS Blueprints Version

1.14.1

### Node.js Version

v18.17.0

### Environment details (OS name and version, etc.)

macOS

shapirov103:
@flo-mair making sure I understand - this is brand new cluster that you are provisioning and not an upgrade of an existing cluster, e.g. moving from lower versions of Karpenter to post 0.32?

flo-mair:
yes this is a new cluster

jsamuel1:
on a new cluster, the Karpenter helm chart should always be installing the CRDs on first install only.
[...]

PR #1129, opened 2025-04-04
Addresses #969. Added Stable V1 Karpenter addon,

shapirov103:
modiied tests, examples and docs. CDK moved to the 2.187

*Issue #, if available:* #969

*Description of changes:*


shapirov103:
@elamaran11 please review/approve when you get a chance.

.devcontainer/devcontainer.json
```diff
@@ -6,7 +6,7 @@
       },
       "ghcr.io/devcontainers/features/aws-cli:1": {},
       "ghcr.io/devcontainers-contrib/features/aws-cdk:2": {
-        "version": "2.173.4"
```
elamaran11, line 9: Could we find a way to parameterize this vs changing in so many files?

elamaran11 (approved):
LGTM. Changes look good except for one minor comment. Since this is lot of changes to version of cdk, can we do another round of regression with 1.30 version before we merge?

.devcontainer/devcontainer.json
shapirov103: I had similar thoughts, the reason why it has not happened - it takes 10 sec to do global replace now. Parameterizing this as well as references in the doc will take some time and debugging. However, agree with the suggestion, it is a good first issue for someone to start exploring the codebase.

shapirov103: Created https://github.com/aws-quickstart/cdk-eks-blueprints/issues/1133

Issue #1104, opened 2024-12-04
EKS Auto Mode please


taivo123:
### Describe the feature

EKS Auto Mode please

### Use Case

I'm always frustrated when I have to do more work than needed, EKS Auto Mode helps alleviate that :) 

### Acknowledgements

- [ ] I may be able to implement this feature request

### CDK version used

latest

### Node.js Version

latest

### Environment details (OS name and version, etc.)

Mac

shapirov103:
Waiting for at least L1 constructs to be available in CDK.

ruchirshetye-aws:
Hello team, I see, L1 [CfnCluster](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_eks.CfnCluster.html) now supports enabling Auto Mode. 

```
new eks.CfnCluster(this, 'MyCfnCluster', {
  resourcesVpcConfig: {
....
...
  computeConfig: {
    enabled: true,
    nodePools: ['nodePools'],
    nodeRoleArn: 'nodeRoleArn',
  },
 ```
 Can this be prioritized for EKS Blueprints?

shapirov103:
All clusters providers are using the L2 construct to provision the cluster. Please upvote this issue to add this support to the blueprints: https://github.com/aws/aws-cdk/issues/32364

Changing the cluster provider to use L1 will require a bit of refactoring, custom cluster provider that will use L1 is possible but will require L1 to create cluster and then use something like an importclusterprovider.

shapirov103:
Addressed with #1135. Please see https://awslabs.github.io/cdk-eks-blueprints/cluster-providers/automode-cluster-provider/

Please feel free to reopen if you see any issues.

PR #1135, opened 2025-04-15
Enable using EKS Automode with the EKS v2 CDK alpha.

zjaco13:
*Issue #, if available:*
#1104
*Description of changes:*
Adds functionality for EKS Automode.


shapirov103:
@zjaco13 please merge from main and resolve conflicts and big thanks for the PR, looks like a lot of stuff, will review shortly.

lib/cluster-providers/automode-cluster-provider.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  nodeRole?: IRole
+
+  tags?: {
+    [key: string]: string;
+  }
+
+}
+
+export class AutomodeClusterProvider extends GenericClusterProvider {
```
shapirov103, line 25: It seems cleaner to create a GenericV2ClusterProvider first and then derive automode cluster provider from it. It creates a cleaner separation of concerns and less spagetti if/then checking is it v2 or v1 and/or automode. 
Happy to discuss.

lib/addons/vpc-cni/index.ts
```diff
@@ -375,6 +375,7 @@ export class VpcCniAddOn extends CoreAddOn {
     (this.coreAddOnProps.configurationValues as any) = populateVpcCniConfigurationValues(props);
   }
 
+  @autoMode()
```
shapirov103, line 378: does not seem like we should fail here, it is in the list of addons that must be above a particular version. So for vpc-cni minimum version should be 1.19.0-eksbuild.1. It seems like adjust the decorator to take the expected min version would be a better path, e.g. conflictsWithAutomode(version: version).

shapirov103 (commented):
@zjaco13 awesome work. So far I posted mostly design/structural comments. The main one is:
> It seems cleaner to create a GenericV2ClusterProvider first and then derive automode cluster provider from it. It creates a cleaner separation of concerns and less spagetti if/then checking is it v2 or v1 and/or automode.
Happy to discuss.

Please see other comments and let's come up with an upgrade strategy. I have two options: 

1. we move to eks-v2 version in a new major version of blueprints, like 2.x
2. we split eks-v2 version to a separate cluster provider and use it for automode with restrictions now. Gradually improve v2 cluster provider functionality and then eventually deprecate genericClusterProvider.  

Leaning towards 2 as eks-v2 CFN construct is still in alpha. 

Also consider the 3d option:

1. we change nothing in the cluster providers. For automode, we create a separate stack that will import the cluster into v2 construct and just enable automode. This may need more thought though.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -83,6 +85,11 @@ export interface GenericClusterProviderProps extends Partial<eks.ClusterOptions>
      */
     autoscalingNodeGroups?: AutoscalingNodeGroup[];
 
+    /**
+     * EKS Automode compute config
+     */
+    compute?: eksv2.ComputeConfig;
```
shapirov103, line 91: why do we need this compute config for the generic cluster provider which is not V2?

lib/cluster-providers/generic-cluster-provider-v2.ts
```diff
@@ -0,0 +1,417 @@
...
+    return new ClusterBuilderV2();
+}
+
+/**
+ * Function that contains logic to map the correct kunbectl layer based on the passed in version.
+ * @param scope in whch the kubectl layer must be created
+ * @param version EKS version
+ * @returns ILayerVersion or undefined
+ */
+function selectKubectlLayer(scope: Construct, version: eks.KubernetesVersion): ILayerVersion | undefined {
```
shapirov103, line 36: I realize we replicate the functionality with V2, but let's move these generic functions out to reuse in both. Otherwise it is a maintenance issue.

lib/cluster-providers/generic-cluster-provider.ts
```diff
@@ -244,21 +251,28 @@ export class GenericClusterProvider implements ClusterProvider {
 
         this.validateInput(props);
 
-        assert(!(props.managedNodeGroups && props.managedNodeGroups.length > 0
-            && props.autoscalingNodeGroups && props.autoscalingNodeGroups.length > 0),
-            "Mixing managed and autoscaling node groups is not supported. Please file a request on GitHub to add this support if needed.");
+        const computeTypesEnabled = [
```
shapirov103, line 254: if computeTypes is removed from this implementation, we don't need to worry about this validation either.

lib/cluster-providers/mng-cluster-provider.ts
```diff
@@ -17,7 +17,7 @@ export interface MngClusterProviderProps extends Partial<eks.CommonClusterOption
     name?: string;
 
     /**
-     * In this case id is optional and defaults tp the cluster name
+     * In this case id is optional and defaults to the cluster name
      */
     id?: string,
```
shapirov103, line 22: minor: since you touched this file, let's replace endings for interface properties from "," to ";". It must be semicolon.

lib/teams/team.ts
```diff
@@ -128,20 +129,30 @@ export class ApplicationTeam implements Team {
...
             return;
         }
-        const eksCluster : Cluster = clusterInfo.cluster;
-        const awsAuth = eksCluster.awsAuth;
-
         const users = this.teamProps.users ?? [];
         const teamRole = this.getOrCreateRole(clusterInfo, users, props.userRoleArn);
 
-        if (teamRole) {
+        if(clusterInfo.cluster instanceof eksv2.Cluster){
```
shapirov103, line 139: This seems applicable to teams as a whole provided they use access entries. Configmap is a legacy approach. We can defer this to later as well, provided it is captured as an issue to refactor teams to support access entries (keeping config map as supported legacy option).

shapirov103 (commented):
@zjaco13 looks much better/cleaner. 
I added a few comments, nothing major. 

Do you mind adding an example I can test to examples/examples.ts that I run with `npm run examples`? I don't see a way atm to include it into the e2e test, but once we move this out to awslabs org, we can add many blueprints and test each of them on e2e. 

Also given the design is good, we can turn it into an actual PR unless you see any issues with the implementation that may prevent us from releasing this  with the next release.

zjaco13:
> @zjaco13 looks much better/cleaner. I added a few comments, nothing major.
> 
> Do you mind adding an example I can test to examples/examples.ts that I run with `npm run examples`? I don't see a way atm to include it into the e2e test, but once we move this out to awslabs org, we can add many blueprints and test each of them on e2e.
> 
> Also given the design is good, we can turn it into an actual PR unless you see any issues with the implementation that may prevent us from releasing this with the next release.

I do want to chat about some implementation details before merge.  Will add an example to the examples dir.

lib/cluster-providers/generic-cluster-provider-v2.ts
zjaco13: resolved

lib/cluster-providers/generic-cluster-provider.ts
zjaco13: removed

lib/cluster-providers/generic-cluster-provider.ts
zjaco13: resolved

lib/cluster-providers/mng-cluster-provider.ts
zjaco13: resolved

lib/teams/team.ts
zjaco13: will create issue

zjaco13: https://github.com/aws-quickstart/cdk-eks-blueprints/issues/1140

shapirov103 (commented):
All great, let's add a doc under cluster providers (overview and separate page): https://aws-quickstart.github.io/cdk-eks-blueprints/cluster-providers/

Once we get initial feedback, I am thinking to also update the getting started guide in the next minor version.

I will wait for the nodepool options to be added as well to the automode cluster provider so that we have parity with karpenter.

examples/examples.ts
```diff
@@ -87,6 +87,11 @@ builder()
         .compatibilityMode(false)
         .build(app, 'eks-blueprint');
 
+// Autmode cluster
+builder()
```
shapirov103, line 91: Why not use AutomodeBuilder here?

shapirov103: Also have you tried running it? Examples can be run with `npm run examples list` for example, or `npm run examples deploy <stack-name>`. I am getting a failure for this cluster: ValidationError: Lambda Functions in a public subnet can NOT access the internet. If you are aware of this limitation and would still like to place the function in a public subnet, set `allowPublicSubnet` to true

Although other examples that use default VPC public subnets seem to work fine. Could be a v2 issue.

examples/examples.ts
```diff
@@ -87,6 +87,11 @@ builder()
         .compatibilityMode(false)
         .build(app, 'eks-blueprint');
 
+// Autmode cluster
```
shapirov103, line 90: Automode

zjaco13: resolved

examples/examples.ts
zjaco13: Haven't seen that before, seems to be a V2 change.  May need to open an issue on the repo.  Switched it to AutomodeBuilder with its own vpc

docs/builders/automode-builder.md
```diff
@@ -0,0 +1,88 @@
...
+                new blueprints.addons.IstioBaseAddOn(),
+                new blueprints.addons.IstioControlPlaneAddOn(),
+                new blueprints.addons.KubeStateMetricsAddOn(),
+                new blueprints.addons.MetricsServerAddOn(),
+                new blueprints.addons.PrometheusNodeExporterAddOn(),
+                new blueprints.addons.ExternalsSecretsAddOn(),
+                new blueprints.addons.SecretsStoreAddOn(),
+                new blueprints.addons.CalicoOperatorAddOn(),
+                new blueprints.addons.CertManagerAddOn(),
+                new blueprints.addons.AdotCollectorAddOn(),
```
shapirov103, line 73: Was this specific blueprint tested with all these addons? I recall Ela mentioned ADOT had compatibility issues with auto mode, but if it works in your test we can keep it.

docs/builders/automode-builder.md
```diff
@@ -0,0 +1,88 @@
...
+The below usage helps you with a demonstration to use `AutomodeBuilder` to configure a required setup as you prepare a blueprint for setting up EKS Auto Mode on a new EKS cluster.
+
+```typescript
+import * as blueprints from "@aws-quickstart/eks-blueprints";
+import { AutomodeBuilder } from "@aws-quickstart/eks-blueprints";
+import { CfnWorkspace } from "aws-cdk-lib/aws-aps";
+import * as eks from "aws-cdk-lib/aws-eks";
+import { Construct } from "constructs";
+
+export default class AutomodeConstruct {
```
shapirov103, line 28: minor: Unless you plan to move it as is to the patterns repo, I suggest we drop the outer AutomodeConstruct class. If you will add it to the patterns, where we use this convention, we can keep as is.

docs/builders/automode-builder.md
```diff
@@ -0,0 +1,88 @@
...
+## Input Parameters
+
+`Partial<AutomodeClusterProviderProps>` parameters can be used as inputs to `AutomodeBuilder`. Few parameters shown below:
+
+- `version` : Kubernetes version to use for the cluster
+- `nodePools`: The Auto Mode node pools to provision for the cluster
+
+### Demonstration - Running EKS Auto Mode
+
+The below usage helps you with a demonstration to use `AutomodeBuilder` to configure a required setup as you prepare a blueprint for setting up EKS Auto Mode on a new EKS cluster.
```
shapirov103, line 19: Minor (use simpler style): The below usage demonstrates how to use `AutomodeBuilder` to set up EKS Auto Mode on a new EKS cluster.

shapirov103 (commented):
Great work! Added a few minor comments on the docs. It should be ready to merge after they are addressed as I also validated functionally. 

I have another PR #1141 for this functionality as I noticed that we have a lot of dependencies expressed as `@dependable(AwsLoadBalancerControllerAddOn.name)` and such which prevents using standard addons like nginx with the auto mode. I will push that change after merging yours.

shapirov103:
@zjaco13 please also run `make lint` on the codebase, I believe I saw some minor stylistic warnings.

docs/builders/automode-builder.md
zjaco13: I have not tested this specific blueprint

shapirov103: let's not publish it then, customers will literally copy paste and file issues - we may look bad in this case. either we test it, or replace with something that we know works. E.g. what i used in my PR

```ts
bp.EksBlueprint.builder()
  .account(process.env.CDK_DEFAULT_ACCOUNT)
  .region(process.env.CDK_DEFAULT_REGION)
  .resourceProvider(bp.GlobalResources.Vpc, new bp.VpcProvider("vpc-07b9e40ea515577e5"))
  .clusterProvider(new bp.AutomodeClusterProvider({version: KubernetesVersion.V1_32}))
  .addOns(
    new IngressNginxAddOn({
      crossZoneEnabled: true,
      internetFacing: true,
      targetType: "ip",
    })
  )
  .build(app, "eksv2-blueprint");
```

docs/builders/automode-builder.md
shapirov103: OK, we can do after the release.

docs/builders/automode-builder.md
zjaco13: switched to this example

zjaco13:
> @zjaco13 please also run `make lint` on the codebase, I believe I saw some minor stylistic warnings.

should be resolved

docs/cluster-providers/generic-cluster-provider-v2.md
```diff
@@ -0,0 +1,88 @@
+# Generic Cluster Provider V2
+
+The `GenericClusterProviderV2` allows you to provision an EKS cluster which leverages one or more [EKS managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html)(MNGs), one or more autoscaling groups[EC2 Auto Scaling groups](https://docs.aws.amazon.com/autoscaling/ec2/userguide/AutoScalingGroup.html), or EKS AutoMode for its compute capacity. Users can also configure multiple Fargate profiles along with the EC2 based compute capacity.
```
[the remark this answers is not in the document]
zjaco13: done

docs/builders/automode-builder.md
zjaco13: done

shapirov103 (approved):
LGTM
Validate functionally in a standalone account using the example provided.

PR #1138, opened 2025-04-22
Fix codeblock syntax in pipelines.md

muckelba:
I just noticed that while reading the docs so i figured i can just fix it.


shapirov103 (approved):
LGTM

Issue #1155, opened 2025-05-27
addons/aws-loadbalancer-controller: Missing elasticloadbalancing:SetRulePriorities in Service Account IAM policy


sreeteja06:
### Describe the bug

The AWS Load Balancer Controller addon in CDK EKS Blueprints is missing the `elasticloadbalancing:SetRulePriorities` permission in its default IAM policy. This causes failures when creating Ingress resources that use ALB grouping annotations (e.g., `alb.ingress.kubernetes.io/group.name` and `alb.ingress.kubernetes.io/group.order`), as the controller requires this permission to manage rule priorities across grouped ALBs.

This permission is included in the official IAM policy recommended in the AWS documentation for Helm installations [https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy.json) which is referenced here [https://docs.aws.amazon.com/eks/latest/userguide/lbc-helm.html](https://docs.aws.amazon.com/eks/latest/userguide/lbc-helm.html), but appears to be missing from the CDK Blueprints addon implementation [cdk-eks-blueprints/lib/addons/aws-loadbalancer-controller
/iam-policy.ts](https://github.com/awslabs/cdk-eks-blueprints/blob/main/lib/addons/aws-loadbalancer-controller/iam-policy.ts).

### Expected Behavior

When creating an ingress resource with `group.order` annotation, the AWS Load Balancer Controller should be able to set rule priorities in the ALB without any permission errors.

### Current Behavior

When creating an ingress resource with `group.order` annotation, the AWS Load Balancer Controller fails with the following error:

> Failed deploy model due to operation error Elastic Load Balancing v2: SetRulePriorities, https response error StatusCode: 403, RequestID: a45a5983-7fc7-443e-acf4-9c15ecd20f0b, api error AccessDenied: User: arn:aws:sts::<account-id>:assumed-role/<role-name> is not authorized to perform: elasticloadbalancing:SetRulePriorities on resource: arn:aws:elasticloadbalancing:<region>:<account-id>:listener-rule/<alb-id>/<listener-id>/<rule-id>

### Reproduction Steps

1. Deploy an EKS cluster using CDK EKS Blueprints with AWS Load Balancer Controller addon
2. Create an ingress resource with the following annotations:
   ```yaml
[...]

PR #1162, opened 2025-06-11
[addon] Update aws-load-balancer-controller IAM policy

vumdao:
*Issue #1155*

*Description of changes:*

Add missing IAM policy referenced from https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/release-2.12/docs/install/iam_policy.json


shapirov103 (approved):
Thank you, @vumdao, much appreciated and long overdue from our side. Will include in the next patch release shortly.

PR #1163, opened 2025-06-17
Log addon name that conflicts with auto-mode

SomeoneWeird:
*Issue #, if available:* N/A

*Description of changes:*

This changes the warnings/errors that are thrown when an addon conflicts with an auto-mode cluster (or the version). I was getting very frustrated figuring out *which* add-on was the culprit :)

### DIFF
--- a/lib/utils/addon-utils.ts
+++ b/lib/utils/addon-utils.ts
@@ -118,7 +118,7 @@ function parseEksVersion(version: string): [string, number] {
 }
 
 
-export function conflictsWithAutoMode(minExpectedVersion: string | null) {
+export function conflictsWithAutoMode(addOn: string, minExpectedVersion: string | null) {
   // eslint-disable-next-line @typescript-eslint/ban-types
   return function(target: Object, key: string | symbol, descriptor: PropertyDescriptor) {
     const originalMethod = descriptor.value;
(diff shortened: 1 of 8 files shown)


zjaco13 (approved):
LGTM

zjaco13 (changes requested):
@SomeoneWeird Please take a look at my comment!

lib/utils/addon-utils.ts
```diff
@@ -118,7 +118,7 @@ function parseEksVersion(version: string): [string, number] {
 }
 
 
-export function conflictsWithAutoMode(minExpectedVersion: string | null) {
+export function conflictsWithAutoMode(addOn: string, minExpectedVersion: string | null) {
   // eslint-disable-next-line @typescript-eslint/ban-types
   return function(target: Object, key: string | symbol, descriptor: PropertyDescriptor) {
     const originalMethod = descriptor.value;
```
shapirov103, line 124: In here you can get the addon class with 
```ts
  const addOnName = this.constructor.name;
```

This will make it more generic, and avoid a case when developer can accidentally pass the wrong name when using the decorator.

shapirov103:
Addressed by #1169. Your contribution was helpful to identify the issue and provide the solution.

Issue #1168, opened 2025-06-26
Auto-mode: core addons fail with a runtime when version is specified


shapirov103:
### Describe the bug

In automode if you add core addOns (coredns for examples) to the blueprint the synth will fail with 

```
TypeError: Cannot read properties of undefined (reading 'match')
    at parseEksVersion (/Users/shapirov/projects/eks-blueprints/lib/utils/addon-utils.ts:113:25)
    at compareEksVersions (/Users/shapirov/projects/eks-blueprints/lib/utils/addon-utils.ts:100:29)
    at CoreDnsAddOn.descriptor.value (/Users/shapirov/projects/eks-blueprints/lib/utils/addon-utils.ts:141:16)
    at new EksBlueprintConstruct (/Users/shapirov/projects/eks-blueprints/lib/stacks/eks-blueprint-construct.ts:270:34)
    at new EksBlueprint (/Users/shapirov/projects/eks-blueprints/lib/stacks/eks-blueprint-stack.ts:60:39)
    at AutomodeBuilder.build (/Users/shapirov/projects/eks-blueprints/lib/stacks/eks-blueprint-stack.ts:23:16)
    at Object.<anonymous> (/Users/shapirov/projects/eks-blueprints/examples/examples.ts:116:4)
    at Module._compile (node:internal/modules/cjs/loader:1554:14)
    at Module.m._compile (/Users/shapirov/projects/eks-blueprints/node_modules/ts-node/src/index.ts:1618:23)
    at node:internal/modules/cjs/loader:1706:10
```

### Expected Behavior

Should succeed or fail gracefully.

### Current Behavior

Fails with the error. 

### Reproduction Steps

```ts
bp.AutomodeBuilder.builder({
  version: KubernetesVersion.V1_31,
  nodePools: ["system", "general-purpose"],
})
  .account(process.env.CDK_DEFAULT_ACCOUNT)
  .region(process.env.CDK_DEFAULT_REGION)
  .addOns(
    new AwsLoadBalancerControllerAddOn(),
    new bp.CoreDnsAddOn(),
    new IngressNginxAddOn({
      crossZoneEnabled: true,
      internetFacing: true,
      targetType: "ip",
    })
  )
  .build(app, "eksv2-blueprint");
```

### Possible Solution

for non-helm addons version is not a member variable (this potentially was introduced with the support of "auto" versioning). 

### CDK CLI Version

2.1012

### EKS Blueprints Version

1.17.1

[...]

PR #1169, opened 2025-06-26
fix version error for automode addons

zjaco13:
*Issue #, if available:*
#1168
*Description of changes:*
fixes error when version is null or auto

### DIFF
--- a/lib/utils/addon-utils.ts
+++ b/lib/utils/addon-utils.ts
@@ -126,21 +126,25 @@ export function conflictsWithAutoMode(minExpectedVersion: string | null) {
     descriptor.value = function(this: any, ...args: any[]) {
       const clusterInfo: ClusterInfo = args[0];
       const stack = clusterInfo.cluster.stack.stackName;
+      const addonName = this.constructor.name;
       if (!clusterInfo.autoMode) {
         return originalMethod.apply(this, args);
       }
-      if (minExpectedVersion === "fail") {
-        throw new Error(`Deploying ${stack} failed. This add-on is already available on the cluster with EKS Auto Mode.`);
+      let version = null;
+      if('getAddonVersion' in this && typeof this.getAddonVersion === 'function'){
+        version = this.getAddonVersion();
       }
-      else if (minExpectedVersion == null || this.version == null || this.version == "auto") {
-        logger.warn(`This add-on is already available on the cluster with EKS Auto Mode.`);
+      if (minExpectedVersion === "fail") {
+        throw new Error(`Deploying ${stack} failed. Add-on ${addonName} is already available on the cluster with EKS Auto Mode.`);
+      }
+      else if (minExpectedVersion == null || version ==  "auto") {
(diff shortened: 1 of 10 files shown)


lib/utils/addon-utils.ts
```diff
@@ -132,7 +132,7 @@ export function conflictsWithAutoMode(minExpectedVersion: string | null) {
       if (minExpectedVersion === "fail") {
         throw new Error(`Deploying ${stack} failed. This add-on is already available on the cluster with EKS Auto Mode.`);
       }
-      else if (minExpectedVersion == null) {
+      else if (minExpectedVersion == null || this.version == null || this.version == "auto") {
```
shapirov103, line 135: this fixes the runtime but incorrect functionally. 
CoreDNS has the following decorator:  @utils.conflictsWithAutoMode("v1.9.3-eksbuild.7")
The expectation is that  it should compare the versions. 

So for all the derivatives of CoreAddOn - I suppose we can also add version as the member variable.

I propose to keep this fix and apply the change to the core addons so that we can produce meaningful messages when needed. E.g. if they deploy coredns explicitly with the version that does not conflict, we should not warn.

shapirov103 (commented):
Please see my comment. Overall agree with the direction to an extent.

lib/utils/addon-utils.ts
```diff
@@ -126,21 +126,25 @@ export function conflictsWithAutoMode(minExpectedVersion: string | null) {
     descriptor.value = function(this: any, ...args: any[]) {
       const clusterInfo: ClusterInfo = args[0];
       const stack = clusterInfo.cluster.stack.stackName;
+      const addonName = this.constructor.name;
       if (!clusterInfo.autoMode) {
         return originalMethod.apply(this, args);
       }
+      let version = null;
```
shapirov103, line 133: let's use minExpectedVersion?  - optional parameter, rather than allowing null explicitly.

lib/utils/cluster-utils.ts
```diff
@@ -52,6 +53,27 @@ interface Tag {
   });
 }
 
+export function checkEKSVersion(minVersion: KubernetesVersion, clusterVersion: KubernetesVersion): boolean {
```
shapirov103, line 56: my recommendation on books - [Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882).

if I look at the checkEKSVersion function, I have no idea what it is checking. I can discover it is a boolean, but unclear what true | false means in this context. 

Initially, I thought a more correct way to name this: `isEKSVersionSupportedForAutomode()`.  

Seeing, however, that you are passing both minVersion and targetVersion, I suspect it is just a semantic version comparison, like `compareVersions` with standard -1, 0, 1 semantics. 

Why not just use `import semver`? We use it in bunch of places like `semver.lt()` or `semver.gt()`.

there is also compareEksVersions in the same file - did you find something wrong with it?

shapirov103 (commented):
Looks good functionally, added a few tactical comments.

lib/utils/cluster-utils.ts
zjaco13: compareEksVersions is a checker for the addon versions that have eks in the version, making it incompatible with semver.  Will rename that function and get rid of this one to replace with semver

shapirov103 (approved):
LGTM

PR #1176, opened 2025-07-19
create ingressclass addon for use with automode

zjaco13:
*Issue #, if available:*

*Description of changes:*
With AutoMode, ALB Ingress Controller is automatically installed, but no ingress class is created by default.  This addon creates a default ingress class for use with AutoMode


lib/addons/aws-alb-default-ingress-class/index.ts
```diff
@@ -0,0 +1,30 @@
+import { ClusterAddOn, ClusterInfo } from '../../spi';
+import { KubernetesManifest } from 'aws-cdk-lib/aws-eks';
+import { supportsALL } from '../../utils';
+
+
+@supportsALL
```
elamaran11, line 6: We need to add ALB LB Controller Addon as a dependency to this Addon. Or this will create issues for Non AutoMode clusters

elamaran11 (changes requested):
@zjaco13 Thankyou for the PR. Firstly i would question the need for a separate addon for this. Could we not add this as an additional resource created as part of `automode-builder`. I also see this potentially breaking if used for other non-automode clusters. Secondly this also causes confusion when this addon is used for non-automode clusters and user misses this. Third how will you handle the dependency of ALB Addon if used for non-auto mode clusters
Few other minor comments if you still want to make this as addon hashing out challenges above:
1. Addon not added to E2E.
2. Please Add unit test cases
3. mkdocs needs to be updated. 
2.

lib/addons/aws-alb-default-ingress-class/index.ts
zjaco13: @shapirov103 @elamaran11
- Added ALB controller as a conflict
-  created unit tests
- added to the automodebuilder e2e example
- updated docs to specify that the addon is specifically for auto mode
- added a function to the automode-builder to specifically add this addon.

Issue #1216, opened 2025-12-15
EKS Capabilities support


zjaco13:
### Describe the feature

https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-eks-capability.html

### Use Case

https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-eks-capability.html

### Acknowledgements

- [x] I may be able to implement this feature request

### CDK version used

latest

### Node.js Version

latest

### Environment details (OS name and version, etc.)

latest

PR #1219, opened 2025-12-19
Add EKS Capabilities feature

zjaco13:
*Issue #, if available:*
#1216 
*Description of changes:*
Adds a new module for creating EKS Capabilities as part of the framework.  Available to customers as a part of EKS Builder


shapirov103 (commented):
added a few high level comments. Overall, I like the structure that you proposed, it is consistent with the approaches currently applied. There are some trade-offs here and there that I brought up.

lib/stacks/eks-blueprint-construct.ts
```diff
@@ -286,6 +297,12 @@ export class EksBlueprintConstruct extends Construct {
             constructs.forEach((construct, index) => {
                 this.clusterInfo.addProvisionedAddOn(addOnKeys[index], construct);
             });
+            
+            if (blueprintProps.capabilities != null) {
+                for (let capability of blueprintProps.capabilities) {
+                    capability.create(this.clusterInfo);
```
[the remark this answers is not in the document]
zjaco13: I will do this, this way I can also ensure that addons + capabilities that do the same thing don't accidentally get deployed

lib/capabilities/ack-capability.ts
```diff
@@ -0,0 +1,33 @@
+import { CapabilityProps, Capability } from "./capability";
+import { CapabilityType } from "../spi";
+
+/**
+ * Properties for ACK capability, excluding namespace, type, and identityCenterArn
+ * which are not applicable for ACK capabilities.
+ */
+export type AckCapabilityProps = Omit<CapabilityProps, "namespace" | "type" | "identityCenterArn">;
```
shapirov103: is there a reason this specifies namespace and identityCenterArn when I don't see them in the base type in capabilityprops?

lib/capabilities/ack-capability.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * AWS Controllers for Kubernetes (ACK) capability for EKS clusters.
+ * Enables direct management of AWS services from Kubernetes using custom resources.
+ * Uses AdministratorAccess policy by default for broad AWS service management.
+ * ```
```
shapirov103, line 53: minor: dangling "```"  - remove

lib/utils/addon-utils.ts
```diff
@@ -193,6 +193,29 @@ export function mustRunOnAutoMode() {
 
 }
 
+export function conflictsWithCapabilities(...capabilities: string[]) {
+  return function (target: object, key: string | symbol, descriptor: PropertyDescriptor) {
+    const originalMethod = descriptor.value;
+
+    descriptor.value = function( ...args: any[]) {
+      // const dependencies: (Promise<Construct> | undefined)[] = [];
```
shapirov103: do we need the commented out code? drop if not needed

lib/capabilities/argocd-capability.ts
```diff
@@ -0,0 +1,149 @@
...
+    capability.configuration = {
+      argoCd: {
+        awsIdc: {
+          idcInstanceArn: this.options.idcInstanceArn,
+          idcManagedApplicationArn: this.options.idcManagedApplicationArn,
+          idcRegion: this.options.idcRegion
+        },
+        namespace: this.options.namespace,
+        networkAccess: {
+          vpceIds: this.options.networkAccessVpcEndpoints?.map(e => e.vpcEndpointId)
```
shapirov103: if networkAccessVpcEndpoints are not provided, this will have networkAccess: { vpceIds: undefined} - has it been tested?

docs/capabilities/ack-capability.md
```diff
@@ -0,0 +1,36 @@
...
+
+| Property | Type | Default | Description |
+|----------|------|---------|-------------|
+| `capabilityName` | `string` | `blueprints-ack-capability` | Name for the capability resource |
+| `roleArn` | `string` | Auto-created | Existing IAM role ARN to use |
+| `policyName` | `string` | - | Custom managed policy name |
+| `policyDocument` | `iam.PolicyDocument` | - | Custom inline policy |
+| `tags` | `CfnTag[]` | - | CloudFormation tags |
+
+The default IAM policy is `AdministratorAccess`. For production, scope this down to only the AWS services you need:
```
shapirov103: default IAM policy of Administrator - is it what EKS Blueprints provides or is it default for EKS provided there is no way to tell which service they will need to use through ACK?

lib/capabilities/ack-capability.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export class AckCapability extends Capability {
+  /** Default AWS managed policy providing full administrative access */
+  readonly DEFAULT_POLICY_NAME = "AdministratorAccess";
```
shapirov103: Nm my other comment, I see that in this policy is set by us and provide very broad permissions. While I understand the challenge, we should choose safety, so I suggest make this field mandatory, provide and example of setting it to Administrator (make it an opt in) and then comment in the doc that you have  - they need to scope down or the need to use IAM Role Selectors that EKS recommends https://docs.aws.amazon.com/eks/latest/userguide/capability-role.html (pseudo code):

```ts
import * as iam from 'aws-cdk-lib/aws-iam';
import * as blueprints from '@aws-quickstart/eks-blueprints';

const app = new cdk.App();

   // Capability role only needs sts:AssumeRole to the service-specific roles
   const ackPolicy = new iam.PolicyDocument({
     statements: [
       new iam.PolicyStatement({
         actions: ['sts:AssumeRole', 'sts:TagSession'],
         resources: [
           'arn:aws:iam::<ACCOUNT_ID>:role/ACK-S3-Role',
           'arn:aws:iam::<ACCOUNT_ID>:role/ACK-RDS-Role',
         ],
       }),
     ],
   });

   const stack = blueprints.EksBlueprint.builder()
     .account('<ACCOUNT_ID>')
     .capabilities({
       ack: new blueprints.capabilities.AckCapability({
         policyDocument: ackPolicy,
       }),
     })
     .build(app, 'my-cluster');
```
   Then after the cluster is deployed, you'd create the IAMRoleSelector resources in the cluster to map those roles to namespaces:
```yaml
   apiVersion: services.k8s.aws/v1alpha1
   kind: IAMRoleSelector
   metadata:
     name: s3-role-selector
   spec:
     arn: arn:aws:iam::<ACCOUNT_ID>:role/ACK-S3-Role
     namespaceSelector:
       names:
         - s3-resources
   ---
   apiVersion: services.k8s.aws/v1alpha1
   kind: IAMRoleSelector
   metadata:
     name: rds-role-selector
   spec:
     arn: arn:aws:iam::<ACCOUNT_ID>:role/ACK-RDS-Role
     namespaceSelector:
       names:
         - rds-resources
```
The service-specific roles (ACK-S3-Role, ACK-RDS-Role) need a trust policy allowing the Capability Role to assume them, and their own permission policies scoped to the relevant AWS service.

We can mention this pattern in the docs, but it makes sense to think how to make it easy for customers to use this pattern.

lib/spi/capability-contracts.ts
```diff
@@ -0,0 +1,31 @@
...
+   * @param clusterInfo
+   * @returns cdk.Construct
+   */
+  create(clusterInfo: ClusterInfo): Construct
+}
+
+/**
+ * Typed map of EKS capabilities. Each key can only appear once.
+ */
+export interface ClusterCapabilities {
```
shapirov103, line 27: let's say tomorrow EKS adds another capability. Some users will ask how to use it - what will be the approach (without us releasing the new version - we can't sync our release with EKS).

shapirov103 (commented):
Looks great, I like the design. Added a few comments and questions.

lib/capabilities/ack-capability.ts
zjaco13: They are only used for ArgoCD but technically part of overall props: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_eks.CfnCapabilityProps.html

lib/capabilities/ack-capability.ts
zjaco13: That makes sense, will update

lib/spi/capability-contracts.ts
zjaco13: was thinking something like this:

```
interface NewClusterCapabilites extends ClusterCapabilities {
    newCap?: Capability
}

const newClustCap: NewClusterCapabilites = {

}

const builder = EksBlueprint.builder().capabilities(newClustCap)
```

lib/capabilities/ack-capability.ts
zjaco13: done

lib/spi/capability-contracts.ts
```diff
@@ -0,0 +1,6 @@
+import { CfnCapability } from "aws-cdk-lib/aws-eks";
+import { ClusterInfo } from "./types";
+
+export declare interface ClusterCapability {
```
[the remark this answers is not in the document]
zjaco13: done

lib/utils/addon-utils.ts
zjaco13: copied from other conflictsWith, got rid in both

lib/capabilities/ack-capability.ts
shapirov103: Well, I am referring to the `export interface CapabilityProps` that you specify here. This interface does not have namespace and identity center - correctly so, since it is specific to only one case.

zjaco13: Ah yes, will get rid of those

shapirov103 (approved):
LGTM - awesome work!

shapirov103:
Functional Validation complete (e2e cluster test)

PR #1240, opened 2026-03-19
fix: use global SPN which is now supported in all regions

b177y:
SPN construction using `service + urlSuffix` is not valid in all partitions, including EUSC. 

See comment [here](https://github.com/aws/aws-cdk/blob/c68f2f5961b24fba3f4a4b769e355b00d91fd6a1/packages/aws-cdk-lib/region-info/lib/default.ts#L27):
> Service principals are now globally `<SERVICE>.amazonaws.com`, use iam.ServicePrincipal instead.


shapirov103 (approved):
LGTM

PR #1242, opened 2026-03-20
Bugfix/s3mountpoint fails rbac

shapirov103:
*Issue #, if available:*

*Description of changes:* Fixed S3 CSI driver issue - pod failing with no RBAC entries after helm added one more service account for the controller. Also move for 2.4.1

### DIFF
--- a/lib/addons/s3-csi-driver/index.ts
+++ b/lib/addons/s3-csi-driver/index.ts
@@ -3,7 +3,6 @@ import { ClusterInfo } from "../../spi";
 import { HelmAddOn, HelmAddOnUserProps } from "../helm-addon";
 import * as iam from 'aws-cdk-lib/aws-iam';
 import { createNamespace, setPath, supportsALL } from "../../utils";
-import { ReplaceServiceAccount } from "../../utils/sa-utils";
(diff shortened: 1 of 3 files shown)


lib/addons/s3-csi-driver/index.ts
```diff
@@ -3,6 +3,7 @@ import { ClusterInfo } from "../../spi";
 import { HelmAddOn, HelmAddOnUserProps } from "../helm-addon";
 import * as iam from 'aws-cdk-lib/aws-iam';
 import { createNamespace, setPath, supportsALL } from "../../utils";
+import { ReplaceServiceAccount } from "../../utils/sa-utils";
```
zjaco13: With the overwriteServiceAccount field in upstream, do we still need to use ReplaceServiceAccount?

shapirov103: Addressed. The codebase does not use it now. 
The class `ReplaceServiceAccount` will stay though, as it is public export and could have been used externally.

zjaco13 (dismissed):
LGTM

zjaco13 (approved):

PR #1246, opened 2026-04-08
add EKS Kubernetes 1.35 support

jcmtang:
- add Kubernetes 1.35 support
- added 1.35 version mappings for managed EKS add-ons
  - managed add-on versions for Kubernetes 1.35 were checked against `aws eks describe-addon-versions` and updated to the current AWS recommended default versions (e.g. `aws eks describe-addon-versions --addon-name adot --kubernetes-version 1.35 --query 'addons[0].addonVersions[?compatibilities[0].defaultVersion==`true`].addonVersion' --output text --region us-east-1`)
- updated cluster provider kubectl layer selection to use the v1.35 kubectl layer for Kubernetes 1.35

### DIFF
--- a/examples/eks-1-35.ts
+++ b/examples/eks-1-35.ts
@@ -1,0 +1,17 @@
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { KubernetesVersion } from 'aws-cdk-lib/aws-eks';
+import * as blueprints from '../lib';
+
+const app = new cdk.App();
+
+const account = process.env.CDK_DEFAULT_ACCOUNT;
+const region = process.env.CDK_DEFAULT_REGION;
+
+const stack = blueprints.EksBlueprint.builder()
+    .account(account)
+    .region(region)
+    .version(KubernetesVersion.V1_35)
+    .build(app, 'eks-1-35-debug-stack');
+
+void stack;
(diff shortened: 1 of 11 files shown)


shapirov103 (commented):
@jcmtang looks great, thank you for your contribution. 

In order to merge, I will need to validate that the functionality works and provisions a 1.35 cluster. Since we are not moving the e2e test under examples/blueprint-ipv4-construct/index.ts to 1.35 yet, I suggest an example approach that I can validate to make sure it works using the examples (or debug) approach described [here](https://github.com/awslabs/cdk-eks-blueprints/blob/main/docs/internal/readme-internal.md#customer-issue-debugging).

Once added, we will validate internally and merge. Thank you!

jcmtang:
I added an example file. Not entire sure if this was what you were looking for @shapirov103. Let me know if I need to make additional changes.

shapirov103 (approved):
LGTM
Validated functionally with the example, 1.35 cluster works fine.

andyt10:
I'm eyeing this PR off for a 1.35 upgrade and just wondered when/if the default Karpenter version for 1.35 is to be configured here: https://github.com/awslabs/cdk-eks-blueprints/blob/main/lib/addons/karpenter/index.ts#L23

I don't mind adding it, but I am not sure how this project determines which version to use.
Upgrade compatibilities say `>= 1.9`, and latest is `1.12` 
https://karpenter.sh/docs/upgrading/compatibility/

{4404FE86-2F05-4C61-973B-2AF8072EE33E}

shapirov103:
@andyt10 for KarpenterV1 we use 1.9 atm which should be compatible with the latest 3 supported EKS version (1.33+). Using the logic in the compatibility matrix you should be able to use the latest 1.12 with all these versions as well. 

If 1.12 works fine (which I fully expect it to), then the default version [here](https://github.com/awslabs/cdk-eks-blueprints/blob/main/lib/addons/karpenter/karpenter-v1.ts#L22) can be uprevved to 1.12.

PR #1254, opened 2026-06-18
Migrate to new eks-v2 cdk package

zjaco13:
Move all constructs from aws-eks to aws-eks-v2 cdk package (Cfn resources are still under aws-eks)

*Issue #, if available:*

*Description of changes:*


shapirov103:
Please resolve the conflicts. Great PR to merge once ready.

lib/cluster-providers/import-cluster-provider.ts
shapirov103, line 1: I assume this will be a major breaking change for any customer who is using the generic cluster provider, meaning existing stacks on the pipeline will fail the upgrade unless customers apply migration path, which we assessed and it is a serious undertaking for the customer. 
If this reasoning is correct, then customer will have to engage in the migration path or they will have to abandon there is existing deployments of stacks and redeploy as blue/green. This is not a minor version upgrade - it is major. 

So I propose  - keep this for major version upgrade. Create a separate PR to just move off the alpha package - should be a more simple and non-breaking. 

Thoughts?

shapirov103 (commented):
@zjaco13 please see my comment, happy to discuss.

lib/cluster-providers/import-cluster-provider.ts
zjaco13: It will be breaking for pretty much every customer that upgrades the package.  We are already migrated off alpha I think

shapirov103: Deferring the merge then. I will put request changes - that will be 2.x version.

shapirov103 (changes requested):
Setting to "changes requested" to defer till 2.x version or clear migration strategy for existing stacks.

PR #1263, opened 2026-07-17
Karpenter: split controller IAM policy, add missing permissions, fix example AMI

shapirov103:
## What this includes

**1. Split the Karpenter controller IAM policy into several smaller policies.**
Before, all Karpenter permissions were in one IAM managed policy. The cluster name is repeated many times inside it, so for clusters with long names it can go over the IAM policy size limit (6,144 characters) and fail. This splits it into smaller named policies (node lifecycle, IAM integration, EKS integration, resource discovery, zonal shift, interruption), the same way upstream Karpenter did (aws/karpenter-provider-aws#8690).

**2. Added permissions Karpenter needs that were missing.**
`ec2:DescribeCapacityReservations`, `ec2:DescribeInstanceStatus`, `ec2:DescribePlacementGroups`, and `arc-zonal-shift:GetManagedResource`. Without these, features like capacity reservations, placement groups, and zonal shift would fail.

**3. Fixed the example AMI in `blueprint-construct`.**
It pointed at `bottlerocket@v1.34.0`, which is not a real AMI and did not resolve. Changed to `bottlerocket@latest` so it always finds a valid AMI for the cluster's Kubernetes version.

**4. Two small test fixes** (needed after the 1.19.0 merge):
- Node groups that set only `maxSize` no longer fail because of the new default minimum size.
- Added EKS 1.35 to the Karpenter version map (1.35 is now the default).

## How it was tested

- Unit tests pass (41/41).
- Deployed to a live 1.35 cluster: the update was clean, the controller role now has the split policies, and the controller kept running.
- Confirmed Karpenter can still create a node (launched a node, ran a pod, then removed it).
- Confirmed `bottlerocket@latest` resolves to a valid AMI.

### DIFF
--- a/examples/blueprint-construct/index.ts
+++ b/examples/blueprint-construct/index.ts
@@ -84,7 +84,11 @@ export default class BlueprintConstruct {
     this.nodeClassSpec = {
       amiFamily: "Bottlerocket",
       amiSelectorTerms: [
-        { alias: "bottlerocket@v1.34.0" },
+        // Use the `@latest` alias so the AMI resolves to the newest Bottlerocket build for the
+        // cluster's Kubernetes version. A pinned version (e.g. bottlerocket@v1.34.0) refers to a
+        // Bottlerocket OS release, not the K8s version, and can fail to resolve ("failed to discover
+        // any AMIs for alias") — which also breaks as the default cluster version advances.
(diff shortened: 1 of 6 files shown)


elamaran11 (approved):
LGTM

